PR #9, opened 2022-05-13
Add the misconduct question

felixclack:
We want to ask if there is any record of misconduct during the
eligibility check.

I opted to use the form object approach we took with Find a Lost TRN.

The validations for each step are contained within a form object and
this in turn delegates the persistence to a DB backed model.

Right now, only the happy path, ie. no misconduct, is implemented.

I assumed that we can implement the ineligible path in a future PR.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -9,5 +9,4 @@ en:
         misconduct_form:
           attributes:
             free_of_sanctions:
-              inclusion: Tell us if your employment record is free of sanctions
-              presence: Tell us if your employment record is free of sanctions
+              blank: Tell us if your employment record is free of sanctions


config/locales/en.yml
```diff
@@ -2,3 +2,12 @@ en:
...
     email: my-service@email
+
+  activemodel:
+    errors:
+      models:
+        misconduct_form:
+          attributes:
+            free_of_sanctions:
+              inclusion: Tell us if you are free of sanctions
+              presence: Tell us if you are free of sanctions
```
tvararu, line 13:
```suggestion
              inclusion: Tell us if your employment record is free of sanctions
              presence: Tell us if your employment record is free of sanctions
```

tvararu (approved):
:+1:

PR #10, opened 2022-05-16
Add an ineligible path to the flow

felixclack:
Whenever someone answers 'no' to a question, we want to inform them that
they are ineligible to apply for QTS.

The message we display is specific to each question.

I opted to store the copy for the specific ineligible reasons in the
locales file. This makes it easier to manage the copy in a single place
and also ensures that translations in the future are considered.

### DIFF
--- a/app/views/teacher_interface/pages/ineligible.html.erb
+++ b/app/views/teacher_interface/pages/ineligible.html.erb
@@ -3,7 +3,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">
-      You're not eligible to apply for qualified teacher status (QTS) in England
+      You’re not eligible to apply for qualified teacher status (QTS) in England
     </h1>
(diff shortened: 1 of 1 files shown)


app/views/teacher_interface/pages/ineligible.html.erb
```diff
@@ -0,0 +1,11 @@
+<% content_for :page_title, "You're not eligible" %>
```
tvararu, line 1:
```suggestion
<% content_for :page_title, "You’re not eligible" %>
```

app/views/teacher_interface/pages/ineligible.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<% content_for :back_link_url, back_link_url %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      You're not eligible to apply for qualified teacher status (QTS) in England
```
tvararu, line 6:
```suggestion
      You’re not eligible to apply for qualified teacher status (QTS) in England
```

tvararu (approved):
Just some smart quote related nits

PR #16, opened 2022-05-17
Add the 'recognised' question

felixclack:
We want to ask if the person is recognised as a teacher in the country
they trained in.

If the answer is no, we present the ineligible page with a relevant
explanation.

Screen Shot 2022-05-17 at 12 31 08 pm

Screen Shot 2022-05-17 at 12 31 24 pm

Screen Shot 2022-05-17 at 12 38 26 pm

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -8,9 +8,14 @@ en:
       eligibility_check:
         ineligible_reason:
           misconduct: To teach in England, you must not have any findings of misconduct or restrictions on your practice.
+          recognised: This is because you are not recognised as a school teacher in the country where you trained.
 
     errors:
       models:
+        country_form:
+          attributes:
+            recognised:
+              inclusion: Tell us if you are recognised as a teacher in the country you trained in
         misconduct_form:
           attributes:
             free_of_sanctions:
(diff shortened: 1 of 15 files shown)


config/locales/en.yml
```diff
@@ -8,9 +8,14 @@ en:
       eligibility_check:
         ineligible_reason:
           misconduct: To teach in England, you must not have any findings of misconduct or restrictions on your practice.
+          recognised: This is because you are not recognised as a school teacher in the country where you trained.
```
tvararu, line 12: I like how this is growing out :+1:

db/schema.rb
```diff
@@ -10,14 +10,15 @@
...
-ActiveRecord::Schema[7.0].define(version: 2022_05_13_100913) do
+ActiveRecord::Schema[7.0].define(version: 2022_05_17_110140) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
   create_table "eligibility_checks", force: :cascade do |t|
     t.boolean "free_of_sanctions"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.boolean "recognised"
```
tvararu, line 21: At some point we might want to track the `rejection_reason`. One for later, just to keep in mind.

tvararu (approved):

PR #22, opened 2022-05-18
Extend the form generator to add attributes

felixclack:
We want the form generator to write more of the boilerplate for us.

In this change, I have co-opted the convention for defining attributes
on a model, eg. attribute:type, as a way of defining attributes for the
form object and a related one on the EligiblityCheck.

Currently, this only supports boolean types. These are the easiest to
output the HTML components for.

Running `rails generate form FormName attribute:type` results in
creating the relevant form object, adding the attribute to both it and
the `EligibilityCheck`.

Also, the component for the question HTML gets added.

This will reduce the amount of custom code we need to write to add a new
question to the flow.

### DIFF
--- a/lib/generators/form/USAGE
+++ b/lib/generators/form/USAGE
@@ -4,10 +4,11 @@ Description:
     """
 
 Example:
-    bin/rails generate form Thing
+    bin/rails generate form Thing attribute:type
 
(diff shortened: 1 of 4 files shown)


tvararu (approved):
Nice!

PR #27, opened 2022-05-18
Add the 'Do you have a degree?' question

felixclack:
We want to extend the generated code to ask the question 'Do you have a
degree?'.

One wrinkle with how we have implemented the ineligible page is that we
don't accommodate multi-paragraph copy yet.

I opted not to address that in this PR and leave it for a future
change.

Screen Shot 2022-05-18 at 4 32 06 pm
Screen Shot 2022-05-18 at 4 32 12 pm

Screen Shot 2022-05-18 at 4 32 00 pm

### DIFF
--- a/app/controllers/teacher_interface/degrees_controller.rb
+++ b/app/controllers/teacher_interface/degrees_controller.rb
@@ -1,0 +1,23 @@
+module TeacherInterface
+  class DegreesController < BaseController
+    def new
+      @degree_form = DegreeForm.new
+    end
+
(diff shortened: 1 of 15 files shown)


tvararu:
> One wrinkle with how we have implemented the ineligible page is that we
> don't accommodate multi-paragraph copy yet.

Indeed. We could rethink it and use something like markdown maybe.

tvararu (approved):
Looks really clean!

PR #32, opened 2022-05-20
Add govuk-country-and-territory-autocomplete

felixclack:
There is an existing library that implements the country autocomplete
field that is suitable for our purposes.

[govuk-country-and-territory-autocomplete](https://github.com/alphagov/govuk-country-and-territory-autocomplete)

The country list is provided by this library. An assumption here is that
the list it provides is suitable for what we need at the moment.

With this addition, it is no longer possible to reach the ineligible
country page. We're assuming that we will address that in a future
change when we have defined what makes a country ineligible for the
purposes of this check.

I had to add a style for the list of options that renders when using
autocomplete. The default styles appear to rely on the default font
styles, which for some reason aren't being set.

Screen Shot 2022-05-20 at 11 47 17 am

### DIFF
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -18,6 +18,7 @@
     <%= favicon_link_tag asset_path('images/govuk-apple-touch-icon-180x180.png'), rel: 'apple-touch-icon', type: 'image/png', size: '180x180' %>
 
     <%= stylesheet_link_tag "application" %>
+    <link rel="stylesheet" href="/location-autocomplete.min.css" />
     <%= javascript_include_tag "application", defer: true %>
   </head>
 
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -3,3 +3,9 @@ $govuk-new-link-styles: true;
 $govuk-assets-path: "/";
 
 @import "govuk-frontend/govuk/all";
+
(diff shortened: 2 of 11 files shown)


app/views/layouts/base.html.erb
```diff
@@ -18,6 +18,7 @@
     <%= favicon_link_tag asset_path('images/govuk-apple-touch-icon-180x180.png'), rel: 'apple-touch-icon', type: 'image/png', size: '180x180' %>
 
     <%= stylesheet_link_tag "application" %>
+    <link rel="stylesheet" href="/location-autocomplete.min.css" />
```
tvararu, line 21: Could this be included via a Sass `@import` statement in our main `application.scss`?

That way we don't have to copy it to the public folder and keep it in version control.

There are some caching benefits to having it it separate however. If we go this route, I'd prefer to have it as a static / unminified CSS file in our `app/assets/stylesheets` folder, and `stylesheet_link_tag` to it as usual so that `propshaft` picks it up and fingerprints it.

Combining both thoughts above could mean an `accessible-autocomplete.scss` in our stylesheets folder, that is just a single line import of the CSS from the node bundle, but fingerprinted through `propshaft` and I think is "best of both worlds" and truly a premature optimisation for the ages. :rocket:

tvararu (approved):
Nothing blocking, just sweating the details of a tiny CSS file a bit too much.

PR #33, opened 2022-05-23
Manage the autocomplete CSS through propshaft

felixclack:
The location autocomplete CSS is currently stored as a minified copy of
the gem's CSS in the public folder of the app.

This means it sits outside the digest functionality managed by
Propshaft, which in turn means no cache-busting.

In order to get the benefits of the digest, we need to serve the CSS in
as part of the application.css bundle.


tvararu (approved):
:rocket:

PR #35, opened 2022-05-23
Fix validation error message for misconduct

felixclack:
The misconduct question displays the standard Rails validation error
when no selection is made.

This change ensures we present the correct error message.

Screen Shot 2022-05-23 at 11 28 57 am

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -31,7 +31,7 @@ en:
         misconduct_form:
           attributes:
             free_of_sanctions:
-              blank: Tell us if your employment record is free of sanctions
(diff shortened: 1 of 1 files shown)


David-Feetenby (approved):

config/locales/en.yml
```diff
@@ -32,6 +32,7 @@ en:
           attributes:
             free_of_sanctions:
               blank: Tell us if your employment record is free of sanctions
```
thomasleese, line 34: I think we might be able to get rid of this line now, since it's only an inclusion check?

thomasleese (approved):

PR #36, opened 2022-05-23
Configure Sentry

thomasleese:
This initialises Sentry for Rails with a few extra options.

As recommended by the Sentry docs, this configures the breadcrumbs to be picked up from the Active Support and HTTP loggers, this enhances the breadcrumbs we'll see in Sentry issues.

Certain URL parameters (ones containing sensitive information such as auth tokens) are filtered out.

Some exceptions have been included, specifically ones where there has been a client error when making the request and we wouldn't be able to action.

We don't need to specify the DSN or the environment as these are picked up automatically from environment variables, `SENTRY_DSN` and `SENTRY_ENVIRONMENT`. If the environment is not set, it defaults to
`development` (https://github.com/getsentry/sentry-ruby/blob/eb8c59861cc3dc29a17282d432309edd86db8286/sentry-ruby/lib/sentry/configuration.rb#L467-L469).

[Trello Card](https://trello.com/c/3t0fYW7G/440-sentry-error-tracking)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,8 @@ gem "jsbundling-rails"
 gem "pg"
 gem "propshaft"
 gem "puma", "~> 5.0"
+gem "sentry-rails"
+gem "sentry-ruby"
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby] # Windows
 
 gem "govuk-components"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -270,6 +270,14 @@ GEM
     rubocop-rspec (2.10.0)
       rubocop (~> 1.19)
     ruby-progressbar (1.11.0)
+    sentry-rails (5.3.1)
+      railties (>= 5.0)
+      sentry-ruby-core (~> 5.3.1)
+    sentry-ruby (5.3.1)
+      concurrent-ruby (~> 1.0, >= 1.0.2)
+      sentry-ruby-core (= 5.3.1)
+    sentry-ruby-core (5.3.1)
+      concurrent-ruby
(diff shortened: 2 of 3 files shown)


tvararu (approved):

tvararu:
Can you pick up a `SENTRY_DSN` locally and force the app to `raise "Hello slack"` or similar, and then we can check if the errors show up in `#ŧra_digital_tech`?

thomasleese:
> Can you pick up a `SENTRY_DSN` locally and force the app to `raise "Hello slack"` or similar, and then we can check if the errors show up in `#ŧra_digital_tech`?

It seems like Sentry [is having problems at the moment](https://status.sentry.io/incidents/0nwt1nhcf5lv), I tested it locally about 30 minutes ago and the issue has only just appeared in Sentry and I don't see anything in Slack... I'm confident the code works though.

thomasleese:
I've tested this locally with the live DSN and the following issue has been reported: https://sentry.io/organizations/dfe-teacher-services/issues/3294986816/?project=6426061&query=is%3Aunresolved

tvararu:
> I've tested this locally with the live DSN and the following issue has been reported: https://sentry.io/organizations/dfe-teacher-services/issues/3294986816/?project=6426061&query=is%3Aunresolved

Yep, nice, I think we still need to set up a "Slack alert" in the sentry settings. Can be done separately as it doesn't involve code 👍

thomasleese:
> Yep, nice, I think we still need to set up a "Slack alert" in the sentry settings. Can be done separately as it doesn't involve code 👍

Yeah, I think this'll have to be done when Sentry is working again!

PR #37, opened 2022-05-23
Remove historic countries from autocomplete

thomasleese:
The `govuk-country-and-territory-autocomplete` is designed to cater to a wide variety of "Country picking needs." One example is the situation where an old person is specifying the country they were born in. Those countries might not exist anymore.

However, in the case of our service, those countries can be confusing.

This approach mirrors those taken by https://github.com/DFE-Digital/npq-prototype/commit/f1c44082606f83efbb8351fe9cac4b5a75abf5d9 and https://github.com/DFE-Digital/npq-registration/pull/308.

This approach copies the contents of the countries as provided by the library in to this repo, which does mean if the library changes, it won't be reflected for us. I think this is a reasonable downside to avoid a more complicated solution for something which is unlikely to be needed often.

[Trello Card](https://trello.com/c/OireZ4bN/444-remove-historic-countries)


tvararu (approved):
Pragmatic 👍

PR #38, opened 2022-05-24
Fix the country autocomplete error state

felixclack:
When a Rails form has a validation error message, the form builder
outputs a different ID for the element.

This breaks the autocomplete JS because it is targeting the original ID.

While making this change, I took the opportunity to wrap the
autocomplete function with a check that the relevant HTML element is
present to avoid it erroring on other pages that don't need it.

Screen Shot 2022-05-24 at 9 20 00 am

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -3,7 +3,20 @@ import openregisterLocationPicker from "govuk-country-and-territory-autocomplete
 
 initAll();
 
-openregisterLocationPicker({
-  selectElement: document.getElementById("location-form-country-code-field"),
-  url: "/teacher/locations.json",
-});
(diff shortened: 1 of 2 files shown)


tvararu (approved):
> I'm surprised the ID changes in that case!

That's the behaviour of the [form builder](https://govuk-form-builder.netlify.app/); I agree in this particular scenario it is surprising but I don't think it's a bad call.

An alternative to this could be to not `getElementById` the auto-generated ID that the form builder creates. Instead, we could wrap the parent element with an explicit ID of our choosing and `querySelector` for a child that is more likely to stay the same.

That being said that feels like overkill, this should work fine for a long time.

PR #39, opened 2022-05-24
Add a feature flag for :service_open

felixclack:
We want to control access to the service for the general public.

I have introduced feature flags, as per our approach in [Find a lost TRN](https://github.com/DFE-Digital/find-a-lost-trn/pull/57).

I chose to defer the way we manage our ENV variables in development so
that we can have a dedicated PR for that. For the time being, the
default values are test/test.

Also, the support interface to manage the feature flags will be
implemented in a feature change.

Screen Shot 2022-05-24 at 10 15 18 am
Screen Shot 2022-05-24 at 10 15 23 am

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,3 +13,11 @@ inherit_mode:
 Style/NumericLiterals:
   Exclude:
     - db/schema.rb
+
+Style/BlockDelimiters:
+  Exclude:
+    - app/models/feature_flag.rb
+
+Style/MethodCalledOnDoEndBlock:
+  Exclude:
+    - app/models/feature_flag.rb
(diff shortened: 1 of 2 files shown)


thomasleese (approved):
Nice feature flag mechanism! I'd suggest splitting that out in to a separate commit so it can be reviewed separately, but it's not too important.

.rubocop.yml
```diff
@@ -13,3 +13,11 @@ inherit_mode:
...
   Exclude:
     - db/schema.rb
+
+Style/BlockDelimiters:
+  Exclude:
+    - app/models/feature_flag.rb
+
+Style/MethodCalledOnDoEndBlock:
+  Exclude:
+    - app/models/feature_flag.rb
```
tvararu, line 23: Should this be across the project to prevent Rubocop and Prettier arguing about it in the future?

Also I agree we should defer to Prettier because it does more.

tvararu (approved):

PR #44, opened 2022-05-25
Combine common logic from controllers

thomasleese:
Depends on #43 

This moves common logic related to the eligibility check model in to the base controller so it's loaded and saved in the same way. Also adds a `success_url` to all the forms so they have a consistent interface and the controllers are all interacting with the form in the same way, allowing us to make the controllers more generic in the future.

### DIFF
--- a/app/controllers/teacher_interface/base_controller.rb
+++ b/app/controllers/teacher_interface/base_controller.rb
@@ -1,4 +1,18 @@
 module TeacherInterface
   class BaseController < ApplicationController
+    after_action :save_eligibility_check_id
+
+    def eligibility_check
+      @eligibility_check ||=
+        if session[:eligibility_check_id]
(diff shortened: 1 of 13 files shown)


tvararu (approved):
This feels like the right abstraction to me :+1: We opted to repeat ourselves a bit early on, but now feels like a good time to take a step back and do this

@felixclack should also weigh in

tvararu:
Just remembered: do we have to update the [form generator](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/lib/generators/form/form_generator.rb)?

thomasleese:
> Just remembered: do we have to update the [form generator](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/lib/generators/form/form_generator.rb)?

Oh good point, I'll add that now.

PR #47, opened 2022-05-25
Add a feature flags support interface

felixclack:
We want to provide a UI for activating/dectivating the feature flags in
the support interface.

This change introduces a UI that is used in [Find](https://github.com/DFE-Digital/find-a-lost-trn/pull/76).

Screen Shot 2022-05-25 at 10 06 32 am
Screen Shot 2022-05-25 at 10 06 38 am

### DIFF
--- a/app/controllers/support_interface/base_controller.rb
+++ b/app/controllers/support_interface/base_controller.rb
@@ -1,4 +1,6 @@
 module SupportInterface
   class BaseController < ApplicationController
+    http_basic_authenticate_with name: ENV.fetch("SUPPORT_USERNAME", "test"),
+                                 password: ENV.fetch("SUPPORT_PASSWORD", "test")
   end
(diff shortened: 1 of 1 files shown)


thomasleese (commented):
I think this page will be accessible by anyone if the `service_open` flag is set, do we need to put something in `SupportInterface::BaseController` to require authentication regardless of the flag?

tvararu:
> I think this page will be accessible by anyone if the `service_open` flag is set, do we need to put something in `SupportInterface::BaseController` to require authentication regardless of the flag?

Yep, the way it works on Find is:

- `:service_open false` -> Basic auth everywhere
- `:service_open true` -> Basic auth just on `/support`
-

thomasleese (approved):

tvararu (approved):

PR #48, opened 2022-05-25
Group countries in to eligible, legacy or not eligible

thomasleese:
This sends users to a different place depending on the country they pick. Eligible countries are sent through the checker, legacy countries are sent to the old service and the rest are marked as not eligible.

For now, I've hard coded the countries in the `EligibilityCheck` model, but in the future, as we have more customised functionality per country, we would want to build a database or have a static YAML file containing this information. I decided that would make sense to implement in a separate PR.

[Trello Card](https://trello.com/c/uAzoRaR7/461-split-countries-into-private-beta-ones-straight-to-dqt-ones-and-ineligible-ones)

### DIFF
--- a/app/forms/country_form.rb
+++ b/app/forms/country_form.rb
@@ -13,17 +13,13 @@ class CountryForm
     eligibility_check.save!
   end
 
-  def eligible?
-    eligibility_check.country_code != "INELIGIBLE"
-  end
-
   def success_url
-    unless eligible?
-      return(
-        Rails.application.routes.url_helpers.teacher_interface_ineligible_path
-      )
+    if eligibility_check.eligible_country_code?
+      Rails.application.routes.url_helpers.teacher_interface_degree_path
+    elsif eligibility_check.legacy_country_code?
+      "https://teacherservices.education.gov.uk/MutualRecognition/"
+    else
+      Rails.application.routes.url_helpers.teacher_interface_ineligible_path
     end
-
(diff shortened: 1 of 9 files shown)


app/forms/country_form.rb
```diff
@@ -1,29 +1,25 @@
...
-
   def success_url
-    unless eligible?
-      return(
-        Rails.application.routes.url_helpers.teacher_interface_ineligible_path
-      )
+    if eligibility_check.eligible_country_code?
+      Rails.application.routes.url_helpers.teacher_interface_degree_path
+    elsif eligibility_check.legacy_country_code?
+      "https://teacherservices.education.gov.uk/MutualRecognition/"
```
tvararu, line 20: Ah, this explains the `allow_other_host` from earlier.

We can make this an intermediary page in a follow-up PR once we have a design for it, and then it can be a regular `link_to` in the page

thomasleese: Yep, agreed.

spec/system/eligibility_spec.rb
```diff
@@ -205,6 +224,10 @@ def then_i_see_the_ineligible_page
     )
   end
 
+  def then_i_see_the_legacy_service
+    expect(page).to have_current_path("/MutualRecognition/")
```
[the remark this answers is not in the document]
thomasleese: Yeah, I didn't like this either, and couldn't think of a nice way around it. I agree we can leave it like this until we've got the intermediate page.

PR #54, opened 2022-05-26
Fix docker build

tvararu:
The current Docker build produces an `application.css` in the `app/assets/builds` folder, but it doesn't get copied and fingerprinted to the `public/assets` folder.

I don't know why yet, but to unblock the pipeline, running `yarn build && yarn build:css` seems to work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,10 +5,6 @@
 # Build builder image
 FROM ruby:3.1.1-alpine as builder
 
-# RUN apk -U upgrade && \
-#     apk add --update --no-cache gcc git libc6-compat libc-dev make nodejs \
-#     postgresql13-dev yarn
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,10 +5,6 @@
 # Build builder image
 FROM ruby:3.1.1-alpine as builder
 
-# RUN apk -U upgrade && \
-#     apk add --update --no-cache gcc git libc6-compat libc-dev make nodejs \
-#     postgresql13-dev yarn
```
tvararu, line 10: I forgot to remove this when I created our Dockerfile.

prabhakarreddy1234 (approved):

thomasleese (approved):

PR #60, opened 2022-05-27
Allow users to select a region

thomasleese:
This allows users to select a region if the country they selected has more than one region. There hasn't been any design input on this and I've decided to go with a radio button, but we could easily change it to a drop down if that works better.

Screenshot 2022-05-27 at 08 01 26

[Trello Card](https://trello.com/c/7u9sFwI6/467-ask-users-for-their-state-when-selecting-certain-countries-like-canada-australia)

### DIFF
--- a/app/controllers/teacher_interface/region_controller.rb
+++ b/app/controllers/teacher_interface/region_controller.rb
@@ -1,0 +1,29 @@
+module TeacherInterface
+  class RegionController < BaseController
+    before_action :load_regions
+
+    def new
+      @region_form = RegionForm.new
+    end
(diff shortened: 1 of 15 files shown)


dankmitchell:
Radios are good for this amount of options. Nice work @thomasleese 👌

benilovj:
You probably would still want to apply some sorting (alphabetical?) to the region names.

thomasleese:
> You probably would still want to apply some sorting (alphabetical?) to the region names.

Good point, I've amended the commit with this functionality.

PR #62, opened 2022-05-27
Track complete checks

felixclack:
We want a way of being able to tell whether someone completed an
eligibility check.

I considered storing this as multiple fields, to represent the different
states a check could be in, eg. eligible_at and ineligible_at.

This seemed like overkill and introduces slightly more complexity than
the simpler approach of storing the timestamp of when the person reached
either the eligible or ineligible page.

We can calculate the eligibility status from the existing fields, so
having multiple timestamp fields seemed redundant.

I also removed the session value representing when the flow was
completed. This was something we brought across from Find but isn't as
relevant to the check flow.

If, in the future, we do need to know if the final screen was displayed,
we can use the presence of the `completed_at` field to determine this.

### DIFF
--- a/app/controllers/teacher_interface/pages_controller.rb
+++ b/app/controllers/teacher_interface/pages_controller.rb
@@ -1,6 +1,7 @@
 module TeacherInterface
   class PagesController < BaseController
     before_action :load_eligibility_check
+    before_action :complete_eligibility_check, only: %i[eligible ineligible]
 
     def eligible
       session[:eligibility_check_complete] = true
@@ -10,6 +11,10 @@ module TeacherInterface
     end
 
     private
+
+    def complete_eligibility_check
+      @eligibility_check.complete!
+    end
 
     def load_eligibility_check
       @eligibility_check = eligibility_check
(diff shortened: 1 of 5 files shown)


app/controllers/teacher_interface/pages_controller.rb
```diff
@@ -3,10 +3,11 @@ class PagesController < BaseController
     before_action :load_eligibility_check
 
     def eligible
-      session[:eligibility_check_complete] = true
+      @eligibility_check.touch(:completed_at)
```
thomasleese, line 6: Could we move this in to a `before_action` for both methods?

thomasleese: I also wonder what happens if someone goes through the flow again, do we need to create a new `EligibilityCheck` if the current one is marked as complete or do we just take the user straight to the results page?

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/04863ce9a551c53d71c96325ef69f0af652e2cea/app/controllers/teacher_interface/base_controller.rb#L8

thomasleese (approved):
Nice, this looks good. Definitely like the idea of not storing the final state in a redundant field!

tvararu (approved):

PR #63, opened 2022-05-30
Add drafts for the static pages

thomasleese:
This isn't the final content, but it adds the pages in advance of the content being ready.

### DIFF
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -43,6 +43,7 @@
       meta_items: {
         'Accessibility': '/accessibility',
         'Cookies': '/cookies',
+        'Privacy': '/privacy',
       }) %>
   </body>
 </html>
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -41,6 +41,7 @@
     <%= govuk_footer(
       meta_items_title: 'Footer links',
       meta_items: {
+        'Accessibility': '/accessibility',
         'Cookies': '/cookies',
       }) %>
   </body>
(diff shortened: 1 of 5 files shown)


app/views/layouts/base.html.erb
```diff
@@ -41,7 +41,9 @@
     <%= govuk_footer(
       meta_items_title: 'Footer links',
       meta_items: {
+        'Accessibility': '/accessibility',
         'Cookies': '/cookies',
+        'Privacy': '/privacy',
```
tvararu, line 46: I think we should comment the links out. We should be careful about making the service "look complete" when actually some of it is placeholder, as someone who isn't familiar might say "okay it's done" and we might ship it before it's ready.

config/locales/en.yml
```diff
@@ -6,7 +6,8 @@ en:
 
   service:
     name: Apply for qualified teacher status (QTS) in England
-    email: my-service@email
+    email: qts.enquiries@education.gov.uk
```
tvararu, line 9: Is this the right email? Looks good, just would like to double check in Slack if you haven't already.

tvararu (approved):

app/views/layouts/base.html.erb
thomasleese: I figured the HTTP basic authentication prevented us from thinking the service looked complete, but happy to comment them out if you think that's worth it.

tvararu: Maybe it's fine. Just keep it in mind come go-live that someone still has to do a pass for those. :)

PR #64, opened 2022-05-30
Added Push trigger for `dev` deployment

prabhakarreddy1234:
Added `push` trigger to deploy to `dev` environment.

[x] [Attach to trello](https://trello.com/c/v5IrenGI/477-create-github-action-build-workflow)


.github/workflows/deploy.yml
```diff
@@ -16,6 +16,12 @@ on:
   push:
     branches:
       - main
+
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+    - main
```
tvararu, line 23: We shouldn't deploy `dev` on this project from PRs, because it's a Rails project.

When we boot up the app, migrations are executed, which change the tables/columns in the database: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/d77a5c5a35a90f05323825ae4cd01898ec475a2c/Dockerfile#L75

Once the columns/tables are updated, a special field that contains the last run migration is also updated. When the app boots up, if this field is newer than what it expects, it will crash, and we'd have to roll the database back.

As such, if we deployed to `dev` from an open pull request, the first deploy would work. The live container would be replaced by a container from the branch, and the database would update.

But then if someone else opened another PR, it would override the live container, and since the database has been migrated to a newer version, it would crash on boot. Recovering from this state would require us to SSH and run `rails db:rollback`, possibly losing data.

In a nutshell, we can't auto-deploy from open PRs on this project, unless they also have their own isolated databases (like review apps).

tvararu (changes requested):

tvararu (approved):

PR #70, opened 2022-05-30
Record the decision to track analytics

felixclack:
We want to be able to track the usage of the service while maintaining
people's right to privacy.

### DIFF
--- a/adr/00004-tracking-usage.md
+++ b/adr/00004-tracking-usage.md
@@ -1,0 +1,51 @@
+# 4. Tracking usage
+
+Date: 2022-06-16
+
+## Status
+
+Accepted
+
+## Context
+
+We want a way to track the usage of the eligibility checker while defending people's right to privacy
+while they use the service.
+
+Options considered:
+
+- Google Analytics - we can tailor the settings to ensure that usage of Google Analytics is compliant
+  with GDPR. This will require:
+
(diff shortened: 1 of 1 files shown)


adr/00004-tracking-usage.md
```diff
@@ -0,0 +1,51 @@
...
+## Decision
+
+DfE Analytics is the preferred option. It gives us the most control and allows for progressively enhancing the analytics
+capabilities as required. It also has the least amount of bureaucracy associated with signing agreements and updating policies. It's also the preferred option within DfE and we can make improvements to the analytics library for other users as we develop our service.
+
+## Consequences
+
+We need to setup DfE Analytics in accordance with our requirements. Both creating the integration with BigQuery and the dashboards/queries to visualise the data.
+
+It's worth noting that DfE Analytics is perhaps the least mature of the options proposed, and there is a risk associated with this, we'd be an early adopted. However, as the aim of the library is to streamline analytics for all DfE teams down the line, we can help with that goal.
```
tvararu, line 51:
```suggestion
It's worth noting that DfE Analytics is perhaps the least mature of the options proposed, and there is a risk associated with this, we'd be an early adopter. However, as the aim of the library is to streamline analytics for all DfE teams down the line, we can help with that goal.
```

tvararu (approved):

PR #72, opened 2022-05-31
Various country updates

thomasleese:
This changes the list of countries to match what we'd expect them to be. See individual commits for more details.

### DIFF
--- a/public/location-autocomplete-canonical-list.json
+++ b/public/location-autocomplete-canonical-list.json
@@ -64,7 +64,7 @@
   ["Cuba", "country:CU"],
   ["Curaçao", "territory:CW"],
   ["Cyprus", "country:CY"],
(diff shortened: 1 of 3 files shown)


public/location-autocomplete-canonical-list.json
```diff
@@ -64,7 +64,7 @@
   ["Cuba", "country:CU"],
   ["Curaçao", "territory:CW"],
   ["Cyprus", "country:CY"],
-  ["Czechia", "country:CZ"],
+  ["Czech Republic", "country:CZ"],
```
fofr, line 67: I assume you can still use Czechia as a synonym?

thomasleese: Yes, the name will show as `Czech Republic`, but the synonym `Czechia` remains.

PR #74, opened 2022-05-31
Add yarn to setup

David-Feetenby:
Running yarn is necessary to install JS dependencies.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -17,6 +17,9 @@ FileUtils.chdir APP_ROOT do
   system! "gem install bundler --conservative"
   system("bundle check") || system!("bundle install")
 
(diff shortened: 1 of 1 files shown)


bin/setup
```diff
@@ -17,6 +17,9 @@ FileUtils.chdir APP_ROOT do
   system! "gem install bundler --conservative"
   system("bundle check") || system!("bundle install")
 
+  puts "\n== Installing Yarn =="
```
tvararu, line 20:
```suggestion
  puts "\n== Installing yarn dependencies =="
```

tvararu (changes requested):
Just a small nit :+1:

PR #78, opened 2022-06-09
Redirect to mutual recognition unless service can be started

thomasleese:
This adds a new feature flag to determine if the service can be started, which means users will be taken to the start page, otherwise redirected to the legacy mutual recognition service.

This means `/teacher` becomes the entry point for the service, and users are either redirected to `/teacher/start` or the legacy mutual recognition service. I've decided _not_ to redirect users if the feature isn't on and they navigate directly to `/teacher/start` to make the service easy for us to access, but this decision could be up for discussion.

[Trello Card](https://trello.com/c/B6tT13Yz/495-set-up-redirect-to-mutualrecognition-behind-feature-flag)

### DIFF
--- a/app/models/feature_flag.rb
+++ b/app/models/feature_flag.rb
@@ -3,13 +3,19 @@ class FeatureFlag
 
   attr_accessor :description, :name, :owner
 
-  PERMANENT_SETTINGS = [].freeze
+  PERMANENT_SETTINGS = [
+    [
+      :service_open,
+      "Allow users to access the service without HTTP basic auth",
+      "Felix Clack"
+    ]
+  ].freeze
 
   TEMPORARY_FEATURE_FLAGS = [
     [
-      :service_open,
-      "Allow users to access the service and check their eligibility",
-      "Felix Clack"
+      :service_start,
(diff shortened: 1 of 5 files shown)


app/models/feature_flag.rb
```diff
@@ -3,7 +3,13 @@ class FeatureFlag
 
   attr_accessor :description, :name, :owner
 
-  PERMANENT_SETTINGS = [].freeze
+  PERMANENT_SETTINGS = [
+    [
+      :service_start,
+      "Allow users to use the service, rather than being sent to the legacy mutual recognition site",
```
tvararu, line 9: Can you update the description of the `service_open` flag below, since that also has a similar one?

It should communicate our intent which is to use basic auth pre go-live, and also on the dev/preprod/test environments (most of the time) because those are not meant to be public facing and shouldn't be confused as such.

Also that one should likely be a permanent setting. `:service_start` could be the temporary one I think.

tvararu (approved):
I think 2 flags is actually the better call :+1:

app/models/feature_flag.rb
thomasleese: Good point, I've amended the commit with that.

PR #84, opened 2022-06-10
Add custom country content

thomasleese:
This adds a number of fields to the region to customise what content is shown to the user on the eligible page. See individual commits for more details.

### DIFF
--- a/app/controllers/eligibility_interface/pages_controller.rb
+++ b/app/controllers/eligibility_interface/pages_controller.rb
@@ -16,6 +16,7 @@ module EligibilityInterface
 
     def eligible
       @mutual_recognition_url = MUTUAL_RECOGNITION_URL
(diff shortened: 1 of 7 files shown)


spec/views/teacher_interface/pages/eligible_spec.rb
```diff
@@ -0,0 +1,79 @@
+require "rails_helper"
+
+RSpec.describe "teacher_interface/pages/eligible", type: :view do
```
tvararu, line 3: Ah, view specs, nice. I was going to suggest that another possible improvement could be to extract the conditional content in into a [ViewComponent](https://viewcomponent.org/). This works as well. :+1:

tvararu (approved):
Skimmed most of the content itself, logic is good!

PR #85, opened 2022-06-10
Move legacy status to regions

thomasleese:
This supports the private beta where we only want certain regions to be allowed through the flow, rather than whole countries.

[Trello Card](https://trello.com/c/DXkVA1Sz/511-update-countries-to-match-selected-private-beta-ones)


tvararu (approved):
Nice!

PR #89, opened 2022-06-13
Only complete eligibility check if persisted

thomasleese:
This prevents a Sentry error from being recorded if someone where to navigate directly to the eligible page, for example.

[Sentry Issue](https://sentry.io/organizations/dfe-teacher-services/issues/3345259577/)


tvararu (approved):
Discussed offline: we should consider implementing something like the `EnforceQuestionOrder` concern from Find: https://github.com/DFE-Digital/find-a-lost-trn/blob/main/app/controllers/concerns/enforce_question_order.rb#L34

PR #91, opened 2022-06-13
Render country name with 'the' prefix

thomasleese:
Certain countries, for example 'United States' is typically written as 'the United States' in a sentence. This PR adds a hard coded list of countries where it is necessary to add a 'the' prefix before the name.

This solution is quite a simplistic solution as it only works for English but since our content will always be in English I think that's okay. I did look at using a Gem like `countries`, but it doesn't provide us with an easy way to check whether to include a 'the' or not, instead we'd have to use its names instead of the GOV.UK autocomplete component ones. I was also unsure about putting this method in the model, but a view component felt like overkill.


tvararu (approved):

app/models/country.rb
```diff
@@ -22,9 +22,16 @@ class Country < ApplicationRecord
       .map { |row| [row.last.split(":").last, row.first] }
       .to_h
 
+  COUNTRIES_WITH_DEFINITIVE_ARTICLE =
```
tvararu, line 25: `DEFINITE`* :wink:

thomasleese: Nooo :sob:

PR #93, opened 2022-06-13
Set up Sidekiq

thomasleese:
This sets up Sidekiq in the app so we can start to use background tasks.

Depends on #98

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem "propshaft"
 gem "puma", "~> 5.0"
 gem "sentry-rails"
 gem "sentry-ruby"
+gem "sidekiq"
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby] # Windows
 
 gem "govuk-components"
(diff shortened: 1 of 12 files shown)


adr/00004-sidekiq-and-redis.md
```diff
@@ -0,0 +1,21 @@
+# 6. Sidekiq and Redis
```
tvararu, line 1: We already have an ADR that covers this https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/adr/00002-reuse-existing-architecture.md#decision

thomasleese: Ah yep, I forgot about that one.

tvararu (approved):
Find has the sidekiq web interface at `/support/sidekiq` and links to it from inside Support

Route setup for basic auth protected Sidekiq: https://github.com/DFE-Digital/find-a-lost-trn/blob/main/config/routes.rb#L17-L31

image

PR #99, opened 2022-06-14
Set up DfE Analytics

thomasleese:
Depends on #93


config/initializers/dfe_analytics.rb
```diff
@@ -0,0 +1,4 @@
+DfE::Analytics.configure do |config|
+  config.queue = :analytics
+  config.environment = ENV.fetch("SENTRY_ENVIRONMENT", "development")
```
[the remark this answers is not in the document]
tvararu: For that one, I like the [`HostingEnvironment` module in Apply for Teacher Training](https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/lib/hosting_environment.rb). One clever thing the other Apply does is that they colour / customise the text of the phase banner and header underline depending on which env you're on (mocked it up quickly in this screenshot for illustration)

image

Gemfile
```diff
@@ -17,6 +17,7 @@ gem "sentry-ruby"
 gem "sidekiq"
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby] # Windows
 
+gem "dfe-analytics", git: "https://github.com/DFE-Digital/dfe-analytics"
```
duncanjbrown, line 20: Until we've got a proper gem host set up (DfE is getting us onto RubyGems... eventually) please pin to a tag so we don't accidentally break you! `v1.0.0` is latest rn

thomasleese: The exact version is pinned in the `Gemfile.lock` already, so that's covered. It's following the same pattern we use for other libraries in the `Gemfile` by not including the version there.

config/initializers/dfe_analytics.rb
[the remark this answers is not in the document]
thomasleese: I'll add it to the post MVP list of things to do, as it feels separate to this PR.

Gemfile
duncanjbrown: Ah, I must admit I'm a bit fuzzy on the behaviour of `bundler` here — I had assumed that if you ran `bundle update` it would fetch whatever's at tip of `main`?

thomasleese: Yes, this is correct, but that's actually the behaviour we want since `bundle update` is a call to update the dependencies. Fixing any issues that might arise from a `bundle update` is part of the process for updating dependencies.

The important thing is that doing `bundle install` will use the pinned versions from the lock file, so there will never be any unintended breakages.

thomasleese: Yeah, I suppose it's not ideal, but I'd expect the person doing the `bundle update` or reviewing the Dependabot PR to review the changes to understand what they are. The disadvantage of pinning the version here by a Git tag is that `bundle update` and Dependabot would stop updating it (because it's a Git tag rather than a proper Gem version), so there's a risk we'll end up stuck on an old version of the Gem.

tvararu: I think both methods have their pros/cons... I'm tempted to go with Thomas' suggestion with the understanding that we'll have to be more careful when merging dependency updates. Since it's a dependency that lives inside of our own organisation, I think that's acceptable.

Hopefully this won't be a sticking point for long, if we can get `dfe-analytics` on RubyGems!

tvararu:
I think this could benefit from an environment variable to toggle it off; `BIGQUERY_DISABLE=true` or such. That way, developers can work locally without setting it up and they won't experience console errors.

tvararu (approved):
:shipit:

PR #104, opened 2022-06-15
Add interface for enabling private beta countries

thomasleese:
Currently this functionality exists only in a Rake task, which means we can't run it in production. Instead, we'll add a section to the support interface which will enable the private beta countries.

I decided we don't need any checks or confirmation messages as this action is idempotent, and if anything goes wrong we can disable the "service start" feature and everyone will be sent directly to the mutual recognition service.

Screenshot 2022-06-15 at 09 52 39

[Trello Card](https://trello.com/c/V7UTXFKW/371-break-down-dev-build)


tvararu (approved):
:+1:

PR #106, opened 2022-06-15
Add duration performance

thomasleese:
This adds status on how long it takes users to check their eligibility and displays it on the performance dashboard.

[Trello Card](https://trello.com/c/yv8Q2Hea/457-analytics)

### DIFF
--- a/app/controllers/performance_controller.rb
+++ b/app/controllers/performance_controller.rb
@@ -16,6 +16,8 @@ class PerformanceController < ApplicationController
     @eligible_checks, @submission_data = stats.submission_results
 
     @countries, @country_data = stats.country_usage
+
+    @duration_data = stats.duration_usage
   end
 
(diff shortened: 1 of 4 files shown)


tvararu (approved):
:+1:

What does the table look like on mobile? Wondering if we could have a fallback that displays the same data in a [summary list](https://design-system.service.gov.uk/components/summary-list/), and toggle which is being displayed using a media query.

thomasleese:
> What does the table look like on mobile? Wondering if we could have a fallback that displays the same data in a [summary list](https://design-system.service.gov.uk/components/summary-list/), and toggle which is being displayed using a media query.

I like this idea, I think we might need to get a designer involved as there's explicit mention of not using a summary list for tabular data, but I can see that the table component isn't fantastic on mobile (although it's manageable).

PR #110, opened 2022-06-16
Add a worker CloudFoundary app

thomasleese:
This will be used to run Sidekiq and process background jobs. I've decided to use the same memory, disk quota and instances as the web app, but if we decide this isn't appropriate in the future we can always change it. I've also set the deployment strategy to standard as we don't need two Sidekiq processes running in parallel during deployment, and it keeps things simple.

### DIFF
--- a/terraform/app.tf
+++ b/terraform/app.tf
@@ -1,5 +1,8 @@
 locals {
-  app_environment_variables = try(local.application_environment_variables, {})
+  app_environment_variables = merge(try(local.application_environment_variables, {}), {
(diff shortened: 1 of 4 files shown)


tvararu (approved):

terraform/app.tf
```diff
@@ -1,12 +1,20 @@
 locals {
-  app_environment_variables = try(local.application_environment_variables, {})
+  azure_environment_variables = try(yamldecode(
```
tvararu, line 2: Better :+1:

PR #123, opened 2022-06-17
Add basic country management interface

thomasleese:
This adds a simple support interface for managing the countries and regions. Currently it shows a page where users can select a country and region they want to edit, and then they can edit that region. It doesn't allow adding or removing regions, or countries.

Screenshot 2022-06-17 at 16 27 03

Screenshot 2022-06-17 at 16 27 13


thomasleese:
We should wait until after we've launched and pressed the "Private beta countries" button to merge this in.

tvararu (approved):
Good dev design! :paintbrush:

tvararu:
Button has been pressed in Production btw

PR #128, opened 2022-06-20
Run seeds when setting private beta countries

thomasleese:
This ensures that the countries exist and avoids us needing to run a Rake task in a console in a production environment.


tvararu (approved):
The seeds are idempotent so this should be good :+1:

PR #135, opened 2022-06-21
Fix typo in request_referer column

thomasleese:
In [dfe-analytics the name is `request_referer`](https://github.com/DFE-Digital/dfe-analytics-dataform/search?q=request_referer) to match the (incorrect!) spelling of the HTTP header.

See https://en.wikipedia.org/wiki/HTTP_referer#Etymology


tvararu (approved):
:+1: I have never actually noticed that...

PR #146, opened 2022-06-24
Use Devise for staff HTTP Basic auth

thomasleese:
This changes how we authenticate access to the support interface, by using a Devise/Warden strategy which uses HTTP Basic auth if there are no staff members, or normal Devise authentication if there are staff members.

This means that regardless of how the user got to the support/assessor interface (either via database authentication or via this new strategy), the controllers will handle them in the same way reducing potential duplication.

I've also added a feature so we can enable support for HTTP basic auth at any point, but by default it's only enabled if there are no staff users.


tvararu (approved):
:shipit:

PR #165, opened 2022-06-28
Refactor performance controller

David-Feetenby:
Move timezone.now out of performance_controller
so that the logic is done inside performance_states


thomasleese (approved):
Nice improvement!

PR #166, opened 2022-06-28
Build staff invitation functionality

thomasleese:
This completes the installation of `devise_invitable` following #147 and #146. There are various steps involved in this which I've split in to separate commits, but the main ones are:

- Creating an anonymous support user type
- Sending Devise emails using GOV.UK Notify
- Adding test for staff invitation

To finish off the staff management, there will be one follow up PR which adds a section in the support interface to invite staff and view a table of staff users.


Gemfile
```diff
@@ -32,6 +32,7 @@ group :development, :test do
 end
 
 group :development do
+  gem "pry"
```
tvararu, line 35: I've been using `binding.break` from Rails 7 instead of Pry lately https://www.mintbit.com/blog/byebug-replaced-with-debug-rails7

Is Pry better?

app/components/flash_message_component.rb
```diff
@@ -1,14 +1,18 @@
...
+  DEVISE_PRIMARY_KEYS = { alert: :warning, notice: :info }.freeze
 
   def initialize(flash:)
     super
     @flash = flash.to_hash.symbolize_keys!
   end
 
   def message_key
-    flash.keys.detect { |key| ALLOWED_PRIMARY_KEYS.include?(key) }
+    key =
```
tvararu, line 11: Does the CSS for the component still render correctly with Devise flash keys? The `classes` method generates a CSS class based on this key.

tvararu (approved):

Gemfile
tvararu: The one case where both kind of fail in my testing is when combined with foreman; it swallows the repl somehow and I can't see what I'm typing. In that case I fall back to running `rails s` manually

spec/support/dfe_analytics.rb
```diff
@@ -0,0 +1,5 @@
+RSpec.configure do |config|
+  config.around(:example, type: :system) do |example|
+    ClimateControl.modify(BIGQUERY_DISABLE: "true") { example.run }
+  end
+end
```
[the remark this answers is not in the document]
duncanjbrown: Does requiring `dfe/analytics/testing` in your rspec helper help? That should engage `DfE::Analytics::Testing.fake!` which noops for all events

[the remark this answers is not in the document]
thomasleese: Oh nice, I didn't know that existed. I'll refactor to use that instead.

PR #169, opened 2022-06-28
Add basic staff management interface

thomasleese:
It currently only allows you to invite a new staff user, sign in as a staff user, and then view a list of all the staff users.

Screenshot 2022-06-28 at 17 06 02
Screenshot 2022-06-28 at 17 06 44


app/views/support_interface/staff/index.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+    if staff.created_by_invite?
+      summary_list.row do |row|
+        row.key { 'Invitation status' }
+
+        row.value do
+          if staff.invitation_accepted?
+             govuk_tag(text: "Accepted", colour: "green")
+          else
+            govuk_tag(text: "Not accepted", colour: "red")
```
tvararu, line 26: I think red is more for failed states:

```suggestion
            govuk_tag(text: "Not accepted", colour: "grey")
```

Though maybe red is better to draw attention?

Just bikeshedding, feel free to ignore.

tvararu (approved):

app/views/support_interface/staff/index.html.erb
thomasleese: Hmm good point, I like that it draws attention as really we should have too many users with open invitations about. I think I'll leave it and see what a designer has to say.

PR #174, opened 2022-06-29
Make "Start now" button a submit button 

thomasleese:
This allows us to POST to an endpoint which initially creates the `EligibilityCheck` instance. This means the `created_at` time is a better representation of when a user actually started using the service. Currently, the eligibility check is only created after a user answers the first question, which means the duration performance metrics are slightly skewed.


app/views/eligibility_interface/start/show.erb
```diff
@@ -21,6 +21,11 @@
 
     <p class="govuk-body">There’s no charge to apply for QTS. You do not need to be in the UK to use this service.</p>
 
-    <%= govuk_start_button(text: 'Start now', href: eligibility_interface_countries_path, classes: 'govuk-!-margin-top-2 govuk-!-margin-bottom-9') %>
+    <%= form_with url: eligibility_interface_start_path, method: :post do |f| %>
+      <button type="submit" formnovalidate="formnovalidate" class="govuk-button govuk-button--start govuk-!-margin-top-2 govuk-!-margin-bottom-9" data-module="govuk-button" data-prevent-double-click="true">
+        Start now
+        <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" focusable="false" aria-hidden="true"><path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z"></path></svg>
+      </button>
```
tvararu, line 28: I guess we can't use the start button from govuk-components? https://govuk-components.netlify.app/components/start-button/#input-erb-start-button

Maybe not, since that's a link and we need an actual submit button in this form.

The button helpers do automatically create a form, but then they don't actually look like a start button https://govuk-components.netlify.app/helpers/button/#output-html-regular-button

Feels like an omission from the upstream `govuk-components` library. Should we raise an issue, or just submit a patch if it's not a lot of work?

I think either approach could work:

- allowing `govuk_start_button` to generate a form + be an actual button
- allowing `govuk_button_to` to look like a start button

Non-blocking.

tvararu (approved):

PR #189, opened 2022-07-04
Create review app

merrilmathew:
### Context

Review app deployment helps the latest changes on PR to be checked before merged to main. This PR adds the workflow to deploy a review app based on PR number and then delete workflow to clean up deployment when PR is closed.

### Trello card

https://trello.com/c/hgF6Kw5D

### DIFF
--- a/.github/workflows/delete-review-app.yml
+++ b/.github/workflows/delete-review-app.yml
@@ -1,0 +1,80 @@
+name: Delete Review App
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
+
+jobs:
+  delete-review-app:
+    name: Delete Review App ${{ github.event.pull_request.number }}
+    concurrency: deploy_review_${{ github.event.pull_request.number }}
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'deploy')
+    runs-on: ubuntu-latest
+    environment: review
+    steps:
+      - name: Extract configuration from tfvars
+        id: config
+        run: |
+          KEY_VAULT_NAME=$(jq -r '.key_vault_name' $TFVARS)
+          PAAS_SPACE=$(jq -r '.paas_space' $TFVARS)
+
+          if [ -z "$KEY_VAULT_NAME" ]; then
+            echo "::error ::Failed to extract key_vault_name from $TFVARS"
+            exit 1
+          fi
+
+          if [ -z "$PAAS_SPACE" ]; then
+            echo "::error ::Failed to extract paas_space from $TFVARS"
+            exit 1
+          fi
+
+          echo ::set-output name=key_vault_name::$KEY_VAULT_NAME
+          echo ::set-output name=paas_space::$PAAS_SPACE
+        shell: bash
+        env:
+          TFVARS: workspace_variables/review.tfvars.json
+        working-directory: terraform
+
+      - uses: Azure/login@v1
+        with:
+          creds: ${{ secrets.azure_credentials }}
+
+      - uses: Azure/get-keyvault-secrets@v1
+        id: get_secrets
+        with:
+          keyvault: ${{ steps.config.outputs.key_vault_name }}
+          secrets: "TFSTATE-CONTAINER-ACCESS-KEY,PAAS-USER,PAAS-PASSWORD"
+
+      - uses: hashicorp/setup-terraform@v1
+        with:
+          terraform_version: 1.0.10
+          terraform_wrapper: false
+
+      - name: Set Environment variables
+        run: |
+          PR_NUMBER=${{ github.event.pull_request.number }}
+          TF_STATE_FILE=review-pr-$PR_NUMBER.tfstate
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+          echo "TF_STATE_FILE=$TF_STATE_FILE" >> $GITHUB_ENV
+          pr_state_file=$(az storage blob list -c afqts-tfstate \
+            --aacount-key ${{ steps.get_secrets.outputs.TFSTATE-CONTAINER-ACCESS-KEY }} \
+            --prefix $TF_STATE_FILE --query "[].name" -o tsv)
+          if [ -n "$pr_state_file" ]; then echo "TF_STATE_EXISTS=true" >> $GITHUB_ENV; fi;
+
+      - name: Terraform
+        if: env.TF_STATE_EXISTS == 'true'
+        id: terraform
+        run: |
+          make ci review terraform-destroy
+        env:
+          ARM_ACCESS_KEY: ${{ steps.get_secrets.outputs.TFSTATE-CONTAINER-ACCESS-KEY }}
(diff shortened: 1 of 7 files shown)


terraform/variables.tf
```diff
@@ -65,7 +76,4 @@ locals {
     values(cloudfoundry_route.education)
   ])
 
-  apply_qts_app_name     = "apply-for-qts-in-england-${var.environment_name}"
-  postgres_database_name = "apply-for-qts-in-england-${var.environment_name}-pg-svc"
```
prabhakarreddy1234, line 69: I think we are already passing `environment_name` as  `review` from  review.tfvars.json.  These locals should build the names correctly for review app also. May be it's not needed to pass these as `TF_VARs` from make file into terraform for review app.   Any other reason for removing these 3 local vars ?

Makefile
```diff
@@ -37,6 +37,20 @@ production:
...
+review:
+	$(if $(pr_id), , $(error Missing environment variable "pr_id"))
+	$(eval DEPLOY_ENV=review)
+	$(eval AZURE_SUBSCRIPTION=s165-teachingqualificationsservice-development)
+	$(eval RESOURCE_NAME_PREFIX=s165d01)
+	$(eval ENV_SHORT=rv)
+	$(eval ENV_TAG=rev)
+	$(eval env=review-pr-$(pr_id))
+	$(eval backend_config=-backend-config="key=review/$(env).tfstate")
+	$(eval export TF_VAR_apply_qts_app_name=apply-for-qts-in-england-$(env))
```
prabhakarreddy1234, line 50: Probably these TF_VARs not needed since we build these names using `DEPLOY_ENV` value and tf local vars should build these names dynamically.  Is there any other reason for passing these vars from Makefile ?

merrilmathew: See above https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/189#discussion_r914640527

.github/workflows/deploy.yml
```diff
@@ -16,11 +16,20 @@ on:
   push:
     branches:
       - main
+
+  pull_request:
+    branches:
+      - main
+    types:
```
saliceti, line 23: Don't we need "opened" as well?

.github/workflows/delete-review-app.yml
```diff
@@ -0,0 +1,87 @@
...
+          PR_NUMBER=${{ github.event.pull_request.number }}
+          TF_STATE_FILE=review-pr-$PR_NUMBER.tfstate
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+          echo "TF_STATE_FILE=$TF_STATE_FILE" >> $GITHUB_ENV
+          pr_state_file=$(az storage blob list -c afqts-tfstate \
+            --aacount-key ${{ steps.get_secrets.outputs.TFSTATE-CONTAINER-ACCESS-KEY }} \
+            --prefix $TF_STATE_FILE --query "[].name" -o tsv)
+          if [ -n "$pr_state_file" ]; then echo "TF_STATE_EXISTS=true" >> $GITHUB_ENV; fi;
+
+      - uses: DFE-Digital/github-actions/setup-cf-cli@master
```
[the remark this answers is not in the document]
saliceti: This could also be removed from https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/.github/workflows/actions/deploy-environment/action.yml#L61

.github/workflows/deploy.yml
merrilmathew: Yes. will add that. thanks.

.github/workflows/delete-review-app.yml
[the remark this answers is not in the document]
merrilmathew: no we don't. Will remove.

.github/workflows/deploy.yml
```diff
@@ -72,6 +81,30 @@ jobs:
...
+    if: contains(github.event.pull_request.labels.*.name, 'deploy') || github.event_name != 'pull_request'
+    environment:
+      name: review
+      url: ${{ steps.deploy.outputs.environment_url }}
+
+    outputs:
+      environment_url: ${{ steps.deploy.outputs.environment_url }}
+
+    steps:
+      - uses: actions/checkout@v3
```
saliceti, line 98: This could be moved inside of the deploy action
And removed from https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/.github/workflows/deploy.yml#L89 as well

Makefile
saliceti: Can we pass only 1 variable like "app_suffix" instead? and build the app name, pg name and redis name inside of terraform

.github/workflows/deploy.yml
```diff
@@ -72,6 +82,30 @@ jobs:
...
 
+  deploy_review:
+    name: Deploy to review environment
+    concurrency: deploy_review_${{ github.event.pull_request.number }}
+    needs: [docker]
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'deploy') || github.event_name != 'pull_request'
+    environment:
+      name: review
+      url: ${{ steps.deploy.outputs.environment_url }}
```
saliceti, line 93: This will update the URL of the "review" environment which is shared between all the review apps, so the dev may not be able to see their app URL.
Can we remove it from here and display it as a PR comment instead? See: https://github.com/DFE-Digital/teaching-vacancies/blob/main/.github/workflows/build_and_deploy.yml#L266

.github/workflows/deploy.yml
```diff
@@ -72,6 +80,31 @@ jobs:
...
+          docker_image: ${{ needs.docker.outputs.docker_image }}
+          azure_credentials: ${{ secrets.AZURE_CREDENTIALS }}
+          pr_id: ${{ github.event.pull_request.number }}
+
+      - name: Post sticky pull request comment
+        if: github.event_name == 'pull_request'
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          message: |
+            Review app deployed to <https://apply-for-qts-in-england-review-pr-${{ github.event.pull_request.number }}.london.cloudapps.digital>
```
saliceti, line 106: Could we use the "environment_url" output from the action instead?

terraform/app.tf
```diff
@@ -92,6 +92,6 @@ resource "cloudfoundry_service_instance" "redis" {
 }
 
 resource "cloudfoundry_service_key" "redis_key" {
-  name             = "${local.redis_database_name}_key"
+  name             = "${var.redis_database_name}_key"
```
saliceti, line 95: Shouldn't we keep local.redis_database_name here?

terraform/workspace_variables/review.backend.tfvars
```diff
@@ -0,0 +1,3 @@
+storage_account_name = "s165d01afqtstfstatedv"
+
```
saliceti, line 2: Can we add a comment like "The key is provided dynamically for each review app via the Makefile"

saliceti (approved):
🎉

PR #194, opened 2022-07-05
Fix path after completed requirements

thomasleese:
This should be degree, otherwise we miss a question.

This isn't causing a problem in production because the question order enforcer is working to redirect users to the degree question.


thomasleese:
I reckon we can probably remove these methods from the forms, but I'll do the refactor in a separate PR.

David-Feetenby (approved):

PR #206, opened 2022-07-06
Swap teaching qualification and degree questions

David-Feetenby:
Swap the enforcement order
Update back links
Change path in the views
Repair tests where appropriate

### DIFF
--- a/spec/models/eligibility_check_spec.rb
+++ b/spec/models/eligibility_check_spec.rb
@@ -292,15 +292,19 @@ RSpec.describe EligibilityCheck, type: :model do
       it { is_expected.to eq(:qualification) }
     end
 
-    context "when degree is present" do
+    context "when qualification is present" do
       let(:attributes) do
-        { degree: true, completed_requirements: true, region: create(:region) }
+        {
+          qualification: true,
+          completed_requirements: true,
+          region: create(:region)
+        }
       end
 
-      it { is_expected.to eq(:teach_children) }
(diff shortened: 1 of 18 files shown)


spec/models/eligibility_check_spec.rb
```diff
@@ -289,15 +289,15 @@
         { completed_requirements: true, region: create(:region) }
       end
 
-      it { is_expected.to eq(:degree) }
+      it { is_expected.to eq(:qualification) }
     end
 
     context "when degree is present" do
       let(:attributes) do
         { degree: true, completed_requirements: true, region: create(:region) }
```
thomasleese, line 297:
```suggestion
        { qualification: true, completed_requirements: true, region: create(:region) }
```

As the next question is the qualification here we should change the test case, and then assert that the status becomes `degree`.

spec/models/eligibility_check_spec.rb
```diff
@@ -289,15 +289,15 @@
...
       end
 
-      it { is_expected.to eq(:qualification) }
+      it { is_expected.to eq(:teach_children) }
     end
 
     context "when qualification is present" do
```
thomasleese, line 303:
```suggestion
    context "when degree is present" do
```

Then this becomes the test for degree.

spec/models/eligibility_check_spec.rb
thomasleese: Otherwise we've got two tests that assert for `teach_children`.

PR #207, opened 2022-07-06
Add support for multiple teaching authority emails and websites

thomasleese:
This changes the model for regions so that teaching authorities can have multiple contact email address or websites. This was required for Romania, but it's possible we'll need this in the future so we should build it now.

There'll be a follow up PR to remove the old fields, but we can't deploy that until they're not being used.

[Trello Card](https://trello.com/c/wDXbRINE/589-support-console-allow-multiple-email-and-website-addresses-to-be-entered-into-country-specific-info)


tvararu (approved):
Nice work, good approach as well to making this a forwards only migration so we can clean up the old fields separately to shipping the new ones

PR #223, opened 2022-07-08
Add an ADR about storing files

thomasleese:
This hopefully answers all our questions about how we're going to store files, allow users to upload files, and model these in our database. It should give us the confidence to proceed with building the service knowing that when we come to support file storage the questions have already been asked and answered.


adr/00006-storing-files.md
```diff
@@ -0,0 +1,51 @@
...
+We will use [Sidekiq Cron] to schedule automated jobs which can delete expired files. We should also periodically look for files which have been orphaned (the associated model in the database has been deleted without the file being deleted) to ensure we are definitely not storing any personal data for longer than we need to.
+
+[azure blob storage]: https://azure.microsoft.com/en-gb/services/storage/blobs/
+[active storage]: https://edgeguides.rubyonrails.org/active_storage_overview.html
+[active record encryption]: https://edgeguides.rubyonrails.org/active_record_encryption.html
+[sidekiq cron]: https://github.com/ondrejbartas/sidekiq-cron
+
+## Consequences
+
+We're use the existing DfE Azure service to create containers for each environment our application runs in to store the files. We'll create access keys for these containers and store them in secrets so the application can access them.
```
tvararu, line 34:
```suggestion
We'll use the existing DfE Azure service to create containers for each environment our application runs in to store the files. We'll create access keys for these containers and store them in secrets so the application can access them.
```

PR #228, opened 2022-07-11
Add teaching authority contact details to countries

thomasleese:
This adds the address, websites and email addresses fields for teaching authorities from regions to the countries. We need this as some regions also have a national contact detail which we want to show our users.

[Trello Card](https://trello.com/c/306J3OJY/607-model-multiple-competent-authorities)


tvararu (approved):
Very clean, I like the new concern

PR #243, opened 2022-07-14
Build basic application journey

thomasleese:
This builds a basic journey for teachers which involves starting an application, signing up/signing in and then submitting an application. See individual commits for more details.

[Trello Card](https://trello.com/c/Otcav7OS/602-build-application-submission-process)

## Screenshots

Screenshot 2022-07-14 at 11 43 27
Screenshot 2022-07-14 at 11 48 40
Screenshot 2022-07-14 at 11 49 06
Screenshot 2022-07-14 at 11 49 24
Screenshot 2022-07-14 at 11 49 30
Screenshot 2022-07-14 at 11 49 35


tvararu:
I know these are WIP designs, but I think we should wrap the pages with the `two-thirds` component as 90% that will be a required design change. Plus it looks better :smile:

PR #253, opened 2022-07-20
Set up Active Storage

thomasleese:
This sets up the configuration for Active Storage as we're going to start using it to allow teachers to upload files and assessors to review the files.


tvararu (approved):
`image_processing` seems to require `imagemagick` and `libvips`: https://github.com/janko/image_processing#installation

We might need to add these to the Dockerfile deps https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/Dockerfile#L19

There's a few ways we can go about this:

1. Just add the deps without testing if they're needed, as they probably are, and are sensible/widely used packages
1. We can wait until we use this functionality and it breaks while we're testing, then we know for sure we need to add them; the risk here is that the person working on this might not know we didn't add the deps
1. We can build and run the Docker container locally and attempt to poke in the Rails console some of the `image_processing` helpers to be sure that the deps are needed and work correctly; Here's a snippet on how to build + run locally from the Find readme https://github.com/DFE-Digital/find-a-lost-trn/#docker

Also, `image_processing` doesn't seem directly related to the work in this PR and it could be PRed in later alongside the deps + code that exercises the functionality.

Approving regardless of which way we go

tvararu:
Follow-up thought: do we need to add `imagemagick` and `libvips` to `asdf` or otherwise mention them as deps in our README?

thomasleese:
> Also, `image_processing` doesn't seem directly related to the work in this PR and it could be PRed in later alongside the deps + code that exercises the functionality.

I think this is a fair comment, I've dropped the commit from this PR so we can merge it in and start building models using ActiveStorage.

PR #257, opened 2022-07-21
Build initial check your answers page

thomasleese:
This builds an initial version of the check your answers page, with further design improvements to come.

## Screenshots

Screenshot 2022-07-21 at 13 03 36

Screenshot 2022-07-21 at 13 03 31

### DIFF
--- a/app/views/teacher_interface/work_histories/_summary.html.erb
+++ b/app/views/teacher_interface/work_histories/_summary.html.erb
@@ -1,0 +1,53 @@
+<% work_histories.each do |work_history| %>
+  <% if work_history.current_or_most_recent_role? %>
+    <p class="govuk-body">Your current or most recent role</p>
+  <% else %>
+    <p class="govuk-body">Previous workplace</p>
+  <% end %>
+
+  <%= govuk_summary_list do |summary_list|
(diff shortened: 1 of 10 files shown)


app/views/teacher_interface/work_histories/_summary.html.erb
```diff
@@ -0,0 +1,53 @@
+<% work_histories.each do |work_history| %>
+  <% if work_history.current_or_most_recent_role? %>
+    <p class="govuk-body">Your current or most recent role</p>
+  <% else %>
+    <p class="govuk-body">Previous workplace</p>
+  <% end %>
+
+  <%= govuk_summary_list do |summary_list|
+    summary_list.row do |row|
```
tvararu, line 9: We could use a loop to create these rows with a bit less code. I don't feel strongly about that right now, just one to consider in the future if we end up having to make lots of similar changes to the rows.

tvararu (approved):

app/views/teacher_interface/work_histories/_summary.html.erb
thomasleese: Yeah, I agree it's something to refactor in the future.

PR #269, opened 2022-07-26
Improve design of "Check your answers" pages

thomasleese:
This makes two changes to the design of the "Check your answers" pages to match the prototype:

- Add a "Save and come back later" button to each question
- Use the summary list card style for the summary list

## Screenshots

Screenshot 2022-07-26 at 10 46 03

Screenshot 2022-07-26 at 10 46 10

### DIFF
--- a/app/components/check_your_answers_summary_component.rb
+++ b/app/components/check_your_answers_summary_component.rb
@@ -1,0 +1,37 @@
+class CheckYourAnswersSummaryComponent < ViewComponent::Base
+  def initialize(model:, title:, fields:)
+    super
+    @model = model
+    @title = title
+    @fields = fields
+  end
+
+  attr_reader :title
+
+  def rows
+    fields.map { |key, opts| opts ? row_for_field(key, opts) : nil }.compact
+  end
+
+  private
+
+  def row_for_field(key, opts)
+    {
+      key: opts.fetch(:key, key.to_s.humanize),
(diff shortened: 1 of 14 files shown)


app/components/check_your_answers_summary_component.rb
```diff
@@ -0,0 +1,37 @@
+class CheckYourAnswersSummaryComponent < ViewComponent::Base
```
tvararu, line 1: Is this component check your answers specific? Could it be a generic summary card component?

tvararu (approved):
Only thing to consider is whether we can make that component generic :+1: lgtm

app/components/check_your_answers_summary_component.rb
thomasleese: Perhaps `CheckYourAnswers` is the wrong prefix, but I wanted to make it clear that because it accepts a `model`, retrieves the values from it and hard codes a "Change" link, it's not _quite_ generic enough to just be a summary card component (i.e. if it was going to be generic I'd have built it in `govuk-components` and have the same API as the `govuk_summary_list` helper).

Either way, we need something which wraps the summary list card component and does the specific logic this component does.

Perhaps `ModelSummaryComponent` is good enough?

tvararu: I see, in that case, keep it like this :+1:

PR #289, opened 2022-07-29
Expose deployed commit

gpeng:
While we're deploying manually it's helpful to be able to quickly diff main HEAD with the currently deployed commit.

~~* Add /sha route that renders the currently deployed commit sha.~~

UPDATE: expose SHA instead to fix OKComputer

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -79,6 +79,7 @@ jobs:
           context: .
           push: false
           tags: ${{ steps.image.outputs.tag }}
+          build-args: SHA=${{ github.sha }}
 
       - name: Run Snyk to check Docker image for vulnerabilities
         if: github.actor != 'dependabot[bot]'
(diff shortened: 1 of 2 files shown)


thomasleese:
We've got OkComputer configured to show the app version:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/35eb5e058f904bf418183623d40339664cfc4d95/config/initializers/okcomputer.rb#L10

But I think it's not working as it expects a `SHA` env var, [looks like we can customise it by passing `env: "COMMIT_SHA"` though](https://github.com/sportngin/okcomputer/blob/d847ecbfb8ae9a56a58a392aa7dede07b29ce3c1/lib/ok_computer/built_in_checks/app_version_check.rb#L17).

gpeng:
Oh right. I didn't spot that 🤦  I'll change the env var to SHA and get rid of the other bit.

thomasleese (approved):

PR #294, opened 2022-08-01
User testing

David-Feetenby:
Generate test usernames for test access to the elibilility checker

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,8 @@ class ApplicationController < ActionController::Base
   include DfE::Analytics::Requests
 
   default_form_builder(GOVUKDesignSystemFormBuilder::FormBuilder)
+  before_action :authenticate_staff!,
+                unless: -> { FeatureFlag.active?(:service_open) }
 
   def current_user
     nil
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,10 +3,6 @@ class ApplicationController < ActionController::Base
 
   default_form_builder(GOVUKDesignSystemFormBuilder::FormBuilder)
 
-  http_basic_authenticate_with name: ENV.fetch("SUPPORT_USERNAME", "test"),
-                               password: ENV.fetch("SUPPORT_PASSWORD", "test"),
-                               unless: -> { FeatureFlag.active?(:service_open) }
-
   def current_user
     nil
   end
--- a/app/controllers/eligibility_interface/base_controller.rb
+++ b/app/controllers/eligibility_interface/base_controller.rb
@@ -3,6 +3,12 @@ module EligibilityInterface
     before_action :load_region
     after_action :save_eligibility_check_id
 
+    http_basic_authenticate_with name: ENV.fetch("TEST_USERNAME", "test"),
+                                 password: ENV.fetch("TEST_PASSWORD", "test"),
+                                 unless: -> {
+                                   FeatureFlag.active?(:service_open)
+                                 }
+
     def eligibility_check
       @eligibility_check ||=
         if session[:eligibility_check_id]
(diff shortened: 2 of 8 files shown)


app/controllers/eligibility_interface/base_controller.rb
```diff
@@ -26,6 +26,12 @@ def current_namespace
       "eligibility"
     end
 
+    http_basic_authenticate_with name: ENV.fetch("TEST_USERNAME", "test"),
+                                 password: ENV.fetch("TEST_PASSWORD", "test"),
+                                 unless: -> {
+                                   FeatureFlag.active?(:service_open)
+                                 }
```
thomasleese, line 33: I'd suggest putting this block near the top, below the `after_action` line, so it's easy to see immediately that this controller has authentication.

thomasleese (commented):
Code change looks good, I'd suggest adding something in the commit message that explains why we've removed the authentication from the `ApplicationController`.

> This moves the authentication from the application controller, as we've already got separate authentication on the support and teacher interfaces.

app/controllers/eligibility_interface/base_controller.rb
```diff
@@ -3,6 +3,12 @@ class BaseController < ApplicationController
     before_action :load_region
     after_action :save_eligibility_check_id
 
+    http_basic_authenticate_with name: ENV.fetch("TEST_USERNAME", "test"),
+                                 password: ENV.fetch("TEST_PASSWORD", "test"),
```
thomasleese, line 7:
```suggestion
    http_basic_authenticate_with name: ENV.fetch("TEST_USERNAME", ENV.fetch("SUPPORT_USERNAME", "test")),
                                 password: ENV.fetch("TEST_PASSWORD", ENV.fetch("SUPPORT_PASSWORD", "test")),
```

I'm thinking this might be better to do this so we only need to set the `TEST_USERNAME` and `TEST_PASSWORD` in the `test` environment and can leave the rest with the existing `SUPPORT_` environment variables.

app/models/feature_flag.rb
```diff
@@ -24,6 +24,13 @@ class FeatureFlag
       "Allow starting an application on this service directly after " \
         "completing an eligibility check.",
       "Thomas Leese"
+    ],
+    [
+      :test_user,
```
thomasleese, line 29:
```suggestion
      :staff_test_user_auth,
```

Just to make it slightly more explicit what this does.

app/controllers/application_controller.rb
```diff
@@ -2,10 +2,8 @@ class ApplicationController < ActionController::Base
   include DfE::Analytics::Requests
 
   default_form_builder(GOVUKDesignSystemFormBuilder::FormBuilder)
-
-  http_basic_authenticate_with name: ENV.fetch("SUPPORT_USERNAME", "test"),
-                               password: ENV.fetch("SUPPORT_PASSWORD", "test"),
-                               unless: -> { FeatureFlag.active?(:service_open) }
+  before_action :authenticate_staff!,
```
thomasleese, line 5: We should be aware that this does mean that users who authenticate using the "test" credentials _will_ have access to the support console while the feature flag is enabled, since they'll be authenticated as staff users. I think this is fine if we're happy for it to work this way, and we're aware of the need to turn off the flag while testing is not going on.

Also worth being aware (and perhaps documenting) that for this to work it will also require `staff_http_basic_auth` to be enabled, as it's only when that feature flag is enabled that the staff HTTP basic auth strategy is used (otherwise we use Devise ordinary username/password strategy against instances of the `Staff` model).

David-Feetenby: The test_user will have access to support if the flag is enabled and the open_service flag is disabled. This is explained in the description of the test_user flag and while not ideal, I don't think will cause any problems as long as we are aware.

thomasleese (approved):

gpeng (approved):

PR #313, opened 2022-08-03
Move rspec test job to deploy

merrilmathew:
### Context

Rspec tests should run prior to deployment. Deployment to environment
should have dependency on rspec tests passing. This commit modifies the
deploy workflow to include the rspec job and removes workflow that
previously ran concurrently to deploy workflow.


gpeng (approved):
I think this looks good to get the deploy dependent on the tests passing. 

Discussed offline: It would be good to work towards building the image and then running the tests once on the image, rather than on a fresh checkout of the code, before deploying through the pipeline. When we're in a 'pipeline' deploy from a PR merge we only need to run the tests once on the image and then just the smoke tests after successful deploy to each environment. For a manual deploy to an individual environment maybe we still do the tests again on the image. Also would be good to have alerting if the pipeline fails.

All of that doesn't have to be for this PR though ^. This is an improvement 👍

PR #314, opened 2022-08-04
Record traces in Sentry

thomasleese:
This allows us to use the performance monitoring features of Sentry to check how our service is doing. I've set the sample rate to 10% as a start, which we can adjust in the future as we see fit.


David-Feetenby (approved):

thomasleese:
Sentry performance monitoring is too expensive sadly.

PR #319, opened 2022-08-05
Run rspec against built docker image

merrilmathew:
Rspec should run against the docker image built in the build step. This
ensures that the build output is ready to deploy.


gpeng:
@merrilmathew have we settled on not doing this? I think that was what we discussed with @saliceti . Shall we close it?

gpeng:
Closing this as we're not going to do it for now.

PR #322, opened 2022-08-05
Refactor document uploads

thomasleese:
This refactors how document uploads are implemented to use the form pattern. The main benefit being that the logic for validation is better encapsulated and it means the uploads can be optional.

### DIFF
--- a/app/forms/teacher_interface/upload_form.rb
+++ b/app/forms/teacher_interface/upload_form.rb
@@ -1,0 +1,68 @@
+class TeacherInterface::UploadForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveStorageValidations
+
+  attr_accessor :document
+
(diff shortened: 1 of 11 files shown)


app/forms/teacher_interface/upload_form.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def blob
+    self
+  end
+
+  def byte_size
+    size
+  end
+end
```
thomasleese, line 68: I'd like to get this PR in before the user research tomorrow, but I'm going to write a card to change this as I don't like it.

I think it's better, since the validation we want is very specific, to not use the library and instead build our own basic validation on the content type and file size so we con't need to worry about making a `ActionDispatch::Http::UploadedFile` instance look like a `ActiveStorage::Blob` instance.

David-Feetenby (approved):

PR #325, opened 2022-08-08
Delete work history and qualifications

thomasleese:
This adds the ability for users to delete work history and qualification entries from their application form, if they decide they no longer need them. For qualifications, it's not possible to delete the teaching qualification, or the university degree if it's separate.

[Trello Card](https://trello.com/c/wVLuCnOY/712-get-application-form-ready-for-user-research)

## Screenshots

Screenshot 2022-08-08 at 08 32 30
Screenshot 2022-08-08 at 08 32 34

### DIFF
--- a/app/models/work_history.rb
+++ b/app/models/work_history.rb
@@ -65,4 +65,17 @@ class WorkHistory < ApplicationRecord
     application_form.work_histories.empty? ||
       application_form.work_histories.ordered.first == self
   end
+
+  def summary_title
+    school_name.presence || city.presence || country.presence || job.presence ||
+      I18n.t(
+        (
+          if current_or_most_recent_role?
+            "application_form.work_history.current_or_most_recent_role"
+          else
+            "application_form.work_history.previous_role"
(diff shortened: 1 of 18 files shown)


David-Feetenby (approved):

app/models/work_history.rb
```diff
@@ -65,4 +65,17 @@ def current_or_most_recent_role?
     application_form.work_histories.empty? ||
       application_form.work_histories.ordered.first == self
   end
+
+  def summary_title
```
gpeng, line 69: I think we should try and avoid having view logic in the models. Could this go in a view object (or helper?). Happy to approve as is if you want to card up a refactor given we need to this in.

gpeng (approved):

app/models/work_history.rb
thomasleese: Agreed, I've got a list of follow up improvements I'd like to make to these PRs after the user research is done and I'll add this to the list.

PR #338, opened 2022-08-10
Include eligibility checks without a region in performance

thomasleese:
This updates the "Usage by country" section of the performance dashboard to include eligibility checks where the country is not known to us and therefore the region is never selected and the user saw immediately an "ineligible" result. This allows us to better understand the latent demand for the service in the currently we don't currently recognise.


David-Feetenby (approved):

David-Feetenby:
[Work history back link broken](https://trello.com/c/EVNETnI0/731-work-history-back-link-broken)

PR #341, opened 2022-08-10
Specify GOVUK_NOTIFY_TEMPLATE_ID_DEVISE in Terraform

thomasleese:
This sets the environment variable via Terraform so we can customise it. The dev and review environments are going to use a template where personalisation is visible in the logs, whereas the other environments will use one where personalisation is not visible. This ensures that we don't leak any secret tokens in the logs.

[Trello Card](https://trello.com/c/qQtrd0R2/726-dont-log-login-token)

### DIFF
--- a/terraform/app.tf
+++ b/terraform/app.tf
@@ -11,6 +11,8 @@ locals {
     BIGQUERY_DATASET    = local.bigquery_dataset,
     BIGQUERY_TABLE_NAME = local.bigquery_table_name,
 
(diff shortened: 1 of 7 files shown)


thomasleese:
We've decided to go down a different route:

- In `dev` and `review` we'll use a team API key so emails _are_ sent out.
- We'll change the single Devise template to hide personalisation.
- Therefore we don't need two templates.

PR #343, opened 2022-08-10
Fix work_history back link

David-Feetenby:
Fix work_history back link in 'edit_has_work_history' by adding logic to check if the application_form is empty


thomasleese (approved):

David-Feetenby:
[Work history back link broken](https://trello.com/c/EVNETnI0/731-work-history-back-link-broken)

PR #350, opened 2022-08-11
Only show "Check your answers" if section is complete

thomasleese:
Currently we take the user to the "Check your answers" page if they've started a section, however this is confusing as we expect users will want to continue filling in the form.

This has also involved changing a few of the redirections as we can now explicitly send people to the check page where appropriate, and not otherwise.

[Trello Card](https://trello.com/c/fhqwhs5i/730-personal-the-link-should-open-the-form-if-the-section-is-in-progress-not-the-confirmation-page)


app/controllers/teacher_interface/personal_information_controller.rb
```diff
@@ -3,10 +3,17 @@ class PersonalInformationController < BaseController
     before_action :load_application_form
 
     def show
-      unless application_form.task_item_started?(
-               :about_you,
-               :personal_information
-             )
+      if application_form.task_item_completed?(
```
gpeng, line 6: This seems to be causing the user to get stuck in the form flow if it isn't complete. It isn't redirecting back to the task list at the end of the flow.

gpeng (changes requested):
This seems to be breaking the back links on the first form in each section.

app/controllers/teacher_interface/personal_information_controller.rb
thomasleese: I think this is because of a change I made in a separate PR which needed fixing for this PR: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/commit/47c0210eac7ade88cb72388bee1dfe8c39f45a2b#diff-2ea1bcadb9b63aea1a8ee920cc9af2f7ea3b4da1b57ae87cfe88c2e471ea5417R95

I've amended the commit and it should be working now.

gpeng (approved):

PR #351, opened 2022-08-11
Add translation rows to check summary

gpeng:
### Context

When a document is uploaded we allow the user to provide a translated version. So 'some' documents may also have a translation.

We need to display these in the `CheckYourAnswersSummaryComponent` as a separate row.

### Changes

* Check document type fields in `CheckYourAnswersSummaryComponent` and add a translation row to the output.
* Don't add a row if the document has no translations

### Review

* Run and add some translations. 
* Added translations should appear in the summary. 
* Deleting a translation file should remove the row from the summary. 
* Documents without a translation should not have an extra row

Screenshot 2022-08-11 at 11 28 47

### DIFF
--- a/app/components/check_your_answers_summary_component.rb
+++ b/app/components/check_your_answers_summary_component.rb
@@ -10,40 +10,59 @@ class CheckYourAnswersSummaryComponent < ViewComponent::Base
   attr_reader :title
 
   def rows
-    fields.map { |key, opts| opts ? row_for_field(key, opts) : nil }.compact
+    fields_with_translations.map { |field| row_for_field(field) }
   end
 
   private
 
-  def row_for_field(key, opts)
+  attr_reader :model, :fields
+
+  def fields_with_translations
+    fields_as_array.flat_map do |field|
+      has_translation?(field) ? [field, translation_row_for(field)] : field
+    end
+  end
+
+  def has_translation?(field)
+    object = model.send(field[:key])
+
+    return false unless object.is_a?(Document)
+    object.translated_uploads.any?
+  end
+
+  def row_title_for(field)
+    field.fetch(:title, field[:key].to_s.humanize)
+  end
+
+  def translation_row_for(field)
+    field.merge(translation: true, title: "#{row_title_for(field)} translation")
+  end
+
+  def fields_as_array
+    fields.filter_map { |key, options| options&.merge(key:) }
+  end
+
+  def row_for_field(field)
     {
-      key: opts.fetch(:key, key.to_s.humanize),
-      value: format_value(model.send(key), opts),
-      href: opts.fetch(:href)
+      key: field[:key],
+      title: row_title_for(field),
+      value: format_value(model.send(field[:key]), field),
+      href: field.fetch(:href)
     }
   end
 
-  def format_value(value, opts)
+  def format_value(value, field)
     return "" if value.nil?
 
     if value.is_a?(Date)
-      format = opts[:format] == :without_day ? "%B %Y" : "%e %B %Y"
+      format = field[:format] == :without_day ? "%B %Y" : "%e %B %Y"
       return value.strftime(format).strip
     end
 
-    if value.is_a?(Document)
-      return(
-        value
-          .uploads
-          .order(:created_at)
-          .map { |upload| upload.attachment&.filename }
-          .compact
-          .join(", ")
-      )
-    end
+    return file_names_for(value, field[:translation]) if value.is_a?(Document)
 
     if value.is_a?(Array)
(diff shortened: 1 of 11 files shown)


app/components/check_your_answers_summary_component.rb
```diff
@@ -10,11 +10,70 @@ def initialize(model:, title:, fields:, delete_link_to: nil)
...
+    fields_as_array.dup.tap do |fields_to_return|
+      translatable_field_indexes.reverse.each do |index|
+        translatable_item = fields_as_array[index]
+
+        fields_to_return.insert(
+          index + 1,
+          translation_row_for(translatable_item)
+        )
+      end
+    end
```
thomasleese, line 39:
```suggestion
    fields_as_array.flat_map do |field|
      if has_translation?(field)
        [field, translation_row_for(field)]
      else
        [field]
      end
    end
```

I think this makes the code easier to read, and avoids the need for `translatable_field_indexes`.

app/components/check_your_answers_summary_component.rb
```diff
@@ -52,5 +102,15 @@ def format_value(value, opts)
...
 
+  def file_names_for(document, translations)
+    scope =
+      translations ? document.translated_uploads : document.original_uploads
+    scope
+      .order(:created_at)
+      .map { |upload| upload.attachment&.filename }
+      .compact
+      .join(", ")
+  end
```
thomasleese, line 113:
```suggestion
  def file_names_for(document, translation)
    document.uploads
      .where(translation:)
      .order(:created_at)
      .filter_map { |upload| upload.attachment&.filename }
      .join(", ")
  end
```

Just shortens the code a little, but I don't mind either way.

I also wonder if we should copy the array field and have `.join("<br />").html_safe` but that's out of scope.

app/components/check_your_answers_summary_component.rb
```diff
@@ -10,11 +10,70 @@ def initialize(model:, title:, fields:, delete_link_to: nil)
...
   private
 
+  class Field
+    attr_reader :key, :options
+
+    def initialize(field_array)
+      @key = field_array[0]
+      @options = field_array[1]
+    end
+  end
```
thomasleese, line 27: I wonder if there's much benefit to having this class. I can see we needed a way of retaining the reference to the original key, but I wonder if it's better to model it as a hash with an extra field for the title, so something like this:

```rb
# original field
{
  key: "document",
}

# translated field
{
  key: "document",
  title: "document translations",
  translations: true
}
```

I think it would simplify some of this code, remove the need for `fields_as_array` and `translation_row_for`, and it's got the added benefit that we can support an optional `title` field for all the field types rather than just translated documents.

Getting the title of a field would have to change to:

```rb
opts.fetch(:title, opts.fetch(:key, key.to_s.humanize)))
```

I notice at the moment we've got a new method `row_title_for` in this PR, but we only use it for translations, whereas we could use it for all the fields as well if we wanted to.

Just my opinion though, I'm happy to keep the current implement to move forward.

thomasleese (commented):
Look good, I've left a few comments. It's not a deal breaker if we want to merge this in, but might simplify the code.

app/components/check_your_answers_summary_component.rb
thomasleese: There is a downside to this which is that everywhere we currently write `key: ...` when using the `CheckYourAnswersSummaryComponent` we'd have to change it to `title: ...` but actually I think this makes more sense than `key` anyway.

gpeng: The `Field` class could be a Hash as you suggest. We refactored it to that as originally we were passing the array around all over the place and accessing it with [index] and it was horrible. I'll have a look but object vs hash here is probably just a stylistic difference and we could do the other things with either. I think we still need `fields_as_array` as `fields` is a hash originally but we need to position the new rows after the translatable row which was simpler to do using the index of the array.

thomasleese: Yeah you're right about the `fields_as_array`, I sneakily edited my comment after writing it 😅

I think I still prefer my suggestion as it allows us to be consistent about the `title` and `key` and treat it the same for all types of rows, but I'm happy for you to ignore this suggestion too.

app/components/check_your_answers_summary_component.rb
gpeng: I think I'll leave this one as `.translated_uploads` and `.original_uploads` encapsulate the definition of 'a translation' in the document model rather than bleeding the database field into the view component. (we've arguably done that anyway with `created_at` 🤷)

thomasleese (approved):

thomasleese (approved):

PR #360, opened 2022-08-12
DQT integration spike

gpeng:
### Context

Apply for QTS needs to integrate with the DQT API to award qualified teacher status to successful applicants. The purpose of this spike is to try a simple integration with the existing API to highlight any data that we are missing that is required for DQT and to write some test data if possible.

The existing API requires a number of calls to award QTS.

* [Request a TRN](https://github.com/DFE-Digital/qualified-teachers-api/blob/main/docs/api-specs/v2.json#L452) this call creates the initial record in DQT and returns a new TRN. Duplicate checking is carried out during the execution of this call at the DQT end. If the data looks like it could be a duplicate then this is flagged and manual checks are done. So... the TRN return may be synchronous or it may take an indeterminate period of time. 
* [Retreive a TRN](https://github.com/DFE-Digital/qualified-teachers-api/blob/main/docs/api-specs/v2.json#L409) if a TRN isn't returned with initial request there is another endpoint that can be checked using the request ID that we supply in the first call. This can be polled until a TRN is returned
* [Recommend for award](https://github.com/DFE-Digital/qualified-teachers-api/blob/main/docs/api-specs/v2.json#L9) for this call we supply the assessment date and the endpoint synchronously returns an award date (assessment date + 1 day). 

### The code

This PR contains services to make each request using the data we have stored in the `Teacher`, `ApplicationForm` and `Qualification` models. In production these services would need to be linked together into a workflow of some description and split out into Sidekiq jobs which I haven't included in here. The current intention is for all of this to happen at the end of the assessment process.

### To run

NB: I haven't successfully run this code myself as yet as I'm getting a 401 unauthorised. I need to discuss that with @gunndabad 

With a DQT_API_KEY environment variable set:

* `Dqt::TrnRequest.call(application_form: <a complete application form>)` This should return a response like this: `{"requestId":"72888c5d-db14-4222-829b-7db9c2ec0dc3","status":"Completed","trn":"1234567"}`. If the response doesn't contain a `trn` then the requestId can be used to...
* `Dqt::RetrieveTrn.call(trn_request_id: <requestId from the previous step>)` the response to this call will be the same format as the previous one. In production we'd poll with this until we got a TRN

Once we have a TRN it needs to be set as `Teacher#trn` . That field is added to the `Teacher` model in this PR. This would be done as part of the workflow in production code.

* `Dqt::RecommendForAward.call(teacher: <teacher with TRN>, assessment_date: <a date in the past>)`. I expect this date would be when the assessor marks the application as approved. Assuming the teacher isn't prohibited from teaching by this point in DQT this will return an award date which is assessment date + 1 day. At this point the teacher has QTS.

### Things we need to work out

This implementation is based on the current API but there are changes planned that we'll need to incorporate. Trello card https://trello.com/c/iZEOLGbt/288-apply-for-qts-api-requirements Discussion: https://ukgovernmentdfe.slack.com/archives/C02JFQ8M3UY/p1659101313480789

* the API only allows one qualification. We allow the user to enter multiple degrees. In this PR we're just sending the first one we find that's not the ITT qual. Register sends the first too.
* Country of institution on the teaching qualification is collected as free text but the API expects the country code so we need to map the value the user enters to a code so need a pre-determined list of countries. One solution to this would be to make that field an autocomplete.
* We don't collect qualification 'subject'. We collect 'title' which is what I've used here but I'm not sure that's necessarily the subject
* We don't have anything we can derive the qualification 'class' from. [(the options the API accepts)](https://github.com/DFE-Digital/qualified-teachers-api/blob/main/docs/api-specs/v2.json#L576-L592). I'm passing 'NotKnown' in this code for now.
* Same for qualification type [(the options)](https://github.com/DFE-Digital/qualified-teachers-api/blob/main/docs/api-specs/v2.json#L842-L929) maybe 'DegreeEquivalent' or 'Unknown'
* For the TRN and the award we need to pass `ittProviderUkprn`. We don't have this and there won't be one. This is the only field that we can't work around at the moment.

### Other thoughts

Doing all of this at the very end of the process means that all of the duplicate checking that the API provides in the RegisterForTrn call will have to be done manually and the manual comparison will be between data in two systems. This doesn't seem ideal. It also runs the risk that when we do submit for TRN that the API still thinks there is a duplicate and refers the request for manual checking (which has already happened once). This is inefficient but it also will introduce an indeterminate delay between the assessor approving the application and QTS being awarded. Registering for TRN earlier would allow the automated duplicate checking to run and make the QTS award instant. That would mean we'd have to then mark the record as a 'Fail' outcome (or something like that) if the assessor did not pass the application. If the current portal writes directly to DQT on receipt of the application it may be there is already something in there to handle this.


app/services/dqt/trn_request_params.rb
```diff
@@ -0,0 +1,123 @@
...
+        date: qualification.certificate_date,
+        heQualificationType: "Unknown"  #https://github.com/DFE-Digital/qualified-teachers-api/blob/main/docs/api-specs/v2.json#L840-L930
+      }
+    end
+
+    def country_code_for(_country)
+      # TODO: mapping these. Probably need to make country an autocomplete
+      # register's mapping -> https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/lib/hesa/code_sets/countries.rb
+      "CA" # hard code Canada for now
+    end
```
thomasleese, line 113: We've got this list in the app already https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/public/location-autocomplete-canonical-list.json but it sounds like it might not line up with the ones in DQT, so we may need to maintain a separate one or change our country list to match DQT.

app/services/dqt/client.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Dqt
```
thomasleese, line 3: I wonder whether these files belong in `app/services` or `app/lib`. So far we've been putting miscellaneous "service" classes in `app/lib` so it might be better to be consistent with that, but I can see that these classes are more like proper "service" classes in that they interact with a service... so not sure!

thomasleese (commented):
Generally looks pretty good, I've just got two comments, but nothing major.

app/services/dqt/client.rb
gpeng: Yeah. They definitely don't belong where I've put them. I just stuck them all in the same directory while I was getting something working. I expect there'll be some services, some stuff in lib and a bunch of jobs when we build it properly.

PR #363, opened 2022-08-15
Application received email

thomasleese:
This PR adds a new mailer and associated email which is sent to a teacher when they submit their application containing a confirmation message and information on how they can return to the service.

[Trello Card](https://trello.com/c/UpJbp5Jx/751-send-an-email-to-the-applicant-when-an-application-has-been-submitted)


app/controllers/teacher_interface/application_forms_controller.rb
```diff
@@ -34,7 +34,14 @@ def edit
     end
 
     def update
-      @application_form.submit! if @application_form.can_submit?
+      if application_form.can_submit?
+        application_form.submit!
```
gpeng, line 38: Happy with this as is for now but what do you think about moving this sort of stuff into services. I'm not sure it's a massive benefit one way or the other here but just to establish a pattern of encapsulating this sort of 'multiple things going on' operation.

gpeng (approved):
This is fine dev wise. Could you screen shot an email so as the UCD team can product review from the PR?

app/controllers/teacher_interface/application_forms_controller.rb
thomasleese: I think it probably is worth doing, I'll amend this PR to include it.

PR #364, opened 2022-08-16
Add "teaching authority other" to regions

thomasleese:
This field, which we currently have on countries, is used to display generic information on how to contact the teaching authority which doesn't fit in any of the other fields. The reason we need it now is to explain that some region's websites are only accessible in the country.

[Trello Card](https://trello.com/c/z8AMhJ1c/763-add-teaching-authority-other-information-for-regions)

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -101,6 +101,7 @@
     - teaching_authority_emails
     - teaching_authority_websites
     - teaching_authority_name
+    - teaching_authority_other
     - application_form_enabled
   :staff:
     - id
--- a/spec/system/support_interface/countries_spec.rb
+++ b/spec/system/support_interface/countries_spec.rb
@@ -36,6 +36,7 @@ RSpec.describe "Countries support", type: :system do
     when_i_fill_teaching_authority_address
     when_i_fill_teaching_authority_emails
     when_i_fill_teaching_authority_websites
+    when_i_fill_teaching_authority_other
     and_i_save_and_preview
     then_i_see_the_preview
     and_i_see_a_success_banner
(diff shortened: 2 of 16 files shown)


config/analytics.yml
```diff
@@ -101,6 +101,7 @@
     - teaching_authority_emails
     - teaching_authority_websites
     - teaching_authority_name
+    - teaching_authority_other
```
gpeng, line 104: Do we need to update data form for this field? I wouldn't mind doing that with you if so as I've not done any of that before.

spec/system/support_interface/countries_spec.rb
```diff
@@ -159,6 +160,8 @@ def when_i_fill_teaching_authority_certificate
   end
 
   def when_i_fill_teaching_authority_other
+    fill_in "region-teaching-authority-other-field", with: "Other"
+  rescue Capybara::ElementNotFound
```
gpeng, line 164: Will this cause a default wait for the first case on line 163? I'm not sure, that might just be for js enabled tests. Would be good to avoid that if so.

config/analytics.yml
thomasleese: Yes, we will need to change this in dataform. I'll put something in the calendar to pair on it!

spec/system/support_interface/countries_spec.rb
thomasleese: I don't believe so, at least this is a pattern I've used in a few other places in the tests and I haven't noticed any slowdowns, but I don't know for sure.

gpeng (approved):

PR #366, opened 2022-08-16
Fix user email not found bug

David-Feetenby:
Remove insecure error mesage and replace with validation

### DIFF
--- a/app/controllers/teachers/sessions_controller.rb
+++ b/app/controllers/teachers/sessions_controller.rb
@@ -1,31 +1,46 @@
 # frozen_string_literal: true
-
 class Teachers::SessionsController < Devise::SessionsController
   include TeacherCurrentNamespace
 
   layout "two_thirds"
 
+  def new
+    @new_session_form = TeacherInterface::NewSessionForm.new
+  end
+
   def create
-    if resource_params[:create_or_sign_in] == "create"
-      redirect_to :eligibility_interface_countries
-      return
-    end
+    @new_session_form =
+      TeacherInterface::NewSessionForm.new(
+        email: new_session_form_params[:email],
+        create_or_sign_in: new_session_form_params[:create_or_sign_in]
+      )
 
-    self.resource = resource_class.find_by(email: resource_params[:email])
-
-    if resource
-      if resource.active_for_authentication?
-        resource.send_magic_link
-      else
-        resource.resend_confirmation_instructions
+    if @new_session_form.valid?
+      if @new_session_form.create?
+        redirect_to :eligibility_interface_countries
+        return
       end
 
+      self.resource = resource_class.find_by(email: @new_session_form.email)
+
+      if resource
+        if resource.active_for_authentication?
+          resource.send_magic_link
+        else
+          resource.resend_confirmation_instructions
+        end
+      end
       redirect_to :teacher_check_email
     else
-      set_flash_message(:alert, :not_found_in_database, now: true)
-      self.resource = resource_class.new(email: resource_params[:email])
-      render :new
+      render :new, status: :unprocessable_entity
     end
+  end
+
+  def new_session_form_params
+    params.require(:teacher_interface_new_session_form).permit(
+      :email,
+      :create_or_sign_in
+    )
   end
 
   def destroy
--- a/app/forms/teacher_interface/new_session_form.rb
+++ b/app/forms/teacher_interface/new_session_form.rb
@@ -1,0 +1,18 @@
+class TeacherInterface::NewSessionForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :email, :string
+  attribute :create_or_sign_in, :string
+
+  validates :create_or_sign_in, presence: true
+  validates :email, presence: true, valid_for_notify: true, if: -> { sign_in? }
+
+  def sign_in?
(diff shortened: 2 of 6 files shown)


app/controllers/teachers/sessions_controller.rb
```diff
@@ -1,33 +1,52 @@
...
+    @new_session_form =
+      TeacherInterface::NewSessionForm.new(
+        email: new_session_form_params[:email],
+        create_or_sign_in: new_session_form_params[:create_or_sign_in]
+      )
+
+    if @new_session_form.create?
       redirect_to :eligibility_interface_countries
-      return
+      return true
```
thomasleese, line 16:
```suggestion
      return
```

I don't think the `true` is necessary here.

app/controllers/teachers/sessions_controller.rb
```diff
@@ -1,33 +1,52 @@
...
     end
   end
 
+  def new
+    @new_session_form = TeacherInterface::NewSessionForm.new
+  end
+
+  def new_session_format
+    @valid_email ? true : false
+  end
```
thomasleese, line 41: I think we can get rid of this method.

thomasleese (commented):
Looks good, just got two final small comments!

app/forms/teacher_interface/new_session_form.rb
```diff
@@ -0,0 +1,13 @@
+class TeacherInterface::NewSessionForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :email, :string
+  attribute :create_or_sign_in, :string
+
+  validates :email, presence: true, valid_for_notify: true
```
thomasleese, line 8: I had some further thoughts about this validation, and I think we could do more here.

```rb
validates :create_of_sign_in, presence: true
```

This checks that the user at least checks one of the radio buttons, so we can show an error if they press submit without having checked one of the radio buttons.

```rb
validates :email, presence: true, valid_for_notify: true, if: -> { !create? }
```

The `if: -> { !create? }` block means that this validation will only run, if `create_of_sign_in == "sign_in"` (i.e. that the user has said they want to sign in).

Once this validation is in place, we could then call `.valid?` in the controller before calling `.create?` and `redirect_to : eligibility_interface_countries`.

app/controllers/teachers/sessions_controller.rb
```diff
@@ -1,33 +1,48 @@
...
-      set_flash_message(:alert, :not_found_in_database, now: true)
-      self.resource = resource_class.new(email: resource_params[:email])
-      render :new
+      render :new, status: :unprocessable_entity
     end
   end
 
+  def new
+    @new_session_form = TeacherInterface::NewSessionForm.new
+  end
```
thomasleese, line 37: I'd suggest moving this function so it's defined above `create`, it's not a big deal but it [matches the general Rails convention](https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions) so might make it faster to find the method for a new dev.

app/controllers/teachers/sessions_controller.rb
```diff
@@ -1,33 +1,48 @@
...
-    end
+    @new_session_form =
+      TeacherInterface::NewSessionForm.new(
+        email: new_session_form_params[:email],
+        create_or_sign_in: new_session_form_params[:create_or_sign_in]
+      )
 
-    self.resource = resource_class.find_by(email: resource_params[:email])
+    if @new_session_form.valid?
+      self.resource = resource_class.find_by(email: @new_session_form.email)
```
thomasleese, line 19: We can move this line below the `if @new_session_form.create?` block as it's not used yet (saves an unnecessary database query).

app/forms/teacher_interface/new_session_form.rb
```diff
@@ -0,0 +1,14 @@
+class TeacherInterface::NewSessionForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :email, :string
+  attribute :create_or_sign_in, :string
+
+  validates :create_or_sign_in, presence: true
+  validates :email, presence: true, valid_for_notify: true, if: -> { sign_in? }
```
thomasleese, line 9: I think we're missing a `sign_in?` method here (we're using one but it's not defined), I suggested in https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/366#discussion_r947512089 it would like something like:

```rb
def sign_in?
  create_or_sign_in == "sign_in"
end
```

app/controllers/teachers/sessions_controller.rb
```diff
@@ -1,33 +1,48 @@
...
-      redirect_to :teacher_check_email
+      self.resource = resource_class.find_by(email: @new_session_form.email)
+
+      if resource
+        if resource.active_for_authentication?
+          resource.send_magic_link
+        else
+          resource.resend_confirmation_instructions
+        end
+        redirect_to :teacher_check_email
```
thomasleese, line 32: Sorry, just one final spot, this line should go after the `end` below it, otherwise if we try and enter a valid email address that isn't in the database (i.e. `if resource` fails) we still want to take the user to the "check email" page as if it worked, to not expose the fact that the email address didn't exist.

thomasleese (approved):

PR #367, opened 2022-08-16
Update seeds and run in review

gpeng:
Update the seeds.rb file to match the production country lists. We need to seed review (and dev) environments and generate example data for them so we should have something approaching the production country config.

I've split this off from the example data work as it's useful on its own and can be extended with the other stuff.

On top of adding more 'production' state to the seeds file I've also amended the loop to skip regions that already exist to avoid overwriting existing configured regions.

Seems to be working ok -> https://apply-for-qts-in-england-review-pr-367.london.cloudapps.digital/support/countries

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,180 +6,705 @@
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
 
+# regenerate from production database with
+#
+# Country.all.to_h do |country|
+#   [
+#     country.code,
+#     country.regions.filter_map do |region|
+#       next if region.name.blank?
+#       {
+#         name: region.name,
+#         status_check: region.status_check,
+#         sanction_check: region.sanction_check,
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -191,11 +716,16 @@
...
   else
-    regions.each do |name|
-      country
-        .regions
-        .find_or_create_by!(name:)
-        .update!(legacy: false, application_form_enabled: true)
+    regions.each do |region|
+      next if country.regions.where(name: region[:name]).any?
+
+      country.regions.create!(
```
thomasleese, line 722:
```suggestion
      country.regions.create!(region)
```

Looks like we don't need to pass the individual hash values here.

thomasleese (approved):

db/seeds.rb
gpeng: That is a very good point... Fixed in [a2903ed](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/367/commits/a2903ede7ca086e13cf16ca965478e4b8d07f200)

PR #389, opened 2022-08-18
Add upper bound for date of birth validation

thomasleese:
At the moment we only check that the teacher is at least 18 years old, but we want to ensure the teacher is not really old as well. I've set it at 100 years old, but that's an arbitrary number.

### DIFF
--- a/app/forms/teacher_interface/name_and_date_of_birth_form.rb
+++ b/app/forms/teacher_interface/name_and_date_of_birth_form.rb
@@ -9,7 +9,9 @@ class TeacherInterface::NameAndDateOfBirthForm
   attribute :date_of_birth, :date
 
   validates :application_form, presence: true
-  validate :validate_date_of_birth
+  validates :date_of_birth,
+            allow_blank: true,
+            inclusion: 100.years.ago..18.years.ago
 
   def save
     return false unless valid?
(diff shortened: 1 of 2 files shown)


app/forms/teacher_interface/name_and_date_of_birth_form.rb
```diff
@@ -23,7 +23,8 @@ def save
   private
 
   def validate_date_of_birth
-    if date_of_birth.present? && date_of_birth >= 18.years.ago
+    if date_of_birth.present? &&
+         (date_of_birth >= 18.years.ago || date_of_birth <= 100.years.ago)
```
gpeng, line 27: Can we do `date_of_birth.between?`here? Also wondering if this could just be a standard rails inclusion validation?

thomasleese: Good point, I've refactored to use `inclusion`.

gpeng: Actually, will that set the dates from when the app last started? I don't think it really matters that much here.

thomasleese: Hmm true, I think it's okay, but something to keep an eye on...

gpeng (approved):

PR #391, opened 2022-08-18
Add instructions on how to install VIPS

thomasleese:
Also I've restructured the README to make it easier to read.

This was missed in b73027332bda5b2c60d6702f3d3e6ac1bc9c1d1b.


README.md
```diff
@@ -63,8 +65,17 @@ asdf plugin add azure-cli
 asdf install
 ```
 
-You'll also need to install PostgreSQL 13.x. The way to do this is different on
-each operating system, but on macOS you can try the following:
+#### VIPS
+
+You'll need to install `lipvips`. This various on each operating systen, but on macOS you can try this:
```
gpeng, line 70:
```suggestion
You‘ll need to install `lipvips`. This varies on each operating system, but on macOS you can try this:
```

gpeng (approved):

PR #392, opened 2022-08-18
Try installing glib

gpeng:
The deploy pipeline seems to be broken. The error suggests something, something, glib 🤷


gpeng:
Closing as fixed in https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/393

PR #396, opened 2022-08-19
[775] Create example users as 'confirmed'

gpeng:
Confirmation emails were being sent for each user created with the example_data:generate task. We don't want to disable the emails in review as we may want to test that functionality so create the example users as 'confirmed'. They would be in any case if they had started an application.

Also fix a deprecation warning from bin/lint

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-bin/bundle exec rubocop --auto-correct-all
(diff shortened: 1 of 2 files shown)


gpeng:
[Notify errors from example data generation](https://trello.com/c/zA1drqqg/775-notify-errors-from-example-data-generation)

thomasleese (approved):

PR #397, opened 2022-08-19
[778] Set correct `HostingEnvironment.host` for review

gpeng:
The `HostingEnvironment.host` is used to generate the login email URLs. For all other envs the env name is the prefix and the URL is an `education.gov.uk` one but review apps are prefixed with the application name and have a cloud apps URL so need to be treated differently.

The mails are sent from a worker instance so we need to handle that if we go with this method.

This will fix the login and confirmation email links

Screenshot 2022-08-19 at 16 27 56
.

### DIFF
--- a/app/lib/hosting_environment.rb
+++ b/app/lib/hosting_environment.rb
@@ -1,37 +1,51 @@
 module HostingEnvironment
-  def self.name
-    ENV.fetch("HOSTING_ENVIRONMENT", "dev")
-  end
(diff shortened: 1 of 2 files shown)


thomasleese (approved):

gpeng:
[Review app email links are incorrect](https://trello.com/c/2KQCpFB5/778-review-app-email-links-are-incorrect)

PR #412, opened 2022-08-23
[800] Basic application show page

gpeng:
Basic placeholder page for the application overview at `/assessor/application_forms/<id>`

This also fixes the failing build on `main` due to a missing `require` in the index page system spec.

Screenshot 2022-08-23 at 17 55 20

PR #413, opened 2022-08-24
Application form search result component

thomasleese:
This adds a new component which renders an application form search result in the list of applications. To extract the information from the application forms I've put that in a helper so it can be used in other components (for example a summary card component).

Depends on #412 for the link in the search result.

## Screenshot

Screenshot 2022-08-24 at 11 57 27

### DIFF
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -5,6 +5,7 @@ $govuk-assets-path: "/";
 @import "govuk-frontend/govuk/all";
 @import "location-autocomplete.min";
 
+@import "_description_list";
 @import "_environments";
 @import "_summary_list_card";
 @import "_support";
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -5,13 +5,13 @@ $govuk-assets-path: "/";
 @import "govuk-frontend/govuk/all";
 @import "location-autocomplete.min";
 
+@import "_application_forms";
 @import "_description_list";
 @import "_environments";
 @import "_search_results";
 @import "_summary_list_card";
 @import "_support";
 @import "_task_list";
-@import "_applications";
 
 ul.autocomplete__menu {
   li {
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -7,6 +7,7 @@ $govuk-assets-path: "/";
 
 @import "_description_list";
 @import "_environments";
+@import "_search_results";
 @import "_summary_list_card";
 @import "_support";
 @import "_task_list";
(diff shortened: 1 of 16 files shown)


app/assets/stylesheets/application.sass.scss
```diff
@@ -5,11 +5,13 @@ $govuk-assets-path: "/";
...
 @import "location-autocomplete.min";
 
+@import "_description_list";
 @import "_environments";
+@import "_search_results";
 @import "_summary_list_card";
 @import "_support";
 @import "_task_list";
-@import "_applications";
+@import "application_forms";
```
gpeng, line 14: Should this be `_application_forms`?

app/helpers/application_form_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+      [
+        I18n.t("application_form.summary.country"),
+        CountryName.from_country(application_form.region.country)
+      ],
+      [I18n.t("application_form.summary.region"), application_form.region.name],
+      [
+        I18n.t("application_form.summary.created_at"),
+        application_form.created_at.strftime("%e %B %Y")
+      ],
+      [I18n.t("application_form.summary.days_remaining_sla"), "17"],
```
gpeng, line 17: I wonder if we should set this to 'Not implemented' or something like that as I expect this might not get looked at for at least a few sprints and it'll look like we've implemented it at a casual glance if we use a number.

app/views/assessor_interface/application_forms/index.html.erb
```diff
@@ -1,22 +1,23 @@
-<% content_for :page_title, "Application forms" %>
+<% content_for :page_title, "Applications" %>
```
gpeng, line 1: I'm struggling with this 'Applications' vs 'Application forms' thing 🤣

app/helpers/application_form_helper.rb
```diff
@@ -0,0 +1,35 @@
+module ApplicationFormHelper
```
gpeng, line 1: I generally prefer going with view objects rather than using helpers like this. Happy to discuss and not do that as I realise using helpers is pretty established. It just feels cleaner to me to have things encapsulated in an object rather than a module that gets included all over the place. No naming issues, you only have to pass the object in once rather than to every method 🤷

app/helpers/application_form_helper.rb
thomasleese: Sounds reasonable, I'll update it!

app/helpers/application_form_helper.rb
thomasleese: Yeah, I do generally agree with you. I've been trying to stick with standard Rails patterns until they become unmaintainable/hard to scale before switching to more bespoke patterns. I decided to go with helpers here as they're supported natively in view components and it didn't quite feel big enough to cause problems. I also think it would be better to not have them automatically included in each controller and include them manually, which we could configure Rails to do.

I'm happy to switch to view objects, do we have a consensus over where the files should be put in the repo? I also wonder if the existing helpers should be refactored to use view objects at the same time so I wonder about doing that in a separate PR - what do you think?

gpeng: Ok cool. Let's merge this and refactor them all in another PR. I'll card that up.

gpeng (approved):

PR #419, opened 2022-08-24
[816] Create application form earlier

gpeng:
* Create an application form when a user first registers rather than when they confirm their email.

This means that if they confirm on another browser/machine then they won't have to redo their eligibility check.

### Review

I've not added any new tests here for this change. The existing system spec checks that the application is present when the user logs in. As this hasn't changed the tests still provide coverage I think. I could maybe add a test that they don't have to redo the eligibility check but I'm not sure there's much value in that.


thomasleese (approved):
Looks good to me!

PR #420, opened 2022-08-25
[802][SPIKE]  Application form status management

gpeng:
Applications will go through various states from their initial creation to an end state, likely award or decline. Many of these states e.g. 'Submitted' or 'Declined' will be set as a result of a user action. So... 'explicitly' set. Some states may be set by the system e.g. 'Expired' so explicitly set but not by a user.

We intend to make 'an assessment' a series of steps that will also have some sort of state. The state of the application may be set by a combination of the sub-states of the application sections. This is an emerging design and isn't fully formed yet but I've included some code that would handle that.

### Storing state 'on the application form' 

We could use an enum or simple state machine on the `ApplicationForm` model initially. I did consider this but whether we do that or not really depends on if we decide we need to keep a history of state changes in the service. We've discussed adding the Audited gem to keep an audit of changes so a combination of an enum + Audited would give us that and be much simpler. We will also have a history of state changes in BigQuery but that won't be accessible from within the service. The main reason for keeping it in the service that I can see at the moment is the requirement for SLA calculations. The state of the application form is crucial for working that out and the rules for SLA calculation are quite fluid currently so keeping all of the state changes will allow us to refine the rules with all of the data required in service.

The code here still stores a state string on the application form. This is largely redundant but we have requirements for state filtering in combination with other filters and it will greatly simplify the queries for that. It will also simplify the analytics queries in BigQuery.

### Transition rules

I've split these out into separate classes that will be simple to test and easy to understand. I expect we could do the same with rules for setting implicit states which would again make them easily testable and scalable. I'm not sure of the requirements there so I haven't included that, just some comments where it would go.

### Timeline events

We have designs for displaying a timeline. I've included a simple `TimelineEvent` class that we would use for this to show how it would run alongside the state. This may meet the need for a history of events within the service by itself in some ways but based on experience with similar functionality in other DfE services there's a good chance we'll end up wanting to log state changes that we don't want to surface in the timeline so keeping this separate will avoid that and doesn't add too much complexity.

### Using a state machine gem

There are loads and I had a look over some that I haven't used before. We could get some of the functionality here and lots of extra features (many of which we'd also get with enums) but didn't seem to have much by way of options for storing state history separately.

### Not covered in this spike

We will likely want to attribute events and possibly state changes to a user. I've added a staff_id column to the events table but not tackled how we pass the user into the model layer. We also have two types of user so in reality this may need to be polymorphic.

### Recommendations

I think something like the approach sketched out here would be flexible but probably only worth the bother if we decide we definitely need a state history in the service. I feel like we probably do given the SLA calculation requirement. It would be possible to derive this from some sort of TimelineEvent implementation or from Audited audit data but that seems like overloading those things to me.

Given the requirements for SLA etc are still in flux we could also start with an enum implementation or similar and expand to something like this when the requirements are clearer.

### DIFF
--- a/app/lib/application_form_state.rb
+++ b/app/lib/application_form_state.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+module TransitionRules
+  class Awarded
+    def self.can_transition?(application_form:, to:)
+      if application_form.current.state == ApplicationFormState::AWARDED &&
+           to == ApplicationFormState::DECLINED
+        return false
+      end
+
+      true
+    end
+  end
+
+  class Declined
+    def self.can_transition?(application_form:, to:)
+      if application_form.current.state == ApplicationFormState::DECLINED &&
+           to == ApplicationFormState::AWARDED
+        return false
+      end
+
+      true
+    end
+  end
+end
+
+class ApplicationFormState
+  SUBMITTED = "submitted"
+  AWARDED = "awarded"
+  DECLINED = "declined"
+  REVIEW_COMPLETE = "review_complete"
+
+  TRANSITION_RULES = [
+    TransitionRules::Awarded,
+    TransitionRules::Declined
+  ].freeze
+
+  def initialize(application_form:)
+    @application_form = application_form
+  end
+
+  def current
+    # should this always call update first to pick up any implicit changes?
+    state_changes.last
+  end
+
+  # explicit state changes ie from a user action like submitting the application or awarding
+  def change(to:)
+    raise "invalid state transition" unless can_transition?(to:)
+
+    ApplicationForm.transaction do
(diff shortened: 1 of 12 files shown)


gpeng:
[Application form 'status'](https://trello.com/c/5roToVD7/802-application-form-status)

gpeng:
We've reviewed this spike. The outcome:

* extract `TimelineEvent` and implement that now
* use a string based enum for state on ApplicationForm initially which is all we need to implement the planned work for the next sprint or two at least
* expand to this state history stuff when we're sure we need it for a feature, possibly the SLA calculations
* not add the audited gem as we have an audit trail in BigQuery (maybe we'll review this in the future though as non-web request driven updates won't be attribtued in BigQuery.

PR #421, opened 2022-08-25
[821] Add TimelineEvent

gpeng:
* Add class for recording events with a title, annotiation and association to a user (polymorphic for teacher or staff)

We'll use this class to log user interactions with the application form.

I've added a title field to the original plan as the current timeline designs all have a title. We could probably infer this from the type in most cases so happy to drop that 🤷 There are other requirements in the timeline but this should cover the basic case we have now.

We have discussed storing extra data in the events such as references to the assigned user etc. I haven't covered that here as yet. 

Screenshot 2022-08-25 at 11 13 29

### DIFF
--- a/app/models/timeline_event.rb
+++ b/app/models/timeline_event.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: timeline_events
+#
+#  id                  :bigint           not null, primary key
+#  annotation          :string           default(""), not null
+#  creator_type        :string
+#  event_type          :string           not null
+#  created_at          :datetime         not null
+#  updated_at          :datetime         not null
+#  application_form_id :bigint           not null
+#  creator_id          :integer
+#
+# Indexes
+#
+#  index_timeline_events_on_application_form_id  (application_form_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (application_form_id => application_forms.id)
+#
+class TimelineEvent < ApplicationRecord
+  belongs_to :application_form
+  belongs_to :creator, polymorphic: true, optional: true
+
+  enum event_type: { assessor_assigned: "assessor_assigned" }
+end
--- a/db/migrate/20220824172503_create_timeline_events.rb
+++ b/db/migrate/20220824172503_create_timeline_events.rb
@@ -1,0 +1,12 @@
+class CreateTimelineEvents < ActiveRecord::Migration[7.0]
+  def change
+    create_table :timeline_events do |t|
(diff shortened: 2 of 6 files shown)


app/models/timeline_event.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (application_form_id => application_forms.id)
+#
+class TimelineEvent < ApplicationRecord
+  belongs_to :application_form
+  belongs_to :creatable, polymorphic: true, optional: true
+
+  EVENT_TYPES = [ASSESOR_ASSIGNED = "assessor_assigned"].freeze
```
thomasleese, line 29: We could model this as an `enum`:

```rb
enum event_type: { assessor_assigned: "assessor_assigned" }
```

I don't feel too strongly either way.

app/models/timeline_event.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+#  index_timeline_events_on_application_form_id  (application_form_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (application_form_id => application_forms.id)
+#
+class TimelineEvent < ApplicationRecord
+  belongs_to :application_form
+  belongs_to :creatable, polymorphic: true, optional: true
```
thomasleese, line 27:
```suggestion
  belongs_to :creator, polymorphic: true, optional: true
```

Perhaps this would be better, but it does go against the convention of `able` for polymorphic fields. I'd suggest it shouldn't be optional though until we find there's a need for it.

db/migrate/20220824172503_create_timeline_events.rb
```diff
@@ -0,0 +1,13 @@
+class CreateTimelineEvents < ActiveRecord::Migration[7.0]
+  def change
+    create_table :timeline_events do |t|
+      t.string :event_type, null: false
+      t.references :application_form, foreign_key: true
```
thomasleese, line 5:
```suggestion
      t.references :application_form, null: false, foreign_key: true
```

Sadly the default is `null: true`.

db/migrate/20220824172503_create_timeline_events.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string :title, null: false
+      t.string :annotation, null: true
```
thomasleese, line 7:
```suggestion
      t.string :annotation, null: false, default: ""
```

I prefer this for strings as it means there aren't two ways of representing a blank value (`nil` and `""`). Instead the blank value can only be an empty string.

thomasleese (approved):

PR #423, opened 2022-08-25
[823] ApplicationForm.status -> .state

gpeng:
* Add a state column with a default of 'draft' (should we set this using default on the enum instead? 🤷 )
* Drop the status column
* Changes states to draft, submitted, awarded, declined

This is the beginnings of what will ultimately become a more complex state machine. The default state is changed to `draft` as `active` probably more accurately describes a grouping of most of the other states so we can use that as a scope that covers everything except `draft` (and maybe `declined` or `expired` etc)

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -14,7 +14,7 @@
 #  has_work_history        :boolean
 #  reference               :string(31)       not null
 #  registration_number     :text
-#  status                  :string           default("active"), not null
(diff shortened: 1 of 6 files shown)


gpeng:
[Store application form state](https://trello.com/c/5KkVntFy/823-store-application-form-state)

thomasleese (approved):

PR #424, opened 2022-08-25
Add names to staff users

thomasleese:
We want to show the name of staff users instead of their email address, so we need to add this field to the staff model.

[Trello Card](https://trello.com/c/Q9y1PGyw/822-add-names-to-staff-users)

### DIFF
--- a/app/controllers/staff/invitations_controller.rb
+++ b/app/controllers/staff/invitations_controller.rb
@@ -3,34 +3,14 @@ class Staff::InvitationsController < Devise::InvitationsController
 
   layout "two_thirds"
 
+  before_action :configure_permitted_parameters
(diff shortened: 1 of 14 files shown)


gpeng (approved):
I'm dithering over whether we should go with first_name last_name. I don't suppose it really matters for our use cases currently. Splitting them up for the four users in the future won't be too taxing either.

thomasleese:
> I'm dithering over whether we should go with first_name last_name. I don't suppose it really matters for our use cases currently. Splitting them up for the four users in the future won't be too taxing either.

I think we should avoid it unless we discover a real need for it. I did consider calling the field `full_name` but I think `name` is okay, it's the only "name" that makes sense for a staff user I'd say.

PR #428, opened 2022-08-26
Country filter

thomasleese:
This adds a new filter to the sidebar for viewing the application forms allowing users to filter on the country.

Depends on #427.

[Trello Card](https://trello.com/c/xBMsb0Bj/805-application-filter-country)

## Screenshot

Screenshot 2022-08-26 at 07 43 59

### DIFF
--- a/spec/system/assessor_interface/filtering_application_forms_spec.rb
+++ b/spec/system/assessor_interface/filtering_application_forms_spec.rb
@@ -10,8 +10,11 @@ RSpec.describe "Filtering application forms", type: :system do
     when_i_am_authorized_as_an_assessor_user
     when_i_visit_the_applications_page
 
-    and_i_apply_the_name_filter
+    when_i_apply_the_name_filter
     then_i_see_a_list_of_applications_filtered_by_name
+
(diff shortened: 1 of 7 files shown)


spec/system/assessor_interface/filtering_application_forms_spec.rb
```diff
@@ -37,11 +40,21 @@ def then_i_see_a_list_of_applications_filtered_by_name
     expect(page).to have_content("Cher Bert")
   end
 
+  def and_i_apply_the_country_filter
+    fill_in "Country", with: "France"
+    click_button "Apply filters"
+  end
+
+  def then_i_see_a_list_of_applications_filtered_by_country
+    expect(page).to have_content("Cher Bert")
```
gpeng, line 49: Should we create an application specifically for each filter? Just thinking this expectation would pass if we accidentally submitted the previous one's filter inputs.

thomasleese: Yep, that's a good point, I've amended the commit with that change.

gpeng (approved):

PR #429, opened 2022-08-26
[806] Assessor filter

gpeng:
Add assessor filter to the application forms page.

There are some styling issues with the form labels vs the prototype and we need to think about how values are cleared from the form as that seems a bit buggy but we can address that separately.

Screenshot 2022-08-26 at 09 26 49

### DIFF
--- a/app/controllers/assessor_interface/application_forms_controller.rb
+++ b/app/controllers/assessor_interface/application_forms_controller.rb
@@ -1,5 +1,6 @@
 class AssessorInterface::ApplicationFormsController < AssessorInterface::BaseController
   def index
+    @assessors = Staff.all
     @application_forms =
       ::Filters::ALL.reduce(ApplicationForm.active) do |scope, filter|
         filter.apply(scope:, params:)
--- a/app/lib/filters.rb
+++ b/app/lib/filters.rb
@@ -1,3 +1,3 @@
 module Filters
-  ALL = [Name, Country].freeze
+  ALL = [Name, Assessor, Country].freeze
 end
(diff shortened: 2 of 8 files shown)


app/controllers/assessor_interface/application_forms_controller.rb
```diff
@@ -1,5 +1,6 @@
 class AssessorInterface::ApplicationFormsController < AssessorInterface::BaseController
   def index
+    @assessors = Staff.all
```
thomasleese, line 3: I've been thinking about this for the state filter, and I wondered about having a method on the filter itself.

Something like:

```rb
class Filters::Assessor
  def options
    Staff.all.map { |staff| [staff.id, staff.name] }
  end
end
```

And then in the view something like:

```erb
<%- Filters::Assessor.new(@application_forms).options.each do |id, label| -%>
```

My thinking behind this was that for my state filter, we want to show the counts of the number of application forms in each state on each checkbox, and for that we need access to the `scope` so we can show the count based on the existing filters. This might allow it to be done in a consistent way for the assessor filter and the state filter, or I can refactor when I build the state filter.

thomasleese (approved):

app/controllers/assessor_interface/application_forms_controller.rb
gpeng: 🤔 Not sure. Let's have a chat about it. I initially thought it was a good idea but it does mix the concerns of the filter up a bit with view stuff.

gpeng: Discussed offline. Decided to use a view object for this stuff and keep the filters for filterin'

PR #431, opened 2022-08-26
Add an ADR about view objects

thomasleese:
We think it would be approriate to start using view objects in the app so we'd like to record this as a decision since it affects the architecture.

### DIFF
--- a/adr/00008-view-objects.md
+++ b/adr/00008-view-objects.md
@@ -1,0 +1,49 @@
+# 8. Application forms
+
+Date: 2022-08-26
+
+## Status
+
+Accepted
+
+## Context
+
+As the application begins to grow, we've started to put view logic in various places (controllers, helpers, and models) which often this means the logic cannot be easily reused in other places. It's not always obvious where the logic should belong, and there are problems with each location:
+
+- View logic in controllers means they become tightly coupled to the views, and the logic is hard to reuse.
+- Helpers are global and modules which makes them hard to unit test and can introduce namespacing issues.
+- Models shouldn't contain view logic otherwise they can end up too big to maintain.
+
+## Decision
+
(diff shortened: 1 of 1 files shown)


adr/00008-view-objects.md
```diff
@@ -0,0 +1,49 @@
...
+- Helpers are global and modules which makes them hard to unit test and can introduce namespacing issues.
+- Models shouldn't contain view logic otherwise they can end up too big to maintain.
+
+## Decision
+
+View objects present a solution to this problem by providing a testable, self-contained class which can take objects from the database layer and transform them approriately for the view layer.
+
+We will introduce the concept of view objects to the application and use them whenever a view becomes big enough that the logic should be tested on its own. This is a subjective decision but can be discussed as part of the code review process and therefore hopefully won't cause too much contention. Some rules of thumb might be: logic that is used in multiple places, formatting of the underlying database fields, fetching more than one model, etc.
+
+The view objects will be stored in an `app/view_objects` directory and will be classes which take data (models, request params, etc) and transform them appropriately for the view. The controller will instantiate a view object and pass it to the view. View objects may also be used in components in a similar way.
```
gpeng, line 23: Are we happy for view objects to retrieve data themselves too? I think probably yes although it's sometimes contentious. I think of them as a model that the view is bound to and so serves the view rather than our AR models which are bound to the database and serve that. So the view object utilises AR to populate itself from the views point of err... view.

thomasleese: Good point, yes I'm completely fine with that, I'll update that sentence.

thomasleese: I've amended the sentence, but also changed the example so it's clearer that this might be a way of implementing a view object.

gpeng (approved):

PR #432, opened 2022-08-26
Add ApplicationFormsIndexViewObject

thomasleese:
This is a view object which represents the application forms index view and managed all the business logic associated with that view.

I was considering moving the filters to be classes within the `ApplicationFormsIndexViewObject` as they deal with the `params` in the same way, but I decided against it in the end as that's not strictly required for this refactor, but it might be something to consider in the future.


gpeng (approved):
Perfect. Well, the name's horrible but I see where you're going with that. It would be good to have `ViewObjects::AssessorInterface::ApplicationFormIndex` but that's beyond my quick try at it without some horrible folder structure. Not sure how you do it without that (or if it's worth the bother trying). I suppose the convention with controllers is to stick the controller bit on the end 🤷

PR #433, opened 2022-08-26
Add filtering on application form status

thomasleese:
This adds the ability for a user to filter the application forms on the state, where the checkboxes include a count of each application form so users can see how many application forms there are in each state.

Depends on #432 

[Trello Card](https://trello.com/c/u97RpiY5/807-application-filter-status)

## Screenshot

Screenshot 2022-08-26 at 14 02 21

### DIFF
--- a/app/lib/filters/state.rb
+++ b/app/lib/filters/state.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class Filters::State < Filters::Base
+  def apply
+    return scope if states.empty?
+
+    scope.where(state: states)
+  end
+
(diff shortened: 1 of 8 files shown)


gpeng (approved):
I'm going to open a PR changing them all to `module Filters` and we can have a battle of wits 😆

thomasleese:
> I'm going to open a PR changing them all to `module Filters` and we can have a battle of wits 😆

Haha sorry, I didn't like that the `RSpec.describe` wasn't the first line of all the `_spec` files :joy:

PR #435, opened 2022-08-26
Add text that shows if there are no results

thomasleese:
This adds a piece of text for if there are no search results when looking for applications.

[Trello Card](https://trello.com/c/Yf6mb6RQ/824-no-results-content-for-applicationform-index)

## Screenshot

Screenshot 2022-08-26 at 15 03 18


gpeng (approved):
We need to ask Wes about this. I don't think the content is correct. Fine for now though.

thomasleese:
> We need to ask Wes about this. I don't think the content is correct. Fine for now though.

It was actually Wes who came up with the content! 😅

PR #436, opened 2022-08-26
Order search results newest to oldest

thomasleese:
The newest application form will show at the top of the list, with the oldest appearing at the bottom.

[Trello Card](https://trello.com/c/w9PFD0Ag/826-order-applicationform-results)


app/view_objects/assessor_interface/application_forms_index_view_object.rb
```diff
@@ -11,7 +11,7 @@ def application_forms
     ::Filters::State.apply(
       scope: application_forms_without_state_filter,
       params:
-    )
+    ).order(created_at: :desc)
```
gpeng, line 14: This is fine for now but thinking about it we're going to want submitted_at which we don't have. I'll card that up.

gpeng (approved):

PR #441, opened 2022-08-30
Ensure clearing country box resets filter

thomasleese:
At the moment [there's a bug with the accessible autocomplete component](https://github.com/alphagov/accessible-autocomplete/issues/229) where clearing the box doesn't get recorded correctly and therefore the previously selected value continues to be sent to the server in the `POST` request.

To get around this, we need to include the original text in the form data and use that to determine whether someone is trying to clear the autocomplete component.

We also need to consider users who have JavaScript disabled and are therefore shown just the select box, and for that we allow the original `location` param to be used if the `location_autocomplete` param isn't present.

### DIFF
--- a/app/view_objects/assessor_interface/application_forms_index_view_object.rb
+++ b/app/view_objects/assessor_interface/application_forms_index_view_object.rb
@@ -5,7 +5,7 @@ class AssessorInterface::ApplicationFormsIndexViewObject
   include Pagy::Backend
 
   def initialize(params:)
-    @params = params
+    @params = remove_cleared_autocomplete_values(params)
   end
 
   def application_forms_pagy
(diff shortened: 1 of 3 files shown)


app/view_objects/assessor_interface/application_forms_index_view_object.rb
```diff
@@ -5,7 +5,13 @@ class AssessorInterface::ApplicationFormsIndexViewObject
   include Pagy::Backend
 
   def initialize(params:)
-    @params = params
+    @params =
+      params.tap do |p|
```
gpeng, line 9: Could we split this out into a method so as we can describe what's happening here... e.g `def remove_cleared_autocomplete_values` or something like that?

thomasleese: Yep, good idea, I've amended the commit with that.

gpeng (approved):

PR #452, opened 2022-08-31
[831] Assign assessor

gpeng:
Add assign assessor UI.

Screenshot 2022-08-31 at 17 48 11

Screenshot 2022-08-31 at 17 54 04

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,7 +13,7 @@ Rails.application.routes.draw do
 
     resources :application_forms, path: "/applications", only: %i[index show] do
       get "assign-assessor", to: "assessor_assignments#new"
-      patch "assign-assessor", to: "assessor_assignments#update"
+      post "assign-assessor", to: "assessor_assignments#create"
     end
   end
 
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -11,7 +11,10 @@
   namespace :assessor_interface, path: "/assessor" do
     root to: redirect("/assessor/applications")
 
-    resources :application_forms, path: "/applications", only: %i[index show]
+    resources :application_forms, path: "/applications", only: %i[index show] do
+      get "assign-assessor", to: "assessor_assignments#new"
+      post "assign-assessor", to: "assessor_assignments#create"
```
thomasleese, line 16:
```suggestion
      resource :assessor_assignment, only: %i[new create], on: :member
```

I tend to prefer using as it abstracts away the routes and does a few things automatically, but I don't feel strongly either way.

thomasleese (approved):

config/routes.rb
gpeng: Happy to change but I can't get it to keep the URLs that we specified in the card doing it that way. We'd have to have `assign-assessor/new` If we're going to do that we may as well go with `assessor-assignments`. I don't really mind that, it just seemed to read better as `assign-assessor`

gpeng: Discussed offline. Leaving as is for now to preserve the URL.

PR #454, opened 2022-09-01
Pageobject Spike

David-Feetenby:
Example Pageobjects/test for using pageobject to refactor from capybara

### DIFF
--- a/spec/support/page_objects/eligiblity_interface/start.rb
+++ b/spec/support/page_objects/eligiblity_interface/start.rb
@@ -1,0 +1,15 @@
+module PageObjects
+  class Start < Base
+    set_url "/eligibility"
+
+    element :heading, "h1"
+
+    element :start_button, "button"
+
+    section :footer, "footer" do
+      element :accessibility_link, ".govuk-footer__link", text: "Accessibility"
+      element :cookies_link, ".govuk-footer__link", text: "Cookies"
+      element :privacy_link, ".govuk-footer__link", text: "Privacy"
+    end
+  end
+end
--- a/spec/system/eligibility_spec.rb
+++ b/spec/system/eligibility_spec.rb
@@ -207,7 +207,7 @@ RSpec.describe "Eligibility check", type: :system do
   end
 
   def when_i_press_start_now
-    click_button "Start now"
+    start_page.start_button.click
   end
 
   def when_i_press_start
@@ -237,7 +237,7 @@ RSpec.describe "Eligibility check", type: :system do
   end
 
   def when_i_visit_the_start_page
-    visit "/eligibility"
+    start_page.load
   end
 
   def when_i_try_to_go_to_the_eligible_page
@@ -275,13 +275,9 @@ RSpec.describe "Eligibility check", type: :system do
   end
 
   def then_i_see_the_start_page
-    expect(page).to have_content(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
-    expect(page).to have_current_path("/eligibility/start")
-    expect(page).to have_title(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
+    expect start_page.heading.text.to eq(
+             "Check your eligibility to apply for qualified teacher status (QTS) in England"
+           )
   end
 
   def then_i_see_the_misconduct_page
@@ -421,5 +417,9 @@ RSpec.describe "Eligibility check", type: :system do
 
   def then_i_have_two_eligibility_checks
     expect(EligibilityCheck.count).to eq(2)
+  end
+
+  def start_page
+    @start_page ||= PageObjects::Start.new
   end
 end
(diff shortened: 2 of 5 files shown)


David-Feetenby:
[PageObject spike](https://trello.com/c/TcTa8G4Z/688-pageobject-spike)

spec/support/page_objects/eligiblity_interface/start.rb
```diff
@@ -0,0 +1,15 @@
+module PageObjects
+  class Start < PageObjects::Base
```
thomasleese, line 2:
```suggestion
  class Start < Base
```

As we're within the `PageObjects` module we shouldn't need this here.

spec/system/eligibility_spec.rb
```diff
@@ -237,7 +237,8 @@ def when_i_select_a_country_with_multiple_regions
   end
 
   def when_i_visit_the_start_page
-    visit "/eligibility"
+    start_page.load
+    # visit "/eligibility"
```
thomasleese, line 241: I think we can remove this line!

spec/system/eligibility_spec.rb
```diff
@@ -275,13 +276,8 @@ def then_i_do_not_see_the_start_page
...
   def then_i_see_the_start_page
-    expect(page).to have_content(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
-    expect(page).to have_current_path("/eligibility/start")
-    expect(page).to have_title(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
+    expect
+    start_page.heading.text.to eq "Check your eligibility to apply for qualified teacher status (QTS) in England"
```
thomasleese, line 280:
```suggestion
    expect(start_page.heading.text).to eq("Check your eligibility to apply for qualified teacher status (QTS) in England")
```

Looks like the `expect` is across multiple lines so I'm not sure it would actually do the expectation.

thomasleese (commented):
Nice improvement! I've never used `SitePrism` before, but it looks like it'll really improve the tests.

I've left a few comments on small things, nothing major. Also I've spotted there's a `.DS_Store` file committed which we shouldn't have in the repo.

spec/system/eligibility_spec.rb
```diff
@@ -275,13 +275,8 @@ def then_i_do_not_see_the_start_page
...
 
   def then_i_see_the_start_page
-    expect(page).to have_content(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
-    expect(page).to have_current_path("/eligibility/start")
-    expect(page).to have_title(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
+    expect
```
gpeng, line 278: We should use brackets here and then split the line on the `.to`

spec/support/page_objects/eligiblity_interface/start.rb
```diff
@@ -0,0 +1,15 @@
+module PageObjects
+  class Start < Base
+    set_url "/eligibility"
+
+    element :heading, "h1"
+
+    element :sign_in, ".app-start-page-banner__button"
```
gpeng, line 7: We should probably call this `:signin_button` or `:start_button` So it'll ready `start_page.start_button.click` in the specs.

gpeng (changes requested):

spec/system/eligibility_spec.rb
```diff
@@ -275,13 +275,9 @@ def then_i_do_not_see_the_start_page
...
 
   def then_i_see_the_start_page
-    expect(page).to have_content(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
-    expect(page).to have_current_path("/eligibility/start")
-    expect(page).to have_title(
-      "Check your eligilibity to apply for qualified teacher status (QTS) in England"
-    )
+    expect start_page.heading.text.to eq(
```
gpeng, line 278: This is still a bit wonky. I think it should match the previous code from the diff.

David-Feetenby: prettier removed the brackets around start_page.heading.text

gpeng: It won't work like that without the brackets as `.to` is a method on the return from `expect` not the string that `.text` is.

David-Feetenby: Prettier allows the brackets only if '.to eq(' are on the line above

gpeng: I've just run prettier on my suggestion and it doesn't change it. I think your VS Code prettier plugin may be using different config or something. Try with `bin/lint` in the console. This is this the diff I have locally that prettier is fine with.
```
   def then_i_see_the_start_page
-    expect start_page.heading.text.to eq(
-             "Check your eligibility to apply for qualified teacher status (QTS) in England"
-           )
+    expect(start_page.heading.text).to eq(
+      "Check your eligibility to apply for qualified teacher status (QTS) in England"
+    )
   end
```

gpeng: Could be the space you're leaving after `expect` maybe. 🤷

David-Feetenby: Yeah it was the space

gpeng (approved):

PR #458, opened 2022-09-01
Add links to check submitted details

thomasleese:
This adds the necessary links to the items in the "Check submitted details" section of the assessor interface application form assessment page.


gpeng (approved):
I think this is good for now. I'm wondering if we'll end up implementing each of these pages as a new/edit `AssessmentSection` or something like that with an `Assessment` having many `AssessmentSection`-ses. No need to worry about that now though, just thinking 😁

PR #460, opened 2022-09-01
Add links to documents in check your answers

thomasleese:
When rendering a check your answers summary component, when displaying a document with a list of uploads we want to include a link to the upload so teachers can see what they uploaded and assessors can see the documents. 

### DIFF
--- a/app/components/check_your_answers_summary/component.rb
+++ b/app/components/check_your_answers_summary/component.rb
@@ -63,7 +63,7 @@ module CheckYourAnswersSummary
         return value.strftime(format).strip
       end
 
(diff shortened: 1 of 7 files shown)


gpeng (approved):
I think this is ok initially. Have we set up authentication and expiration for the links somewhere? I can't see it but I don't know much about it in Azure. It does support it with SAS tokens.

thomasleese:
> I think this is ok initially. Have we set up authentication and expiration for the links somewhere? I can't see it but I don't know much about it in Azure. It does support it with SAS tokens.

My understanding is that we get this automatically, but I'll check today.

PR #461, opened 2022-09-01
Add option for hiding change links to "Check your answers"

thomasleese:
This adds a new option for the "Check your answers" component to hide the "Change" links, so this component can be used in both the teacher interface (where we do want the links) and in the assessor interface (where we don't want the links).

### DIFF
--- a/app/components/check_your_answers_summary/component.html.erb
+++ b/app/components/check_your_answers_summary/component.html.erb
@@ -19,11 +19,14 @@
         <div class="govuk-summary-list__row">
           <dt class="govuk-summary-list__key"><%= row.fetch(:title) %></dt>
           <dd class="govuk-summary-list__value"><%= row.fetch(:value) %></dd>
-          <dd class="govuk-summary-list__actions">
-            <%= govuk_link_to(row.fetch(:href)) do %>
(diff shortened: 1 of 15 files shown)


gpeng (commented):
Should we use `editable` rather than `can_change`? Also what do you think about maybe about showing/hiding the link based on if there is a href or not? That would avoid needing the extra argument although it's less explicit.

thomasleese:
> Should we use `editable` rather than `can_change`?

Initially I had `changeable` but it looked wrong 😅

I'm happy to change it to `editable`, but actually my thinking is this attribute will only be here until we've refactored this to use view objects where the view object will generate a list with or without hrefs...

> Also what do you think about maybe about showing/hiding the link based on if there is a href or not?

I did think about it, and I reckon we probably should move towards this way of doing it, but my thinking behind this PR is that it unblocks all of the cards related to these check pages without needing a larger refactor, and we can do the refactor anyway when we work on the card to implement view objects in place of helpers/partials.

thomasleese:
> Should we use `editable` rather than `can_change`?

@gpeng I've renamed it to `changeable` (I prefer the term `change` since the links say "Change").

gpeng (approved):

PR #462, opened 2022-09-01
Add check qualifications page

thomasleese:
This adds a new page which assessors will use to check the qualifications of the application. For now it just displays the qualifications and a button which takes the user back to the overview page.

Depends on #461, #460, and #458.

[Trello Card](https://trello.com/c/fqAMJgnC/837-check-qualifications)

## Screenshot

Screenshot 2022-09-01 at 17 49 37

### DIFF
--- a/spec/system/assessor_interface/checking_submitted_details_spec.rb
+++ b/spec/system/assessor_interface/checking_submitted_details_spec.rb
@@ -1,0 +1,63 @@
+require "rails_helper"
+
+RSpec.describe "Assessor check submitted details", type: :system do
+  before do
+    given_the_service_is_open
+    given_an_assessor_exists
+    given_there_is_an_application_form
+    given_i_am_authorized_as_an_assessor_user(assessor)
+  end
+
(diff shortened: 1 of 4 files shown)


spec/system/assessor_interface/checking_submitted_details_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    then_i_see_the_qualifications
+
+    when_i_click_continue
+    then_i_see_the_application_page
+  end
+
+  private
+
+  def given_the_service_is_staff_http_basic_auth
+    FeatureFlag.activate(:staff_http_basic_auth)
```
gpeng, line 22: Should we start testing with this off?

gpeng (approved):

spec/system/assessor_interface/checking_submitted_details_spec.rb
thomasleese: Yep good point, I've amended those tests to log in as an assessor user.

PR #463, opened 2022-09-01
Add check work history page

thomasleese:
This adds a new page which assessors will use to check the work history of the application. For now it just displays the work history and a button which takes the user back to the overview page.

Depends on #461, #460, and #458.

[Trello Card](https://trello.com/c/s4cLoZir/838-check-work-history)

### DIFF
--- a/app/controllers/assessor_interface/check_work_histories_controller.rb
+++ b/app/controllers/assessor_interface/check_work_histories_controller.rb
@@ -1,0 +1,11 @@
+module AssessorInterface
+  class CheckWorkHistoriesController < BaseController
+    def show
+      @application_form =
+        ApplicationForm.includes(:work_histories).find(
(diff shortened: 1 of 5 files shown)


gpeng (approved):

app/controllers/assessor_interface/check_work_histories_controller.rb
```diff
@@ -0,0 +1,11 @@
+module AssessorInterface
+  class CheckWorkHistoriesController < BaseController
+    def show
+      @application_form =
+        ApplicationForm.includes(:qualifications).find(
```
gpeng, line 5: Do we need :work_histories here?

thomasleese: Yes, we do!

PR #475, opened 2022-09-05
Add application timeline page

thomasleese:
This adds a new page for viewing the timeline of an application, specifically it renders the timeline events using the MoJ timeline component as per the prototype.

It's a basic view for now to satisfy the user need to view the timeline, but there's plenty of opportunities for improvements.

[Trello Card](https://trello.com/c/RDM1WGwh/850-basic-timeline-page)

## Screenshot

Screenshot 2022-09-05 at 17 56 41


app/controllers/assessor_interface/timeline_events_controller.rb
```diff
@@ -0,0 +1,15 @@
+module AssessorInterface
+  class TimelineEventsController < BaseController
+    def index
+      @application_form =
+        ApplicationForm.includes(:timeline_events).find(
+          params[:application_form_id]
+        )
+
+      @timeline_events =
```
gpeng, line 9: Should we get the timeline events from the `@application_form` given we included the data in that query?

config/locales/assessor_interface.en.yml
```diff
@@ -13,3 +13,13 @@ en:
...
             first_assessment: First assessment
             second_assessment: Second assessment
+    timeline_events:
+      index:
+        title:
+          assessor_assigned: Assessor assigned
+          reviewer_assigned: Reviewer assigned
+          state_changed: Status changed
+        description:
+          assessor_assigned: Assessor assigned to %{assignee_name}.
```
gpeng, line 23: I don't think this reads very well. Maybe we could check with Wes but the user is being assigned as assessor|reviewer to the application.

app/controllers/assessor_interface/timeline_events_controller.rb
thomasleese: Ah, I need to remove the `includes(:timeline_events)` actually, I discovered that due to the `created_at: :desc` it executes another query anyway, and I figured two queries for this page isn't too bad (it's not an N+1 issue).

config/locales/assessor_interface.en.yml
thomasleese: Good point, I left these as a kind of placeholder until the real content/design for this page is done but I'll check with Wes.

thomasleese: Wes has suggested an alternative, I've amended the commit!

gpeng: Poetry 👏

gpeng (approved):

PR #484, opened 2022-09-08
Add an ADR on assessments

thomasleese:
This adds an ADR which described our approach to implementing the assessments. The bulk of this ADR is pseudocode that was developed during a spike in to how we might implement this, which may be a little unusual for an ADR, but I think it conveys the information in a useful way.

[Trello Card](https://trello.com/c/R9PMJPmY/851-spike-assessment-sections-data-structure)

### DIFF
--- a/adr/00009-assessments.md
+++ b/adr/00009-assessments.md
@@ -1,0 +1,135 @@
+# 9. Assessments
+
+Date: 2022-09-08
+
+## Status
+
+Accepted
+
+## Context
+
+We are starting to build the assessor interface for this application, and one of the priorities is to decide on a sensible data model and application structure for modelling the assessment criteria and checks that must be performed on a submitted application form to award or decline QTS.
+
+The design and user research on this part of the system is still being designed, but we understand enough to get started, and we should design the architecture in a way that is flexible enough to support future decisions.
+
+## Decision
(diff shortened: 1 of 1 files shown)


adr/00009-assessments.md
```diff
@@ -0,0 +1,133 @@
...
+# application form and the country/region.
+class InitialAssessment < ApplicationRecord
+  enum section: %i[
+         personal_details
+         qualifications
+         work_history
+         professional_standing
+       ]
+
+  # checked - a nullable boolean holding whether a user has changed this part of the application form
```
gpeng, line 77: I forget why we wanted this to be nullable? Doesn't defaulting to false work? Also should `a user has changed` be `a user has checked`

gpeng: Ah yeah. I remember now. Yes, I think checked is misleading. checked: false implies the section hasn't been 'checked' rather than it has but it wasn't 'passed' or 'approved'. Something like that maybe? Would using an enum be more descriptive?

thomasleese: I think you're right an enum would be better, I've amended the commit with an enum. I still don't love the names, but I guess this ADR doesn't commit us to the exact naming.

PR #485, opened 2022-09-08
[843] Page object refactor for the assessor interface specs

gpeng:
* Update the assessor interface system specs to use page objects
* Add PageHelpers module to centralise the creation of page objects.

I've not updated any other specs to use the PageHelpers module. I'll do that in a separate PR if this approach is merged.


thomasleese (approved):
Looks great, really nice improvement on the tests!

PR #488, opened 2022-09-09
Create section assessments model and factory

thomasleese:
Following on from https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/484 which starts to implement the decisions made, specifically the creation of an `SectionAssessment` model and updates the `AssessmentFactory` class.

Depends on #487.

[Trello Card](https://trello.com/c/R9PMJPmY/851-spike-assessment-sections-data-structure)

### DIFF
--- a/app/lib/assessment_factory.rb
+++ b/app/lib/assessment_factory.rb
@@ -8,10 +8,21 @@ class AssessmentFactory
   end
 
   def call
-    Assessment.create!(application_form:)
+    sections = section_keys.map { |key| AssessmentSection.new(key:) }
+    Assessment.create!(application_form:, sections:)
   end
 
   private
 
   attr_reader :application_form
+
+  def section_keys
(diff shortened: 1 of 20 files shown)


gpeng (commented):
This looks good. Not sure about the section, section_assessments naming though.

thomasleese:
> This looks good. Not sure about the section, section_assessments naming though.

Yeah, I'm really struggling with this name. In the ADR I had put `InitialAssessment` but I don't like that as it if there's no further information then it's not just the initial assessment. I figured `SectionAssessment` worked because it's an "assessment" of a "section", but I don't like that we've got the term assessment in two places in the app now.

Alternatively we could have `SectionCheck`, but then we've got checks inside the section checks (unless we call them checkpoints 🤷🏻‍♂️).

gpeng (approved):

PR #493, opened 2022-09-12
Teacher interface refactor

David-Feetenby:
refactor teacher interface rspecs to use pageobjects

Where granular access of elements was more tricky, I opted for a simpler refactor. 
Some PageObjects are almost identical to others, but i opted to keep them separate where I thought future changes to the pages likely could need distinct objects.

### DIFF
--- a/spec/support/page_helpers.rb
+++ b/spec/support/page_helpers.rb
@@ -1,6 +1,15 @@
 module PageHelpers
-  def application_page
-    @application_page ||= PageObjects::AssessorInterface::Application.new
+  def then_i_see_the(page, **args)
+    expect(send(page.to_sym)).to be_displayed(**args)
+  end
+
+  def when_i_visit_the(page, **args)
+    send(page.to_sym).load(**args)
+  end
+
+  def assessor_application_page
+    @assessor_application_page ||=
+      PageObjects::AssessorInterface::Application.new
   end
 
   def applications_page
@@ -93,11 +102,119 @@ module PageHelpers
     @timeline_page ||= PageObjects::AssessorInterface::Timeline.new
   end
 
-  def then_i_see_the(page, **args)
-    expect(send(page.to_sym)).to be_displayed(**args)
+  def teacher_sign_in_page
+    @teacher_sign_in_page = PageObjects::TeacherInterface::SignIn.new
   end
 
-  def when_i_visit_the(page, **args)
-    send(page.to_sym).load(**args)
+  def teacher_sign_up_page
+    @teacher_sign_up_page = PageObjects::TeacherInterface::SignUp.new
+  end
+
+  def document_form_page
+    @document_form_page = PageObjects::TeacherInterface::DocumentForm.new
+  end
+
+  def check_uploaded_files_page
+    @check_uploaded_files_page =
+      PageObjects::TeacherInterface::CheckUploadedFiles.new
+  end
+
+  def check_email_page
+    @check_email_page = PageObjects::TeacherInterface::CheckYourEmail.new
+  end
+
+  def new_application_form_page
+    @new_application_form_page =
+      PageObjects::TeacherInterface::NewApplicationForm.new
+  end
+
+  def signed_out_page
+    @signed_out_page = PageObjects::TeacherInterface::SignedOut.new
+  end
+
+  def teacher_application_page
+    @teacher_application_page = PageObjects::TeacherInterface::Application.new
+  end
+
+  def new_teacher_form_page
+    @new_teacher_form_page =
+      PageObjects::TeacherInterface::NewApplicationForm.new
+  end
+
+  def subjects_form_page
+    @subjects_form_page = PageObjects::TeacherInterface::SubjectsForm.new
+  end
+
+  def work_history_form_page
+    @work_history_form_page = PageObjects::TeacherInterface::WorkHistoryForm.new
+  end
+
+  def delete_confirmation_page
+    @delete_confirmation_page =
+      PageObjects::TeacherInterface::DeleteConfirmationForm.new
+  end
+
+  def written_statement_page
+    @written_statement_page =
+      PageObjects::TeacherInterface::WrittenStatement.new
+  end
+
+  def personal_information_summary_page
+    @personal_information_summary_page =
+      PageObjects::TeacherInterface::PersonalInformationSummary.new
+  end
+
+  def qualification_summary_page
+    @qualification_summary_page =
+      PageObjects::TeacherInterface::QualificationSummary.new
+  end
+
+  def qualifications_form_page
+    @qualifications_form_page =
+      PageObjects::TeacherInterface::QualificationsForm.new
+  end
+
+  def name_and_date_of_birth_page
+    @name_and_date_of_birth_page =
+      PageObjects::TeacherInterface::NameAndDateOfBirth.new
+  end
+
+  def check_your_answers_page
+    @check_your_answers_page =
+      PageObjects::TeacherInterface::CheckYourAnswers.new
+  end
+
+  def work_history_summary_page
+    @work_history_summary_page =
+      PageObjects::TeacherInterface::WorkHistorySummary.new
+  end
+
+  def upload_document_page
+    @upload_document_page =
+      PageObjects::TeacherInterface::UploadDocumentForm.new
+  end
+
+  def submitted_application_page
+    @submitted_application_page =
+      PageObjects::TeacherInterface::SubmittedApplication.new
+  end
+
+  def check_your_uploads_page
+    @check_your_uploads_page =
+      PageObjects::TeacherInterface::CheckYourUploads.new
+  end
+
+  def alternative_name_page
+    @alternative_name_page =
+      PageObjects::TeacherInterface::AlternativeNameForm.new
+  end
+
+  def age_range_form
+    @age_range_page = PageObjects::TeacherInterface::AgeRangeForm.new
+  end
+
+  def registration_number_form
+    @registration_number_form =
+      PageObjects::TeacherInterface::RegistrationNumberForm.new
   end
 end
--- a/spec/support/system_helpers.rb
+++ b/spec/support/system_helpers.rb
@@ -126,12 +126,18 @@ module SystemHelpers
   end
 
   def then_i_see_the_sign_in_form
-    expect(page).to have_title(
+    expect(teacher_sign_in_page).to have_title(
       "Apply for qualified teacher status (QTS) in England"
     )
-    expect(page).to have_content("Have you used the service before?")
-    expect(page).to have_content("Yes, sign in and continue application")
-    expect(page).to have_content("No, I need to check my eligibility")
+    expect(teacher_sign_in_page).to have_content(
+      "Have you used the service before?"
+    )
+    expect(teacher_sign_in_page).to have_content(
+      "Yes, sign in and continue application"
+    )
+    expect(teacher_sign_in_page).to have_content(
+      "No, I need to check my eligibility"
+    )
   end
 
   def and_i_sign_up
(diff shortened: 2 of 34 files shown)


spec/support/page_objects/teacher_interface/documents/check_uploaded_files.rb
```diff
@@ -0,0 +1,19 @@
+module PageObjects
+  module TeacherInterface
+    class CheckUploadedFilesCard < SitePrism::Section
```
gpeng, line 3: There's a pretty much identical one of these in each page object. Should we make a generic `Card` section? Or maybe just define them inline in the PageObejct if it's only going to have an `h2`?

spec/support/system_helpers.rb
```diff
@@ -141,6 +143,14 @@ def and_i_sign_up
 
     when_i_visit_the_teacher_confirmation_email
   end
+
+  def sign_in_page
```
gpeng, line 147: We can move all of the definitions out into `spec/support/page_helpers.rb` now.

spec/system/teacher_interface/authentication_spec.rb
```diff
@@ -156,11 +156,11 @@ def given_i_clear_my_session
   end
 
   def when_i_visit_the_sign_up_page
```
gpeng, line 158: There's a helper in PageHelpers now that gives you this in the format `when_i_visit_the(:sign_up_page)`

spec/system/teacher_interface/authentication_spec.rb
```diff
@@ -189,26 +189,26 @@ def when_i_click_save_and_sign_out
   end
 
   def then_i_see_the_sign_up_form
```
gpeng, line 191: There's also a helper for this although it only checks the page has loaded (`then_i_see_the(:sign_up_form)`). If you want to check the page content then this is probably still the way to go.

spec/system/teacher_interface/authentication_spec.rb
```diff
@@ -189,26 +189,26 @@ def when_i_click_save_and_sign_out
...
-    expect(page).to have_title("Your email address")
-    expect(page).to have_content("Your email address")
-    expect(page).to have_content(
+    expect(sign_up_page.heading).to have_content("Your email address")
```
gpeng, line 192: I think with these where we know the exact text and it's not that long we should use `.text` as it's more explicit:
```
expect(sign_up_page.heading.text).to eq("Your email address<etc>...")
```

spec/support/page_objects/teacher_interface/documents/check_uploaded_files.rb
David-Feetenby: It might actually not be necessary any longer. I can't fully remember what I actually put them in to achieve. I think they can just be defined in the actual pageobject itself where needed

spec/system/teacher_interface/authentication_spec.rb
David-Feetenby: I would vote to keep it with checking the page content as that would make it slightly more extensive. But, i'm not fussed either way.

spec/support/page_objects/teacher_interface/application/active_application.rb
```diff
@@ -0,0 +1,37 @@
...
+  module TeacherInterface
+    class TaskListItem < SitePrism::Section
+      element :link, "a"
+      element :status, ".govuk-tag"
+    end
+
+    class TaskListSection < SitePrism::Section
+      element :heading, "h2"
+      sections :task_list_items, TaskListItem, ".app-task-list__item"
+    end
```
thomasleese, line 11: As these aren't specific to the teacher application page (we've got a similar definition for assessor applications - https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/74595c0b43f1b0bca80e26d27ee2c6af78e2c550/spec/support/page_objects/assessor_interface/application.rb#L15-L24) I'd suggest refactoring them out in to a common location (`spec/support/page_objects`) and using them in both places.

spec/support/page_objects/teacher_interface/application/active_application.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    class ActiveApplication < SitePrism::Page
```
thomasleese, line 13: I notice this is in a folder structure `page_objects/teacher_interface/application/active_application` however the Ruby module structure is `PageObjects::TeacherInterface::ActiveApplication`.

Typically, the folder structure and module structure should match, although folders are written with underscores and models are written with uppercase letters. For the class to match the folder structure it should be `PageObjects::TeacherInterface::Application::ActiveApplication`, however in this case I'd suggest removing the `application` folder since there aren't too many different page.

spec/support/page_objects/teacher_interface/documents/document_form.rb
```diff
@@ -0,0 +1,10 @@
+module PageObjects
+  module TeacherInterface
+    class DocumentForm < SitePrism::Page
+      set_url "/teacher/"
```
thomasleese, line 4: Is this URL right for this page? I reckon it's missing references to the document form.

spec/support/page_objects/teacher_interface/new_teacher_form.rb
```diff
@@ -0,0 +1,11 @@
+module PageObjects
```
thomasleese, line 1: I notice in this PR you've also added a `SignUp` page object which I think is a duplicate of this one, I'd suggest you can remove this page object.

spec/support/page_objects/teacher_interface/application/form_page_objects/age_range_form.rb
```diff
@@ -0,0 +1,8 @@
+module PageObjects
```
thomasleese, line 1: A lot of these form objects are similar, if not identical, with a heading and some grid rows. I'd suggest the common parts could be refactored out in to a `PageObject::TeacherInterface::ApplicationForm` object and the ones which require specific elements could inherit from it.

spec/system/teacher_interface/authentication_spec.rb
```diff
@@ -245,8 +245,29 @@ def and_i_click_continue
...
+  end
+
+  def new_application_form_page
+    @new_application_form_page =
+      PageObjects::TeacherInterface::NewTeacherForm.new
+  end
+
+  def signed_out_page
+    @signed_out_page = PageObjects::TeacherInterface::SignedOut.new
+  end
```
thomasleese, line 272: These could be moved in to the `page_helpers.rb` file so they're accessible across all the tests (although I'd suggest prefixing them with `teacher_` since we will have assessor sign in pages too).

spec/system/teacher_interface/authentication_spec.rb
```diff
@@ -245,8 +245,29 @@ def and_i_click_continue
...
+
+  def check_email_page
+    @check_email_page = PageObjects::TeacherInterface::CheckYourEmail.new
+  end
+
+  def new_application_form_page
+    @new_application_form_page =
+      PageObjects::TeacherInterface::NewTeacherForm.new
+  end
```
thomasleese, line 268: The class and the method name doesn't seem to quite match here.

spec/system/teacher_interface/documents_spec.rb
```diff
@@ -57,28 +57,45 @@ def when_i_click_delete_on_the_first_document
...
+  end
+
+  def document_form_page
+    @document_form_page = PageObjects::TeacherInterface::DocumentForm.new
+  end
+
+  def check_uploaded_files_page
+    @check_uploaded_files_page =
+      PageObjects::TeacherInterface::CheckUploadedFiles.new
   end
```
thomasleese, line 100: These could also be moved in to the `page_helpers.rb` file.

spec/system/teacher_interface/application_spec.rb
```diff
@@ -685,266 +710,483 @@ def and_i_receive_an_application_email
...
+    )
+    expect(age_range_form.grid).to have_content("From")
+    expect(age_range_form.grid).to have_content("To")
   end
 
   def then_i_see_the_subjects_form
-    expect(page).to have_title("Enter the subjects you can teach")
-    expect(page).to have_content("What subjects are you qualified to teach?")
+    expect(subjects_form_page).to have_title("Enter the subjects you can teach")
+    expect(subjects_form_page.heading.text).to have_content(
```
thomasleese, line 846:
```suggestion
    expect(subjects_form_page.heading).to have_content(
```

If you're using `have_content` you don't need to do `.text` as we're checking the content of the element. If you would like to check the text directly you should use `eq`.

thomasleese (commented):
Generally looks good, I've left a few comments on my thoughts, they're probably not all strictly necessary for this PR, but take a look and let me know what you think!

spec/support/page_objects/teacher_interface/documents/document_form.rb
David-Feetenby: They aren't being used so can just be removed 👍

spec/system/teacher_interface/application_spec.rb
```diff
@@ -494,7 +494,13 @@ def when_i_click_remove
...
 
   def when_i_click_personal_information
-    click_link "Enter your personal information"
+    active_application_page
+      .task_list_sections
+      .first
+      .task_list_items
+      .first
+      .link
+      .click
```
thomasleese, line 503:
```suggestion
    active_application_page.click_item("Enter your personal information")
```

spec/system/teacher_interface/application_spec.rb
```diff
@@ -685,266 +710,482 @@ def and_i_receive_an_application_email
...
+  end
+
+  def age_range_form
+    @age_range_page = PageObjects::TeacherInterface::AgeRangeForm.new
+  end
+
+  def registration_number_form
+    @registration_number_form =
+      PageObjects::TeacherInterface::RegistrationNumberForm.new
   end
```
thomasleese, line 1190: These can go in the `page_helpers.rb` file as well, the idea is all pages objects will be defined in this one place so the individual spec files can be focussed on the tests.

spec/support/page_helpers.rb
```diff
@@ -105,4 +105,34 @@ def then_i_see_the(page, **args)
   def when_i_visit_the(page, **args)
     send(page.to_sym).load(**args)
   end
+
+  def teacher_sign_in_page
```
thomasleese, line 109: I think it would be good if the `then_i_see_the` and `when_i_visit_the` methods are defined either above or below _all_ the page object methods, so it's easier to find them in this file, rather than being in the middle of them.

spec/support/page_objects/teacher_interface/active_application.rb
```diff
@@ -0,0 +1,37 @@
...
+      element :status, ".govuk-tag"
+    end
+
+    class TaskListSection < SitePrism::Section
+      element :heading, "h2"
+      sections :task_list_items, TaskListItem, ".app-task-list__item"
+    end
+
+    class ActiveApplication < SitePrism::Page
+      set_url "/teacher/teacher/application"
```
thomasleese, line 14: Is this URL correct? It seems to have `/teacher/teacher`?

spec/support/page_objects/teacher_interface/active_application.rb
```diff
@@ -0,0 +1,37 @@
...
+      element :link, "a"
+      element :status, ".govuk-tag"
+    end
+
+    class TaskListSection < SitePrism::Section
+      element :heading, "h2"
+      sections :task_list_items, TaskListItem, ".app-task-list__item"
+    end
+
+    class ActiveApplication < SitePrism::Page
```
thomasleese, line 13: I wonder about the naming of this page. We don't really have a concept of an "active" application, since an application is active for both a teacher and an assessor. For [the assessor interface we've simply called it `Application`](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/spec/support/page_objects/assessor_interface/application.rb) and I think we can do the same here.

The pages themselves can be referenced as `teacher_application_page` and `assessor_application_page`, and I think that's actually better than `active_application_page` as just the name on its own is ambiguous as to whether it's for teachers or for assessors.

gpeng (approved):

PR #494, opened 2022-09-12
Add checks and failure reasons for work history

thomasleese:
This adds the checks and failure reasons an assessor needs to perform the assessment of the work history section.

[Trello Card](https://trello.com/c/L0YvXtqv/854-work-history-outcome)

## Screenshot

Screenshot 2022-09-12 at 13 30 52


spec/system/assessor_interface/checking_submitted_details_spec.rb
```diff
@@ -168,6 +178,34 @@ def then_i_see_the_work_history
     )
   end
 
+  def when_i_choose_check_work_history_yes
+    check_work_history_page.form.yes_radio_item.input.click
+    check_work_history_page.form.continue_button.click
+  end
+
+  def and_i_see_check_work_history_completed
+    expect(application_page.work_history_task.status).to have_content(
```
gpeng, line 187: We should use `.text` for this sort of stuff IMO and be explicit as I think this would pass if the status was 'NOT COMPLETED'. 

```suggestion
    expect(application_page.work_history_task.status.text).to eq(
```

gpeng (approved):

PR #499, opened 2022-09-13
Fix page object autoloading

gpeng:
We've been explicitly requiring page objects but it's still a bit unreliable and annoying. Move them into a sub folder and add that to the load path to get them to autoload properly.

* Move the page objects to `spec/support/autoload` and add that to the load path
* Fix typo in eligibility_interface folder name


spec/system/smoke_spec.rb
```diff
@@ -6,14 +6,14 @@
...
-require "support/page_objects/eligiblity_interface/start"
-require "support/page_objects/eligiblity_interface/teach_children"
+require "support/autoload/page_objects/eligibility_interface/country"
+require "support/autoload/page_objects/eligibility_interface/degree"
+require "support/autoload/page_objects/eligibility_interface/eligible"
+require "support/autoload/page_objects/eligibility_interface/misconduct"
+require "support/autoload/page_objects/eligibility_interface/qualification"
+require "support/autoload/page_objects/eligibility_interface/region"
+require "support/autoload/page_objects/eligibility_interface/start"
+require "support/autoload/page_objects/eligibility_interface/teach_children"
```
thomasleese, line 16: Hmm is this going to cause problems because it'll try and use the radio button page objects, but they won't have been autoloaded in the smoke test environment 🤔

thomasleese (approved):
I've approved, as it's hard to test the smoke tests locally, but I suspect we might run in to issues with them.

spec/system/smoke_spec.rb
gpeng: Yeah. I was going to remove all of those altogether but I'm not sure what Rails.env they run in. I've only autoloaded everything in test.

thomasleese: They don't run in a Rails environment at all which makes them a bit difficult to work with - because there's no database available trying to load Rails fails.

It's the reason I had to revert https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/408 and had to put the `require` in place originally: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/commit/5843662aa36645546670de54b7205461ec4fb093

gpeng: I've added the missing stuff. They're working locally for me now.

PR #506, opened 2022-09-15
Improve selected failure reasons validation

thomasleese:
This adds validation to the assessment section model to ensure that selected failure reasons are present if passed is false.

Also, if the assessor passes the assessment, having previously failed it, we should clear the selected failure reasons.

[Trello Card](https://trello.com/c/aS7cIo9S/896-assessment-section-validation) &middot; [Trello Card](https://trello.com/c/TTpNAsSU/895-clear-failure-reasons-on-yes)

### DIFF
--- a/app/controllers/assessor_interface/assessment_sections_controller.rb
+++ b/app/controllers/assessor_interface/assessment_sections_controller.rb
@@ -1,12 +1,8 @@
 module AssessorInterface
   class AssessmentSectionsController < BaseController
-    before_action :load_assessment_section
+    before_action :load_assessment_section_and_assessment
 
     def show
-      @assessment = @assessment_section.assessment
-      @application_form = @assessment.application_form
-      @qualifications = @application_form.qualifications.ordered
-      @work_histories = @application_form.work_histories.ordered
     end
 
     def update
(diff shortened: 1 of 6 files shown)


app/controllers/assessor_interface/assessment_sections_controller.rb
```diff
@@ -33,6 +29,11 @@ def load_assessment_section
             }
           )
           .find_by!(key: params[:key])
+
+      @assessment = @assessment_section.assessment
```
gpeng, line 33: I don't think we need all these for the update do we? I've actually changed this a bit in the stuff I'm doing so not a biggy if you want to merge it as is.

gpeng (approved):

app/controllers/assessor_interface/assessment_sections_controller.rb
thomasleese: We do unfortunately, to be able to render the `:show` template if the validation fails:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/b6352222fbe271f98d8b0996947d799e69e8d884/app/controllers/assessor_interface/assessment_sections_controller.rb#L19

I did wonder about refactoring this controller anyway with a view object, and it sounds like maybe your stuff will do that, so I'll merge it in for now.

PR #510, opened 2022-09-15
[890] Completing an assessment section generates a timeline event

gpeng:
When a assessor completes a section (pass or fail) we want to record it on the timeline.

* Add a new timeline event type
* Add an 'eventable' polymorphic association to the `TimelineEvent`. We could maybe refactor some of the other event types to use this but I haven't as yet. 
* Add `UpdateAssessmentSection` service 
* Add `TimelineEntry` component and refactor timeline to use that

image

### DIFF
--- a/app/components/timeline_entry/component.rb
+++ b/app/components/timeline_entry/component.rb
@@ -1,0 +1,58 @@
+module TimelineEntry
+  class Component < ViewComponent::Base
+    def initialize(timeline_event:)
+      super
+      @timeline_event = timeline_event
+    end
+
+    attr_reader :timeline_event
+
+    def description_vars
+      send("#{timeline_event.event_type}_vars")
+    end
+
+    private
+
+    def state_changed_vars
+      {
+        old_state:
+          render(
+            ApplicationFormStatusTag::Component.new(
+              key: timeline_event.id,
+              status: timeline_event.old_state,
+              class_context: "timeline-event"
+            )
+          ),
+        new_state:
+          render(
+            ApplicationFormStatusTag::Component.new(
+              key: timeline_event.id,
+              status: timeline_event.new_state,
+              class_context: "timeline-event"
+            )
+          )
+      }
+    end
+
+    def assessor_assigned_vars
+      { assignee_name: timeline_event.assignee.name }
+    end
+
+    alias_method :reviewer_assigned_vars, :assessor_assigned_vars
+
+    def assessment_section_recorded_vars
+      section = timeline_event.eventable
+      {
+        section_name: section.key.titleize,
+        section_state:
+          render(
+            ApplicationFormStatusTag::Component.new(
+              key: timeline_event.id,
+              status: section.state,
+              class_context: "timeline-event"
+            )
+          )
+      }
+    end
+  end
+end
--- a/app/controllers/assessor_interface/assessment_sections_controller.rb
+++ b/app/controllers/assessor_interface/assessment_sections_controller.rb
@@ -6,7 +6,7 @@ module AssessorInterface
     def update
       if UpdateAssessmentSection.call(
            assessment_section:,
-           user: current_user,
+           user: current_staff,
            params: assessment_section_params
          )
         redirect_to [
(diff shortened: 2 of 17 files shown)


db/migrate/20220914121554_add_eventable_fields_to_timeline_event.rb
```diff
@@ -0,0 +1,10 @@
+class AddEventableFieldsToTimelineEvent < ActiveRecord::Migration[7.0]
+  def change
+    change_table :timeline_events, bulk: true do |t|
+      t.integer :eventable_id
+      t.index :eventable_id
+      t.string :eventable_type
+      t.index :eventable_type
```
thomasleese, line 7: I think there's a way to use `t.references` in the migration to avoid needing to add the columns manually, but I don't feel too strongly on that.

app/models/timeline_event.rb
```diff
@@ -29,11 +32,13 @@
...
   belongs_to :application_form
   belongs_to :creator, polymorphic: true
+  belongs_to :eventable, polymorphic: true, optional: true
 
   enum event_type: {
          assessor_assigned: "assessor_assigned",
          reviewer_assigned: "reviewer_assigned",
-         state_changed: "state_changed"
+         state_changed: "state_changed",
+         assessment_section_completed: "assessment_section_completed"
```
thomasleese, line 41: I wonder about this name because we've got a `completed` state on assessment sections, I wonder if it will suggest this event is only if the state changed to `completed`. I wonder if something like `assessment_section_recorded` is better?

app/components/timeline_entry/component.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def description_vars
+      return state_change_vars if timeline_event.state_changed?
+      if timeline_event.reviewer_assigned? || timeline_event.assessor_assigned?
+        return assignment_vars
+      end
+      if timeline_event.assessment_section_completed?
+        assessment_section_completed_vars
+      end
+    end
```
thomasleese, line 18: I wonder if this would read nicer as a `case`/`when` or even something like `send("#{timeline_event.event_type}_vars")`. I can reasons not to do that though, so again don't feel very strongly on this.

app/controllers/assessor_interface/assessment_sections_controller.rb
```diff
@@ -29,11 +31,22 @@ def load_assessment_section_and_assessment
...
+      @assessment ||= @assessment_section.assessment
+    end
+
+    def application_form
+      @application_form ||= assessment.application_form
+    end
+
+    def qualifications
+      @qualifications ||= application_form.qualifications.ordered
+    end
```
thomasleese, line 46: Will these be accessible to the view if they're written as methods like this?

app/components/timeline_entry/component.rb
gpeng: eeuw... prettier has done that. It looked much nicer my way 😆

app/controllers/assessor_interface/assessment_sections_controller.rb
gpeng: They will not. I've missed a bit out 😮‍💨 I think your view object suggestion is probably the best way to go but I'll sort this out for now.

app/models/timeline_event.rb
gpeng: Yeah. I like that.

app/controllers/assessor_interface/assessment_sections_controller.rb
gpeng: I've added a view object.

app/controllers/assessor_interface/assessment_sections_controller.rb
```diff
@@ -18,22 +23,8 @@ def update
...
-            }
-          )
-          .find_by!(key: params[:key])
-
-      @assessment = @assessment_section.assessment
-      @application_form = @assessment.application_form
-      @qualifications = @application_form.qualifications.ordered
-      @work_histories = @application_form.work_histories.ordered
+    def assessment_section_view
+      @assessment_section_view ||= AssessmentSectionViewObject.new(params:)
```
thomasleese, line 27:
```suggestion
    def assessment_section_view_object
      @assessment_section_view_object ||= AssessmentSectionViewObject.new(params:)
```

I'd suggest this name to avoid conflicts with "views", but don't feel strongly on this.

thomasleese (approved):

PR #519, opened 2022-09-19
[913] Further information request models

gpeng:
We are building functionality to allow an assessor to request further information from an applicant. 

Some basic models to allow us to build this out.

```
Assessment
  -> has_many `FurtherInformationRequest`
      -> has_many `FurtherInformationRequestItem` (as `items`)
           -> has_many `FurtherInformationRequestDocument` (this might maybe should be has_one)
                 -> has_many `Upload`
```

The existing `Document` class is quite tightly bound to the `Application` with its types and URL helpers. I think we'll most likely end up needing these types in the FI process as the current prototype seems to render a different page depending on the document type that is being requested. It's not clear yet how we map 'failure reasons' to specific document types or whether we should have that level of complexity initially. 

As the design is still in flux with all that, I've refactored `Upload` to be associated polymorphically and added an FI specific  document class `FurtherInformationRequestDocument`. We can switch this out for the `Document` class if it makes sense as we iterate. 

The migration of `Upload` to make it polymorphic is destructive but I think this is ok as we only have test data that we can reseed in non-production environments.

### DIFF
--- a/app/models/further_information_request_item.rb
+++ b/app/models/further_information_request_item.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: further_information_request_items
+#
+#  id                             :bigint           not null, primary key
+#  assessor_notes                 :text
+#  information_type               :string
+#  response                       :text
+#  created_at                     :datetime         not null
+#  updated_at                     :datetime         not null
+#  further_information_request_id :bigint
+#
+# Indexes
+#
+#  index_fi_request_items_on_fi_request_id  (further_information_request_id)
+#
+class FurtherInformationRequestItem < ApplicationRecord
+  belongs_to :further_information_request, inverse_of: :items
+
+  enum :information_type, { text: "text", document: "document" }
+end
--- a/app/models/further_information_request_item.rb
+++ b/app/models/further_information_request_item.rb
@@ -18,6 +18,7 @@
 #
 class FurtherInformationRequestItem < ApplicationRecord
   belongs_to :further_information_request, inverse_of: :items
+  has_one :document, as: :documentable
 
   enum :information_type, { text: "text", document: "document" }
 end
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -20,6 +20,7 @@ class Assessment < ApplicationRecord
   belongs_to :application_form
 
   has_many :sections, class_name: "AssessmentSection"
+  has_many :further_information_requests
 
   enum :recommendation,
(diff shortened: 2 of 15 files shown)


thomasleese (approved):
I'm a little unsure about the document model in this PR, I certainly imagined the `Document` model as we've currently got would be generic enough to handle any kind file uploads (which is why the views/controllers were written separately to the application form).

I realise it's not quite so easy, and so we're not entirely clear on the design which makes this harder to predict, but I think personally I'd prefer us to find a way of extracting the application form specific parts out of the `Document` model so it could be used here instead of creating a new model.

However, if we do decide to go down this route, I'd suggest we rename `Document` so it's more specific to the application form and introduce a `Documentable` concern for anything common. Hopefully we can at least re-use the same views, but perhaps even those will be different for further information.

Just my thoughts, but it doesn't stop the PR going in as is.

gpeng:
I think the main thing on the existing `Document` model that is 'in the way' at the moment are the document types. For the first iteration of the FI stuff I think we need an untyped version. We could add a 'none' option or something like that to the existing model or just pick a type and refine that as we go though. The other bit is the URL stuff but we could refactor that out and it doesn't do any harm in there at the moment. I just thought it would be easier to just add another class for now but I'm happy to not do that if you prefer.

thomasleese:
I think my preference would be a `none` (or probably a `further_information`) option as we can then learn from that how we need to change the `Document` model to keep it generic, I'm worried with this approach we'll never take the chance to look back and consolidate the models if we need to.

gpeng:
Ok. I'll sort that out.

app/models/further_information_request_item.rb
```diff
@@ -0,0 +1,24 @@
...
+#  updated_at                     :datetime         not null
+#  further_information_request_id :bigint
+#
+# Indexes
+#
+#  index_fi_request_items_on_fi_request_id  (further_information_request_id)
+#
+class FurtherInformationRequestItem < ApplicationRecord
+  belongs_to :further_information_request, inverse_of: :items
+  has_many :documents, as: :documentable
```
thomasleese, line 21:
```suggestion
  has_one :document, as: :documentable
```

I suspect we'll end up only needing one document per request item (since a document can have many uploads/translations) but not a major issue - we can see how we use it.

thomasleese (approved):

app/models/further_information_request_item.rb
gpeng: Oh yeah. Meant to do that 🤦

thomasleese (approved):

PR #526, opened 2022-09-20
Error message content update

David-Feetenby:
Update content for error messages in eligiblity checker and application form 
Add two new forms to validate where none existed previously.

Screenshot 2022-09-22 at 14 34 30

Screenshot 2022-09-22 at 14 34 53

Screenshot 2022-09-22 at 14 35 15

Screenshot 2022-09-22 at 14 35 36

### DIFF
--- a/app/controllers/teacher_interface/documents_controller.rb
+++ b/app/controllers/teacher_interface/documents_controller.rb
@@ -4,6 +4,7 @@ module TeacherInterface
     before_action :load_document
 
     def edit
+      @add_another_upload_form = AddAnotherUploadForm.new
       if document.uploads.empty?
         redirect_to new_teacher_interface_application_form_document_upload_path(
                       @document,
@@ -12,14 +13,20 @@ module TeacherInterface
     end
 
     def update
-      add_another = params.dig(:document, :add_another)
-
-      if add_another && ActiveModel::Type::Boolean.new.cast(add_another)
-        redirect_to new_teacher_interface_application_form_document_upload_path(
-                      @document,
-                    )
+      add_another =
+        params.dig(:teacher_interface_add_another_upload_form, :add_another)
+      @add_another_upload_form =
+        TeacherInterface::AddAnotherUploadForm.new(add_another:)
+      if @add_another_upload_form.valid?
+        if @add_another_upload_form.add_another
+          redirect_to new_teacher_interface_application_form_document_upload_path(
+                        @document,
+                      )
+        else
+          redirect_to document.continue_url
+        end
       else
-        redirect_to document.continue_url
+        render :edit, status: :unprocessable_entity
       end
     end
   end
--- a/app/controllers/teacher_interface/uploads_controller.rb
+++ b/app/controllers/teacher_interface/uploads_controller.rb
@@ -25,15 +25,18 @@ module TeacherInterface
     end
 
     def delete
+      @delete_upload_form = DeleteUploadForm.new
     end
 
     def destroy
-      if ActiveModel::Type::Boolean.new.cast(params.dig(:upload, :confirm))
-        @upload.attachment.purge
-        @upload.destroy!
+      confirm = params.dig(:teacher_interface_delete_upload_form, :confirm)
+      @delete_upload_form =
+        TeacherInterface::DeleteUploadForm.new(confirm:, upload: @upload)
+      if @delete_upload_form.save!
+        redirect_to [:edit, :teacher_interface, :application_form, @document]
+      else
+        render :delete, status: :unprocessable_entity
       end
-
-      redirect_to [:edit, :teacher_interface, :application_form, @document]
     end
 
     private
--- a/app/forms/teacher_interface/alternative_name_form.rb
+++ b/app/forms/teacher_interface/alternative_name_form.rb
@@ -9,6 +9,9 @@ class TeacherInterface::AlternativeNameForm
   attribute :alternative_family_name, :string
 
   validates :application_form, presence: true
+  validates :alternative_given_names, presence: true, if: :has_alternative_name
+  validates :alternative_family_name, presence: true, if: :has_alternative_name
+  validates :has_alternative_name, inclusion: { in: [true, false] }
 
   def save
     return false unless valid?
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -148,11 +148,11 @@ en:
         eligibility_interface/degree_form:
           attributes:
             degree:
-              inclusion: Tell us if you have a degree
+              inclusion: Tell us whether you have a university degree
         eligibility_interface/misconduct_form:
           attributes:
             misconduct:
-              inclusion: Tell us if your employment record is free of sanctions
+              inclusion: Tell us whether your employment record has any sanctions or restrictions
         eligibility_interface/qualification_form:
           attributes:
             qualification:
@@ -164,11 +164,33 @@ en:
         eligibility_interface/teach_children_form:
           attributes:
             teach_children:
-              inclusion: Tell us if you have experience with teaching children
+              inclusion: Tell us whether you’re qualified to teach children aged between 5 and 16
         teacher_interface/name_and_date_of_birth_form:
           attributes:
             date_of_birth:
               inclusion: You must be 18 or over to use this service
+              blank: Enter your date of birth in the format 27 3 1980
+              less_than: Your date of birth must be in the past
+            given_names:
+              blank: Enter your given names
+            family_name:
+              blank: Enter your family name
+        teacher_interface/alternative_name_form:
+          attributes:
+            alternative_given_names:
+              blank: Enter your alternate given names
+            alternative_family_name:
+              blank: Enter your alternate family name
+            has_alternative_name:
+              inclusion: Tell us whether you have an alternative name
+        teacher_interface/add_another_upload_form:
+          attributes:
+            add_another:
+              inclusion: Select whether you need to upload another page
+        teacher_interface/delete_upload_form:
+          attributes:
+            confirm:
+              inclusion: Select whether you want to delete this file
         teacher_interface/new_session_form:
           attributes:
             email:
--- a/spec/forms/teacher_interface/add_another_upload_form_spec.rb
+++ b/spec/forms/teacher_interface/add_another_upload_form_spec.rb
@@ -1,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::AddAnotherUploadForm, type: :model do
+  describe "validations" do
+    subject(:form) { described_class.new(add_another:) }
+
+    context "when the fields are blank" do
+      let(:add_another) { "" }
+
+      it { is_expected.to_not be_valid }
+    end
+
+    context "when the fields are true" do
+      let(:add_another) { true }
+
+      it { is_expected.to be_valid }
+    end
+
+    context "when the fields are false" do
+      let(:add_another) { false }
+
+      it { is_expected.to be_valid }
+    end
+  end
+end
--- a/spec/forms/teacher_interface/alternative_name_form_spec.rb
+++ b/spec/forms/teacher_interface/alternative_name_form_spec.rb
@@ -1,16 +1,10 @@
 require "rails_helper"
 
 RSpec.describe TeacherInterface::AlternativeNameForm, type: :model do
-  describe "validations" do
-    it { is_expected.to validate_presence_of(:application_form) }
-  end
-
   let(:application_form) { build(:application_form) }
 
-  describe "#valid?" do
-    subject(:valid?) { form.valid? }
-
-    let(:form) do
+  describe "validations" do
+    subject(:form) do
       described_class.new(
         application_form:,
         has_alternative_name:,
@@ -18,11 +12,67 @@ RSpec.describe TeacherInterface::AlternativeNameForm, type: :model do
         alternative_family_name:,
       )
     end
-    let(:has_alternative_name) { "" }
-    let(:alternative_given_names) { "" }
-    let(:alternative_family_name) { "" }
+    let(:has_alternative_name) { nil }
+    let(:alternative_family_name) { nil }
+    let(:alternative_given_names) { nil }
 
-    it { is_expected.to be true }
+    it { is_expected.to validate_presence_of(:application_form) }
+
+    context "when the fields are all blank" do
+      let(:has_alternative_name) { "" }
+      let(:alternative_given_names) { "" }
+      let(:alternative_family_name) { "" }
+
+      it { is_expected.to_not be_valid }
+    end
+
+    context "when the fields are blank and has_alternative_name is true" do
+      let(:has_alternative_name) { true }
+      let(:alternative_given_names) { "" }
+      let(:alternative_family_name) { "" }
+
+      it { is_expected.to_not be_valid }
+    end
+
+    context "when the fields are filled out and has_alternative_name is true" do
+      let(:has_alternative_name) { true }
+      let(:alternative_given_names) { "given_name" }
+      let(:alternative_family_name) { "family_name" }
+
+      it { is_expected.to be_valid }
+    end
+
+    context "when the fields are blank and has_alternative_name is false" do
+      let(:has_alternative_name) { false }
+      let(:alternative_given_names) { "" }
+      let(:alternative_family_name) { "" }
+
+      it { is_expected.to be_valid }
+    end
+
+    context "when the fields are filled out and has_alternative_name is false" do
+      let(:has_alternative_name) { false }
+      let(:alternative_given_names) { "given_name" }
+      let(:alternative_family_name) { "family_name" }
+
+      it { is_expected.to be_valid }
+    end
+
+    context "when only one field is filled out and has_alternative_name is false" do
+      let(:has_alternative_name) { false }
+      let(:alternative_given_names) { "" }
+      let(:alternative_family_name) { "family_name" }
+
+      it { is_expected.to be_valid }
+    end
+
+    context "when only one field is filled out and has_alternative_name is true" do
+      let(:has_alternative_name) { true }
+      let(:alternative_given_names) { "given_name" }
+      let(:alternative_family_name) { "" }
+
+      it { is_expected.to_not be_valid }
+    end
   end
 
   describe "#save" do
--- a/spec/forms/teacher_interface/delete_upload_form_spec.rb
+++ b/spec/forms/teacher_interface/delete_upload_form_spec.rb
@@ -1,0 +1,53 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::DeleteUploadForm, type: :model do
+  describe "validations" do
+    subject(:form) { described_class.new(confirm:, upload:) }
+
+    context "when the fields are blank" do
+      let(:confirm) { "" }
+      let(:upload) { "" }
+
+      it { is_expected.to_not be_valid }
+    end
+
+    context "when confirm field is true but upload is blank" do
+      let(:confirm) { true }
+      let!(:upload) { "" }
+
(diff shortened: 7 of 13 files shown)


app/controllers/teacher_interface/documents_controller.rb
```diff
@@ -12,14 +13,20 @@ def edit
...
+      add_another =
+        params.dig(:teacher_interface_add_another_page_form, :add_another)
+      @add_another_page_form =
+        TeacherInterface::AddAnotherPageForm.new(add_another:)
+      if @add_another_page_form.valid?
+        if @add_another_page_form.add_another
+          redirect_to new_teacher_interface_application_form_document_upload_path(
+                        @document,
+                      )
+        elsif @add_another_page_form.add_another == false
```
thomasleese, line 25:
```suggestion
        else
```

app/controllers/teacher_interface/uploads_controller.rb
```diff
@@ -25,15 +25,23 @@ def create
...
+      confirm = params.dig(:teacher_interface_delete_file_form, :confirm)
+      @delete_file_form = TeacherInterface::DeleteFileForm.new(confirm:)
+      if @delete_file_form.valid?
+        if @delete_file_form.confirm
+          @upload.attachment.purge
+          @upload.destroy!
+          redirect_to [:edit, :teacher_interface, :application_form, @document]
+        elsif @delete_file_form.confirm == false
+          redirect_to [:edit, :teacher_interface, :application_form, @document]
+        end
```
thomasleese, line 41:
```suggestion
        end
        redirect_to [:edit, :teacher_interface, :application_form, @document]
```

Since both branches of the if statement do the same thing at the end.

app/forms/teacher_interface/add_another_page_form.rb
```diff
@@ -0,0 +1,8 @@
+class TeacherInterface::AddAnotherPageForm
```
thomasleese, line 1:
```suggestion
class TeacherInterface::AddAnotherUploadForm
```

I suggest this is a better name, since it keeps the naming consistent within the app, rather than having two words (page and upload) for the same thing, and `Upload` is the name of the model.

app/forms/teacher_interface/alternative_name_form.rb
```diff
@@ -9,7 +9,9 @@ class TeacherInterface::AlternativeNameForm
   attribute :alternative_family_name, :string
 
   validates :application_form, presence: true
-
+  validates :alternative_given_names, presence: true, if: :has_alternative_name
+  validates :alternative_family_name, presence: true, if: :has_alternative_name
+  validates :has_alternative_name, inclusion: { in: [true, false] }
   def save
```
thomasleese, line 15:
```suggestion

  def save
```

Worth leaving a gap here I think so the method is clearer.

app/forms/teacher_interface/delete_file_form.rb
```diff
@@ -0,0 +1,8 @@
+class TeacherInterface::DeleteFileForm
```
thomasleese, line 1:
```suggestion
class TeacherInterface::DeleteUploadForm
```

Similar comment with the naming of the other form.

config/locales/en.yml
```diff
@@ -164,11 +164,31 @@ en:
...
             date_of_birth:
               inclusion: You must be 18 or over to use this service
+              blank: Enter your date of birth in the format 27 3 1980
+              less_than: Your date of birth must be in the past
+            given_names:
+              blank: Enter your given names
+            family_name:
+              blank: Enter your family name
+        teacher_interface/alternative_name_form:
+          attributes:
```
thomasleese, line 179: Do we need an error message for the inclusion of `has_alternative_name`?

spec/forms/teacher_interface/add_another_page_form_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::AddAnotherPageForm, type: :model do
+  describe "#valid?" do
+    subject(:valid?) { form.valid? }
+
+    let(:form) { described_class.new(add_another:) }
```
thomasleese, line 7:
```suggestion
    subject(:form) { described_class.new(add_another:) }
```

If we change the subject of this block to the form (and remove `:valid?` as the subject), we can do:

```rb
it { is_expected.to_not be_valid }
```

rather than

```rb
it { is_expected.to be false }
```

which I think reads a little nicer.

spec/forms/teacher_interface/alternative_name_form_spec.rb
```diff
@@ -22,7 +22,7 @@
     let(:alternative_given_names) { "" }
     let(:alternative_family_name) { "" }
 
-    it { is_expected.to be true }
+    it { is_expected.to be false }
```
thomasleese, line 25: It makes sense that this would now be `false`, however we now don't have a test in this file which considers the case when the form is valid, so I think we should add one.

We should probably include a test where `has_alternative_name` is `true` and one where `has_alternative_name` is `false` too.

spec/forms/teacher_interface/delete_file_form_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::DeleteFileForm, type: :model do
+  describe "#valid?" do
+    subject(:valid?) { form.valid? }
+
+    let(:form) { described_class.new(confirm:) }
```
thomasleese, line 7:
```suggestion
    subject(:form) { described_class.new(confirm:) }
```

Similar comment to the other form.

spec/forms/teacher_interface/name_and_date_of_birth_form_spec.rb
```diff
@@ -22,7 +22,7 @@
     let(:family_name) { "" }
     let(:date_of_birth) { "" }
 
-    it { is_expected.to be true }
+    it { is_expected.to be false }
```
thomasleese, line 25: Similar comment to the other form about losing a test for when the form is valid.

app/controllers/teacher_interface/uploads_controller.rb
```diff
@@ -25,15 +25,23 @@ def create
...
     def destroy
-      if ActiveModel::Type::Boolean.new.cast(params.dig(:upload, :confirm))
-        @upload.attachment.purge
-        @upload.destroy!
+      confirm = params.dig(:teacher_interface_delete_file_form, :confirm)
+      @delete_file_form = TeacherInterface::DeleteFileForm.new(confirm:)
+      if @delete_file_form.valid?
+        if @delete_file_form.confirm
+          @upload.attachment.purge
+          @upload.destroy!
```
thomasleese, line 37: I wonder if it's worth moving this logic in to the form, so it can easily be unit tested.

Similar to the forms we've got with a `def save` method which updates a model, we could have a method which does the same:

```rb
def save
  return unless valid?

  upload.attachment.purge
  upload.destroy!
end
```

The controller becomes much simpler then, we would need to pass the `upload` in to the form in the same way that we do with `application_form` in other forms.

thomasleese (commented):
Overall looks good, I've spotted a few minor improvements we could make to this PR which I've left as comments!

It's probably worth including some screenshots in the PR description showing the new error messages. I also wonder about splitting up the commit in to various changes, for example each new form could have its own commit.

config/locales/en.yml
David-Feetenby: Actually I think we don't to validate it at all. All its used for now is the conditional validation of the individual fields

spec/forms/teacher_interface/alternative_name_form_spec.rb
David-Feetenby: Thanks to the conditional validation, i think it is now true once again. 
Ill add the two new tests however.

config/locales/en.yml
thomasleese: I think we still need to validate that the user at least choose one of the options, Yes or No.

app/controllers/teacher_interface/uploads_controller.rb
```diff
@@ -25,15 +25,20 @@ def create
...
     def destroy
-      if ActiveModel::Type::Boolean.new.cast(params.dig(:upload, :confirm))
-        @upload.attachment.purge
-        @upload.destroy!
+      confirm = params.dig(:teacher_interface_delete_upload_form, :confirm)
+      upload = load_upload
+      @delete_upload_form =
+        TeacherInterface::DeleteUploadForm.new(confirm:, upload:)
+      if @delete_upload_form.valid?
+        @delete_upload_form.save! if @delete_upload_form.confirm
```
thomasleese, line 37: I think this conditional logic should be moved in to the save method, again so it can be unit tested easily where confirm is true and false.

spec/forms/teacher_interface/add_another_upload_form_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::AddAnotherUploadForm, type: :model do
+  describe "#valid?" do
```
thomasleese, line 4:
```suggestion
  describe "validations" do
```

Since we're not strictly testing a method call directly.

spec/forms/teacher_interface/alternative_name_form_spec.rb
```diff
@@ -23,6 +23,54 @@
     let(:alternative_family_name) { "" }
 
     it { is_expected.to be true }
+
```
thomasleese, line 26: We can refactor these tests in the same way testing the form subject and using the `be_valid` assertion.

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      it { is_expected.to be_valid }
+    end
+
+    context "when the fields are false" do
+      let(:confirm) { false }
+
+      it { is_expected.to be_valid }
+    end
+
+    context "file is successfully deleted"
```
thomasleese, line 25:
```suggestion
    context "file is successfully deleted" do
```

We're missing a block here for the context.

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    let!(:upload) { create(:upload) }
+    let!(:confirm) { true }
+    it { expect { form.save! }.to change { Upload.count }.by(-1) }
```
thomasleese, line 28: I'd suggest explaining what this is doing as we're not using the RSpec shorthand (`is_expected`) so a test description can't be automatically determined.

```suggestion
    it "deletes the upload" do
      expect { form.save! }.to change { Upload.count }.by(-1) }
    end
```

spec/forms/teacher_interface/name_and_date_of_birth_form_spec.rb
thomasleese: Looks like we're still missing a test in this file for the valid case.

app/forms/teacher_interface/alternative_name_form.rb
```diff
@@ -9,6 +9,8 @@ class TeacherInterface::AlternativeNameForm
   attribute :alternative_family_name, :string
 
   validates :application_form, presence: true
+  validates :alternative_given_names, presence: true, if: :has_alternative_name
+  validates :alternative_family_name, presence: true, if: :has_alternative_name
```
thomasleese, line 14: I think we're missing a validation here on `has_alternative_name` to ensure it's either true or false.

app/controllers/teacher_interface/uploads_controller.rb
```diff
@@ -25,15 +25,19 @@ def create
...
     def delete
+      @delete_upload_form = DeleteUploadForm.new
     end
 
     def destroy
-      if ActiveModel::Type::Boolean.new.cast(params.dig(:upload, :confirm))
-        @upload.attachment.purge
-        @upload.destroy!
+      confirm = params.dig(:teacher_interface_delete_upload_form, :confirm)
+      upload = load_upload
```
thomasleese, line 33: This method is already called for this controller method, due to this line at the top of the file:

```rb
before_action :load_upload, only: %i[delete destroy]
```

so it's already accessible as `@upload` which you can then pass in to the form.

spec/forms/teacher_interface/alternative_name_form_spec.rb
```diff
@@ -8,9 +8,7 @@
   let(:application_form) { build(:application_form) }
 
   describe "#valid?" do
```
thomasleese, line 10:
```suggestion
  describe "validations" do
```

Since we're not strictly testing a method call directly.

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::DeleteUploadForm, type: :model do
+  describe "#validations" do
```
thomasleese, line 4:
```suggestion
  describe "validations" do
```

The `#` at the beginning signifies a method name.

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    subject(:form) { described_class.new(confirm:, upload:) }
+
+    context "when the fields are blank" do
+      let!(:confirm) { "" }
```
thomasleese, line 8:
```suggestion
      let(:confirm) { "" }
```

I don't think we need any of the `confirm` definitions to be non-lazy as they don't have any side-effects (creating objects, etc).

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      it { is_expected.to_not be_valid }
+    end
+
+    context "when the fields are true" do
+      let!(:confirm) { true }
+      let!(:upload) { "" }
+
+      it { is_expected.to be_valid }
+    end
```
thomasleese, line 19: I don't think this should be valid, as it means the form is valid if there's no upload, but then the `save!` method wouldn't work if there's no upload and confirm is `true`.

I think you need to add some validation that `upload` is present.

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      it "deletes the upload" do
+        expect { form.save! }.to change { Upload.count }.by(-1)
+      end
+    end
+
+    context "file is not successfully deleted" do
+      let!(:upload) { create(:upload) }
+      let!(:confirm) { false }
+
+      it " checks the upload" do
```
thomasleese, line 41:
```suggestion
      it "doesn't delete the upload" do
```

spec/forms/teacher_interface/name_and_date_of_birth_form_spec.rb
```diff
@@ -22,16 +22,26 @@
...
```
thomasleese, line 25: I think we should group this assertion with the `let`s above it in to a `context` for when the fields are blank so it's easier to read.

I also think this test could do with refactoring to use `be_valid` as with the other form tests.

spec/forms/teacher_interface/add_another_upload_form_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::AddAnotherUploadForm, type: :model do
+  describe "validation?" do
```
thomasleese, line 4:
```suggestion
  describe "validations" do
```

spec/forms/teacher_interface/alternative_name_form_spec.rb
```diff
@@ -7,10 +7,8 @@
 
   let(:application_form) { build(:application_form) }
 
-  describe "#valid?" do
-    subject(:valid?) { form.valid? }
-
-    let(:form) do
+  describe "#validation?" do
```
thomasleese, line 10:
```suggestion
  describe "validations" do
```

thomasleese: You can also merge this section with the describe block at the top of this file.

spec/forms/teacher_interface/add_another_upload_form_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::AddAnotherUploadForm, type: :model do
+  describe "validation" do
```
thomasleese, line 4:
```suggestion
  describe "validations" do
```

spec/forms/teacher_interface/alternative_name_form_spec.rb
```diff
@@ -1,46 +1,92 @@
...
-    let(:form) do
+  describe "validations" do
+    subject(:form) do
       described_class.new(
         application_form:,
         has_alternative_name:,
         alternative_given_names:,
         alternative_family_name:,
       )
-    end
```
thomasleese, line 20: I think we need to keep this `end` as otherwise the `subject` block doesn't end correctly, it should only contain the `described_class.new`.

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      it "deletes the upload" do
+        expect { form.save! }.to change { Upload.count }.by(-1)
+      end
+    end
+
+    context "file is not successfully deleted" do
+      let!(:upload) { create(:upload) }
+      let(:confirm) { false }
+
+      it " doesn't delete the upload" do
```
thomasleese, line 41:
```suggestion
      it "doesn't delete the upload" do
```

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+RSpec.describe TeacherInterface::DeleteUploadForm, type: :model do
+  describe "validations" do
+    subject(:form) { described_class.new(confirm:, upload:) }
+
+    context "when the fields are blank" do
+      let(:confirm) { "" }
+      let!(:upload) { "" }
```
thomasleese, line 9:
```suggestion
      let(:upload) { "" }
```

These can be lazy as they don't create any objects.

spec/forms/teacher_interface/delete_upload_form_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      it { is_expected.to_not be_valid }
+    end
+
+    context "when the fields are true" do
+      let(:confirm) { true }
+      let!(:upload) { "" }
+
+      it { is_expected.to_not be_valid }
+    end
```
thomasleese, line 19: This looks good, I think we need one more validation test where `confirm` is `true` and `upload` is a valid upload.

spec/forms/teacher_interface/alternative_name_form_spec.rb
```diff
@@ -1,28 +1,78 @@
...
-
   let(:application_form) { build(:application_form) }
-
-  describe "#valid?" do
-    subject(:valid?) { form.valid? }
-
-    let(:form) do
+  let(:has_alternative_name){}
+  let(:alternative_family_name){}
+  let(:alternative_given_names){}
```
thomasleese, line 7:
```suggestion
  let(:has_alternative_name) { nil }
  let(:alternative_family_name) { nil }
  let(:alternative_given_names) { nil }
```

Better to be explicit here I think so it's clear, also I reckon it makes more sense to define these within the `describe "validations"` block as it's closer to where they're used (they're not used in the `describe "#save"` block so they don't need to be accessible there for example).

thomasleese (approved):

PR #527, opened 2022-09-21
Add notes to application forms

thomasleese:
This adds the ability for staff users to add free-text notes to application forms which are then visible in the timeline.

[Trello Card](https://trello.com/c/9vTHYLqF/917-add-a-note-to-a-case)

## Screenshots

Screenshot 2022-09-21 at 10 42 54
Screenshot 2022-09-21 at 10 43 00
Screenshot 2022-09-21 at 10 44 36

### DIFF
--- a/app/forms/assessor_interface/create_note_form.rb
+++ b/app/forms/assessor_interface/create_note_form.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class AssessorInterface::CreateNoteForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveRecord::AttributeAssignment
+
+  attr_accessor :application_form, :author
+  attribute :text, :string
+
+  validates :application_form, :author, :text, presence: true
+
+  def save!
+    return false unless valid?
+
+    ActiveRecord::Base.transaction do
+      note = Note.create!(application_form:, author:, text:)
+
+      TimelineEvent.create!(
+        application_form:,
+        event_type: "note_created",
+        creator: author,
+        note:,
+      )
+    end
+  end
+end
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -53,6 +53,7 @@ class ApplicationForm < ApplicationRecord
   has_many :documents, as: :documentable
   has_many :timeline_events
   has_one :assessment
+  has_many :notes
 
   before_create :build_documents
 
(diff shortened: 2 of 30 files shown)


app/models/application_form.rb
```diff
@@ -53,6 +53,7 @@ class ApplicationForm < ApplicationRecord
   has_many :documents, as: :documentable
   has_many :timeline_events
   has_one :assessment
+  has_many :notes
```
gpeng, line 56: Should the notes be associated with the assessment rather than the application_form? I'm not sure it matters really at the moment. Just wondering.

app/forms/assessor_interface/create_note_form.rb
```diff
@@ -0,0 +1,27 @@
...
+    return false unless valid?
+
+    ActiveRecord::Base.transaction do
+      note = Note.create!(application_form:, author:, text:)
+
+      TimelineEvent.create!(
+        application_form:,
+        event_type: "note_created",
+        creator: author,
+        note:,
```
gpeng, line 23: We could use the eventable association here. If we're going to add note explicitly, then maybe we should change the other eventable things (or thing at the moment I think) to be explicit too and get rid of the polymorphic association. This does seem like a good place to use one to me though but happy to discuss.

gpeng (commented):
Just a couple of things but looks good. I don't think the button is in the right place as per the prototype. Isn't it supposed to be on the right?

Screenshot 2022-09-21 at 13 21 45

app/models/application_form.rb
thomasleese: Yes, initially I was going to put it on the assessment, but then I started to think about how it was going to be used, and I realised notes could be added at any point in the process, since they're generic, and not necessarily tied to a particular assessment (especially if we start having multiple assessments for reviewers, etc). I figured since notes could be used for anything, I didn't want to tie them to an assessment.

thomasleese:
> Just a couple of things but looks good. I don't think the button is in the right place as per the prototype. Isn't it supposed to be on the right?

It renders on the right for me in Safari, are you using a different browser where it's not working correctly?

Screenshot 2022-09-21 at 13 31 42

gpeng:
> > Just a couple of things but looks good. I don't think the button is in the right place as per the prototype. Isn't it supposed to be on the right?
> 
> It renders on the right for me in Safari, are you using a different browser where it's not working correctly?
> 
> Screenshot 2022-09-21 at 13 31 42

It's rendering like that in safari and chrome for me.

app/forms/assessor_interface/create_note_form.rb
gpeng: Fair enough. Let's do that 👍

gpeng (approved):

PR #530, opened 2022-09-22
Restrict access to non-draft application forms

thomasleese:
Once an application form has been submitted, we don't want teachers to be able to go in and make changes to the application form. We will need to open changes up again when we implement further information, but until that's built we can lock the application form (and all associated models) as soon as it's submitted.

[Trello Card](https://trello.com/c/PyM3bWiq/927-redirect-teachers-following-login-and-restrict-access)

### DIFF
--- a/app/controllers/teacher_interface/age_range_controller.rb
+++ b/app/controllers/teacher_interface/age_range_controller.rb
@@ -1,5 +1,6 @@
 module TeacherInterface
   class AgeRangeController < BaseController
+    before_action :redirect_unless_application_form_is_draft
     before_action :load_application_form
 
     def edit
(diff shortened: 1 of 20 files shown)


gpeng (approved):
I wonder if implementing this a concern would be better given we're going to need to add further logic into the redirect that won't affect all pages. Fine for now though, we can do that when it gets more complicated.

thomasleese:
> I wonder if implementing this a concern would be better given we're going to need to add further logic into the redirect that won't affect all pages. Fine for now though, we can do that when it gets more complicated.

I had actually written it as a concern initially, but then I realised the `before_action` couldn't go in the concern as some controllers needed a `only:` parameter, and then the concern was just containing a single method and nothing else. Happy to see how this code grows for further information though and I wouldn't be surprised if refactoring it in to a concern makes sense.

PR #540, opened 2022-09-27
Further information request text response

thomasleese:
This adds the ability for a teacher to respond to a text further information request.

[Trello Card](https://trello.com/c/e4ZUBUwF/929-fi-request-text-submission)

## Screenshot

Screenshot 2022-09-27 at 08 21 45

### DIFF
--- a/app/controllers/teacher_interface/further_information_request_items_controller.rb
+++ b/app/controllers/teacher_interface/further_information_request_items_controller.rb
@@ -1,30 +1,60 @@
 module TeacherInterface
   class FurtherInformationRequestItemsController < BaseController
-    before_action :load_application_form
+    before_action :load_application_form,
+                  :load_further_information_request_and_item
 
     def edit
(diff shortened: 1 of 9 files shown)


app/controllers/teacher_interface/further_information_request_items_controller.rb
```diff
@@ -4,8 +4,44 @@ class FurtherInformationRequestItemsController < BaseController
...
+        @further_information_request_item_form =
+          FurtherInformationRequestItemTextForm.new(
+            further_information_request_form_params.merge(
+              further_information_request_item:,
+            ),
+          )
+      end
+
+      if params[:next] == "save_and_continue"
+        if @further_information_request_item_form.save
```
gpeng, line 28: Will this be `nil` if further_information_request_item isn't text? I think this action is getting a bit out of hand in any case. Could we move some of this logic into something that handles any further_information_request_item type and move all the branching out into that?

PR #548, opened 2022-09-28
Record notes with assessment section failure reasons

thomasleese:
This changes how we record failure reasons to associate a note with them, which we can then use to generate further information request items in a later PR.

## Screenshots

Screenshot 2022-09-28 at 13 44 04

### DIFF
--- a/app/controllers/assessor_interface/assessment_sections_controller.rb
+++ b/app/controllers/assessor_interface/assessment_sections_controller.rb
@@ -3,16 +3,25 @@
 module AssessorInterface
   class AssessmentSectionsController < BaseController
     def show
-      assessment_section_view_object
+      @assessment_section_form =
+        assessment_section_form.new(
+          assessment_section:,
+          user: current_staff,
+          passed: assessment_section.passed,
+          selected_failure_reasons: assessment_section.selected_failure_reasons,
+        )
     end
 
     def update
-      if UpdateAssessmentSection.call(
-           assessment_section:
-             assessment_section_view_object.assessment_section,
-           user: current_staff,
-           params: assessment_section_params,
-         )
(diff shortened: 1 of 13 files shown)


app/controllers/assessor_interface/assessment_sections_controller.rb
```diff
@@ -3,16 +3,27 @@
 module AssessorInterface
   class AssessmentSectionsController < BaseController
     def show
-      assessment_section_view_object
+      @assessment_section_form =
+        assessment_section_view_object.assessment_section_form.new(
```
gpeng, line 7: Could this be refactored into a method on the view object? Seems a bit convoluted to set a thing on the view object by passing lots of things that the view object already has.

thomasleese: Yeah, this is actually how I had it set up already, but decided against it as it felt like I was passing a lot of information about the request in to the view object, but I've since refactored it anyway so let me have another go!

thomasleese: I've gone a slightly different way with it, by pulling everything related to the form _out_ of the view object. To me it feels right that the controller is responsible for the logic of submitting the form in its entirety, since the view object seems like the presentation layer.

Putting the form logic in the view object has some problems, specifically that it needs to know about whether it's generating a form for the edit or update method (which feels like we're leaking controllers concerns in to the view object) and requires access to the current staff user.

I'm happy to go with moving it all in the view object if you feel strongly, but I think I prefer this approach.

gpeng: That looks better to me. 👍

gpeng (approved):

PR #549, opened 2022-09-29
Add titles to further information request items

thomasleese:
This determines the title from either the failure reason (if it's a text item) or the document type (if it's a document item).

## Screenshot

Screenshot 2022-09-29 at 09 08 09


gpeng (approved):
This looks good. There was talk of not necessarily displaying the actual 'failure reason' to the user but we can easily add another section to the locale file if we need to display different stuff.

PR #550, opened 2022-09-29
Disable further information requests for decline reasons

thomasleese:
There are certain failure reasons which will prevent the assessor from being able to request further information, and instead only allow an application form to be declined.

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -50,10 +50,12 @@ class Assessment < ApplicationRecord
   end
 
   def can_decline?
-    sections.any? { |section| section.state == :action_required }
+    action_required?
   end
 
-  alias_method :can_request_further_information?, :can_decline?
+  def can_request_further_information?
+    action_required? && !must_decline?
+  end
 
   def available_recommendations
     [].tap do |recommendations|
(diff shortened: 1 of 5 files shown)


app/models/assessment.rb
```diff
@@ -53,7 +53,9 @@ def can_decline?
     sections.any? { |section| section.state == :action_required }
   end
 
-  alias_method :can_request_further_information?, :can_decline?
+  def can_request_further_information?
+    can_decline? && sections.none?(&:decline?)
```
gpeng, line 57: This seems a bit confusing with all the `decline` mentions.

gpeng (approved):

app/models/assessment.rb
thomasleese: Good point, I've separated it in to two lines, does that help?

gpeng: Yeah. I think that's better. What about splitting a couple of private methods out? Something like:

```
  def can_decline?
    action_required?
  end

  def can_request_further_information?
    action_required? && !must_decline?
  end

  def must_decline?
    sections.any?(&:declines_assessment)
  end

  def action_required?
    sections.any? { |section| section.state == :action_required }
  end

```

gpeng (approved):

PR #555, opened 2022-09-30
Move 'en.yml' content to its own appropriate locale file

David-Feetenby:
This PR moves the validation error messages for teacher interface and assessor interface out of 'en.yml' and in to their own .yml files

### DIFF
--- a/config/locales/assessor_interface.en.yml
+++ b/config/locales/assessor_interface.en.yml
@@ -92,3 +92,7 @@ en:
           attributes:
             text:
               blank: Enter the note
+        assessor_interface/assessor_assignment_form:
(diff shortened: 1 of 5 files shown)


thomasleese (approved):

thomasleese (commented):
Looks good, I think there's a few more things we could move around, just from a brief look:

- Anything under `application_form` could go in to the teacher interface
- Anything under `eligibility_checker` could go in the eligibility interface
- Would be tempted to move `mailer` to its own `mailer.en.yml` file

Happy to approve if this is going to be in a separate PR though.

thomasleese (approved):

PR #561, opened 2022-10-03
Calculate working days since submission

thomasleese:
This adds a new field on application forms which we can use to calculate the number of working days that have passed since the form was submitted. We'll use this in the future to calculate an SLA, and it avoids the need to perform this calculation in the analytics as we can send it directly. The job runs every day at 03:00.

[Trello Card](https://trello.com/c/GETqyWhB/948-calculate-working-days-since-application-submission)


lib/calendars/england_and_wales.yml
```diff
@@ -0,0 +1,16 @@
...
+
+  # 2023
+  - 2023-01-02
+  - 2023-04-07
+  - 2023-04-10
+  - 2023-05-01
+  - 2023-05-29
+  - 2023-08-28
+  - 2023-12-25
+  - 2023-12-26
```
gpeng, line 16: We should add something about this to the docs. We need to think about how we'll remember to add 2024 🤔

spec/jobs/update_working_days_since_submission_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    let(:tuesday_today) { Date.new(2022, 10, 4) }
+    let!(:draft_application_form) { create(:application_form) }
+    let!(:monday_application_form) do
+      create(:application_form, :submitted, submitted_at: Date.new(2022, 10, 3))
+    end
+    let!(:friday_application_form) do
+      create(:application_form, :submitted, submitted_at: Date.new(2022, 9, 30))
+    end
+
+    before { travel_to(tuesday_today) { perform } }
```
gpeng, line 18: Would `around` be clearer here than `before`? (I didn't know you could even do this so I might not be 'down with the kids' here)

gpeng (approved):

lib/calendars/england_and_wales.yml
thomasleese: Yes, agreed, I wasn't really sure where to capture updating this... a card on the backlog?

spec/jobs/update_working_days_since_submission_job_spec.rb
thomasleese: With around it would be something like:

```rb
around do |example|
  travel_to(tuesday_today) { example.run }
end

before { perform }
```

Personally I think this actually be a little more unclear to someone who isn't familiar with Rspec, but I'm happy to change it, what do you think?

gpeng: I don't mind that much if you disagree. 👍

PR #564, opened 2022-10-04
Improve application form validation

thomasleese:


David-Feetenby (approved):

app/forms/teacher_interface/base_form.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class TeacherInterface::BaseForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  def save(validate:)
+    return false if validate && !valid?
+
+    update_model
```
gpeng, line 10: If some of the sub classes of this don't need to do any update to the model should we do a `respond_to?` guard here instead of them having to implement an empty `update_model` method? I suppose the downside of that is it would fail silently if we forgot.

thomasleese: Initially I was going to leave an implementation in this class which raised a `NotImplementedError` until I learnt it's [not technically designed for this use](https://stackoverflow.com/questions/13668068/how-to-signal-not-implemented-yet#comment63130968_37293953).

I prefer this implementation where we have to explicitly implement the method in each subclass, even if the form doesn't need one, so it doesn't fail silently as you mentioned.

PR #581, opened 2022-10-10
Submit further information

thomasleese:
This adds the functionality that allows a teacher to submit a further information request.

Depends on #580

## Screenshots

Screenshot 2022-10-10 at 11 20 39
Screenshot 2022-10-10 at 11 27 34

### DIFF
--- a/app/controllers/teacher_interface/further_information_requests_controller.rb
+++ b/app/controllers/teacher_interface/further_information_requests_controller.rb
@@ -8,6 +8,15 @@ module TeacherInterface
     def edit
     end
 
+    def update
+      SubmitFurtherInformationRequest.call(
+        further_information_request: @view_object.further_information_request,
+        user: current_teacher,
+      )
+
+      redirect_to %i[teacher_interface application_form]
+    end
+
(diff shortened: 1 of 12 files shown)


app/controllers/teacher_interface/further_information_requests_controller.rb
```diff
@@ -8,6 +8,17 @@ def show
     def edit
     end
 
+    def update
+      if @view_object.can_submit?
```
gpeng, line 12: I was wondering about this in the submit application flow. If we're going to check here then we should do something different if it returns false. Given we're only showing the button if this is true then this is probably redundant. If we wanted an extra layer of protection maybe we should raise in the service if this is false.

thomasleese: That makes sense, happy to make the service raise an exception.

thomasleese: I've amended the commit to implement it with an exception.

gpeng (approved):

PR #586, opened 2022-10-10
Add cancel button

David-Feetenby:
Add a cancel button to 'QTS Review' page which redirects back to the case.

Screenshot 2022-10-11 at 09 36 20


thomasleese (approved):
It's worth putting a screenshot in the PR description I think.

PR #587, opened 2022-10-10
[920] Require confirmation that the applicant has no sanctions

gpeng:
Prior to submission of a completed application form we want the applicant to confirm that they have no sanctions.

* Add `ApplicationForm#confirmed_no_sanctions`
* Add `SanctionConfirmationForm` form object to validate when the application is submitted
* Add sections to the confirmation page and use them in the page object to avoid ambiguous heading references

Screenshot 2022-10-10 at 17 40 06

### Review

* should we store the date rather than a boolean field?

### DIFF
--- a/app/forms/teacher_interface/sanction_confirmation_form.rb
+++ b/app/forms/teacher_interface/sanction_confirmation_form.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module TeacherInterface
+  class SanctionConfirmationForm < BaseForm
+    attribute :application_form
+    attribute :confirmed_no_sanctions
+
+    validates :confirmed_no_sanctions, presence: true
+
+    private
+
+    def update_model
+      application_form.update!(confirmed_no_sanctions:)
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


app/forms/teacher_interface/sanction_confirmation_form.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module TeacherInterface
+  class SanctionConfirmationForm < BaseForm
+    attribute :application_form
+    attribute :confirmed_no_sanctions
+
+    validate :no_sanctions_confirmed
```
thomasleese, line 8:
```suggestion
    validate :confirmed_no_sanctions, presence: true
```

You could have this and then the locale key would be `blank` and avoids needing to have its own method.

app/forms/teacher_interface/sanction_confirmation_form.rb
```diff
@@ -0,0 +1,22 @@
...
+    private
+
+    def no_sanctions_confirmed
+      unless confirmed_no_sanctions
+        errors.add(:confirmed_no_sanctions, :not_confirmed)
+      end
+    end
+
+    def update_model
+      application_form.update(confirmed_no_sanctions:)
```
thomasleese, line 19:
```suggestion
      application_form.update!(confirmed_no_sanctions:)
```

I've been using `update!` since it should always pass if the form is valid.

thomasleese (approved):

PR #593, opened 2022-10-12
Add work history contact name

thomasleese:
This adds a new field for work history which allows teachers to enter a contact name associated with their work history.

[Trello Card](https://trello.com/c/ps6GWVMI/1010-work-history-request-contact-name)

## Screenshots

Screenshot 2022-10-12 at 15 41 51
Screenshot 2022-10-12 at 15 45 24


app/forms/teacher_interface/work_history_form.rb
```diff
@@ -18,7 +19,8 @@ class WorkHistoryForm < BaseForm
     validates :school_name, presence: true
     validates :city, presence: true
     validates :country_code, presence: true
-    validates :email, presence: true, valid_for_notify: true
+    validates :contact_name, presence: true
```
gpeng, line 22: We'll need to add messages for these in the teacher_interface locale file.

gpeng (changes requested):

app/forms/teacher_interface/work_history_form.rb
thomasleese: Good point, I've added some for these fields are the remaining fields in this form.

gpeng (approved):
We can have a race and see who gets the validation messages in first 😆

PR #614, opened 2022-10-21
DQT TRN Request

thomasleese:


app/services/create_dqt_trn_request.rb
```diff
@@ -0,0 +1,32 @@
...
+    @application_form = application_form
+  end
+
+  def call
+    request_id = SecureRandom.uuid
+
+    dqt_trn_request = DQTTRNRequest.create!(application_form:, request_id:)
+
+    suppress(Faraday::Error) do
+      response = DQT::Client::TRNRequest.call(application_form:, request_id:)
```
gpeng, line 16: I wonder if we need to split this up. I'm just thinking when we start to enqueue this or need to retry it. We'll probably want a service that takes a `DQTTRNRequest`, makes the call and updates the state. If we retry this service it'll get a new request_id and make a new request. We could do that that in the next bit when we think about the jobs but I wonder if it would be easier to test manually if we split it now?

thomasleese: I agree, we will need to split this up. I’d figured we’d implement this when we build the jobs as you mentioned, but I’m happy to implement in this PR if you think it’s worth doing?

gpeng: Might be easier to do it now I think as it'll help with testing it from the console.

thomasleese: I've added an `UpdateDQTTRNRequest` service which can do this now, make a request to the DQT API to retrieve an existing TRN Request and update the model.

gpeng (approved):

PR #621, opened 2022-10-24
Update DQT TRN Requests in the background

thomasleese:
This adds a job which updates DQT TRN requests in the background if they're still pending. For now it will run every 12 hours for up to 14 times, before giving up, and sending the `StillPending` error to Sentry.

Depends on #614 

[Trello Card](https://trello.com/c/2jWRWJQt/1047-poll-for-dqt-response)


gpeng (approved):
This looks good. I'm not sure about the wait on the retries though. I expect (totally anecdotally obviously) that the assessor would award, get told there's a duplicate and then possibly go and resolve it. Maybe we should reduce the time?

PR #624, opened 2022-10-25
Store application form status in the database

thomasleese:
This changes how we calculate and display the status of individual application form "spokes" to store them as fields in the database. The main benefit of this is that it allows the values to be stored in BigQuery and analysed.

This helps us towards removing view logic in the application form model, but there are still a few extra bits which could be moved over in to a view object in the future.

[Trello Card](https://trello.com/c/270iOoey/1045-store-and-send-spoke-status-to-analytics)


thomasleese:
> I'm a bit uneasy about triggering this using callbacks but it's a pragmatic approach for now.

Yeah, I figured we could refactor it if it turned out to be problematic.

PR #626, opened 2022-10-25
Create DQT TRN requests after awarding

thomasleese:
This creates a new job which creates DQT TRN requests after awarding a teacher with QTS.

[Trello Card](https://trello.com/c/B7SrqO78/1048-call-the-dqt-api-on-award)


app/services/create_dqt_trn_request.rb
```diff
@@ -11,12 +11,25 @@ def call
...
     dqt_trn_request = DQTTRNRequest.create!(request_id:, application_form:)
 
-    dqt_trn_request.update_from_dqt_response(
-      DQT::Client::CreateTRNRequest.call(request_id:, application_form:),
-    )
+    begin
+      response =
+        DQT::Client::CreateTRNRequest.call(request_id:, application_form:)
+    rescue StandardError
+      dqt_trn_request.destroy!
```
gpeng, line 18: I don't think we can do this as we'll risk creating duplicates on DQT. If the request goes through but appears to fail to us, a timeout or network issue on the way back, then we need to retry with the same request_id as that's what the API uses to detect duplicate requests. I think we should split the API request and the creating the DQTTRequest into separate operations. Create the record with the ID and then pass it into the job and on into here. We wouldn't need to worry about the retry then. We might need to think about another state on it though I suppose, like new, pending, complete or something with new indicated we've never sent it yet. Maybe I'm getting carried away there and pending is fine.

gpeng (changes requested):

app/services/create_dqt_trn_request.rb
thomasleese: Yeah, that’s true, the is a risk there. I figure though in that case the retry would still fail as we’d be attempted to create a TRN request with the same ID, but we can handle that and instead perform the read operation.

I had thought about doing it as you suggest but liked the idea of there only being a record in our database if the request has succeeded, but you’re right there is a case where that doesn’t work. I will add a third “new” state to represent this, I think it’ll be required to properly understand where a request got to if we need to retry.

gpeng: Could this service create the DQTTrnRequest record and enqueue a job passing the record as an argument?

thomasleese: I was thinking we would pass the request ID in and the service would find or create, just to keep it all within the service but that would work too.

app/models/dqt_trn_request.rb
```diff
@@ -20,14 +20,17 @@
...
 
-  enum :state, { pending: "pending", complete: "complete" }, default: :pending
-
+  enum :state,
+       { initial: "initial", pending: "pending", complete: "complete" },
+       default: :initial
   validates :state, presence: true, inclusion: { in: states.values }
 
   def update_from_dqt_response(response)
     if (trn = response[:trn]).present?
```
gpeng, line 29: Should this just passing in a TRN? I don't think we should be parsing the response in the model. I don't like the idea of passing nil in here either though, or moving the state logic into the worker.

spec/jobs/update_dqt_trn_request_job_spec.rb
```diff
@@ -6,24 +6,130 @@
   describe "#perform" do
     subject(:perform) { described_class.new.perform(dqt_trn_request) }
 
-    let(:dqt_trn_request) { create(:dqt_trn_request) }
+    let(:perform_rescue_exception) do
+      perform
+    rescue StandardError # rubocop:disable Lint/SuppressedException
```
gpeng, line 11: Shall we keep all of these in the `.rubocop.yml` so they're not littered about the place?

app/models/dqt_trn_request.rb
thomasleese: It could be moved into a service class, but that feels like overkill to me. I think I don't mind passing `nil` in if we name the method correctly.

app/jobs/update_dqt_trn_request_job.rb
```diff
@@ -4,12 +4,30 @@ class UpdateDQTTRNRequestJob < ApplicationJob
...
+    dqt_trn_request.pending! if dqt_trn_request.initial?
+
+    if (trn = response[:trn]).present?
+      ActiveRecord::Base.transaction do
+        dqt_trn_request.application_form.teacher.update!(trn:)
+        dqt_trn_request.complete!
+      end
+    end
 
     raise StillPending if dqt_trn_request.pending?
```
gpeng, line 32: Will these get reported to Sentry? We probably don't want that.

gpeng (approved):

app/jobs/update_dqt_trn_request_job.rb
thomasleese: My understanding is that only the last one will end up in Sentry, i.e. once it's exhausted all the retries.

PR #636, opened 2022-10-27
[1063] Check validation on file upload pages

gpeng:
* Fix validation to check for a translation file if the appropriate radio button is selected. This is working now but there's now another issue that if the original file is present but the translation isn't the original file is lost from the upload element when the page reloads. This is an improvement but needs some more work. We're going to review the file upload flow in any case.
* Update the titles that reference the awarding institution
* Add proper error messages to the locale file.

### DIFF
--- a/app/forms/teacher_interface/upload_form.rb
+++ b/app/forms/teacher_interface/upload_form.rb
@@ -5,6 +5,7 @@ module TeacherInterface
     attr_accessor :document
     attribute :original_attachment
     attribute :translated_attachment
+    attribute :written_in_english, type: :boolean
 
     validates :document, presence: true
     validates :original_attachment, file_upload: true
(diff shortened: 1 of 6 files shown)


app/forms/teacher_interface/upload_form.rb
```diff
@@ -28,8 +29,9 @@ def update_model
     end
 
     def attachment_present
-      if original_attachment.blank? && translated_attachment.blank?
-        errors.add(:original_attachment, :blank)
+      errors.add(:original_attachment, :blank) if original_attachment.blank?
+      if written_in_english == "false" && translated_attachment.blank?
```
thomasleese, line 33:
```suggestion
      if written_in_english == false && translated_attachment.blank?
```

I’d expect the value to be converted to a bool, but if this works then no big deal!

thomasleese (approved):

PR #639, opened 2022-10-27
Change set-output to GITHUB_OUTPUT

merrilmathew:
### Context

Github step output has been changed from set-output to GITHUB_OUTPUT.


thomasleese (approved):
Code change looks good, although would be good to include a reasoning behind the change.

PR #643, opened 2022-10-28
Add assessment confirmation page

thomasleese:
This adds a new confirmation page for awarding or decline an application. We expect this will be temporary as the assessors are getting used to the new system.

[Trello Card](https://trello.com/c/Ub9tSDYs/1052-check-and-send-decline-email)

## Screenshots

Screenshot 2022-10-28 at 12 27 32

Screenshot 2022-10-28 at 12 44 31

### DIFF
--- a/config/locales/assessor_interface.en.yml
+++ b/config/locales/assessor_interface.en.yml
@@ -49,6 +49,16 @@ en:
           age_range_subjects: Age range and subjects
           work_history: Work history
           professional_standing: Professional standing
+      confirm:
+        legend:
+          award: Award this QTS application?
+          decline: Decline this QTS application?
+        hint:
+          award: Select ‘Yes’ to confirm you want to award this application. If you’re not sure, or you need to check something, select ‘No’.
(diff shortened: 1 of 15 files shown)


David-Feetenby (approved):

config/locales/assessor_interface.en.yml
```diff
@@ -145,19 +152,21 @@ en:
...
               blank: Select an assessor to assign
+        assessor_interface/assessment_recommendation_form:
+          attributes:
+            recommendation:
+              blank: Select your recommendation
+              inclusion: Select your recommendation
+            declaration:
+              blank: You must confirm that you have reviewed this QTS application
+            confirmation:
+              inclusion: Select whether you want to confirm this application
```
gpeng, line 163: This doesn't make sense when it's a decline that we're confirming.

thomasleese: Good point, after discussing with Wes I've changed this so the message is dynamic depending on whether the user is awarding or declining.

PR #660, opened 2022-11-01
Build award flow

thomasleese:
This builds the award flow to allow assessors to award QTS to an application and then have the teacher automatically receive an email once their details have been recorded in DQT. It required added a new state to the application form, "awarded pending checks" which we can use in the future for any other checks that need to happen before an application is awarded.

Depends on #662 

[Trello Card](https://trello.com/c/lkvfY3JA/1091-awarded-application-form-view)

### DIFF
--- a/app/jobs/update_dqt_trn_request_job.rb
+++ b/app/jobs/update_dqt_trn_request_job.rb
@@ -17,6 +17,7 @@ class UpdateDQTTRNRequestJob < ApplicationJob
     if (trn = response[:trn]).present?
       update_teacher_trn(dqt_trn_request, trn)
       send_award_email(dqt_trn_request)
+      update_application_form(dqt_trn_request)
       dqt_trn_request.complete!
     end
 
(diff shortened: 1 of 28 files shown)


app/jobs/update_dqt_trn_request_job.rb
```diff
@@ -10,25 +10,51 @@ class StillPending < StandardError
...
     dqt_trn_request.pending! if dqt_trn_request.initial?
 
     if (trn = response[:trn]).present?
-      ActiveRecord::Base.transaction do
-        dqt_trn_request.application_form.teacher.update!(trn:)
-        dqt_trn_request.complete!
-      end
+      update_teacher_trn(dqt_trn_request, trn)
+      send_award_email(dqt_trn_request)
+      update_application_form(dqt_trn_request)
```
gpeng, line 20: I think we should extract this out and think about the failure conditions a bit. I think as it is if the application form update fails we could end up sending multiple emails to the applicant. It think we'd want to do the updates in a transaction and only email if the transaction commits. We could probably do that in here but I think we're going beyond the job's responsibility at that point.

thomasleese: My initial implementation had everything in a transaction, but then I was worried about what would happen if creating the email job failed, but the transaction had succeeded (let's say for example the Sidekiq process gets restarted at that exact point). I figured it was better that there's a chance the teachers will receive a duplicate email than not receive the email at all.

I reordered things so the email is sent before the status is updated, but there's a guard clause before sending the email to check we haven't already tried to send one before.

I also wondered about extracting this logic out in to a service, but because of the careful ordering, it felt like the service was then just doing everything the job did.

Happy to talk about it further though!

gpeng (approved):

PR #667, opened 2022-11-02
Show all errors on upload form if one file is missed

thomasleese:
We lose any uploaded files when re-rendering the form (without using JavaScript) so users may not notice that they've got to re-upload their file. For that reason, although confusing, showing to the user both error messages should avoid that.

[Trello Card](https://trello.com/c/QNLAp8oF/1076-full-journey-snags)

## Screenshot

Screenshot 2022-11-02 at 11 31 56


David-Feetenby (approved):

gpeng (approved):
We really need to rethink the UX with this. This seems pragmatic for now though.

PR #679, opened 2022-11-03
Assessment sections error

David-Feetenby:
Update assessor interface assessment section so that the validation is done correctly

### DIFF
--- a/config/locales/assessor_interface.en.yml
+++ b/config/locales/assessor_interface.en.yml
@@ -172,9 +172,11 @@ en:
               blank: Enter the subject
             selected_failure_reasons:
               blank: Select the reasons for your recommendation
+            passed:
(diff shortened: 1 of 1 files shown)


config/locales/assessor_interface.en.yml
```diff
@@ -172,6 +172,8 @@ en:
               blank: Enter the subject
             selected_failure_reasons:
               blank: Select the reasons for your recommendation
+            passed:
+              inclusion: Choose whether the applicant completed this section to your satisfaction
```
thomasleese, line 176:
```suggestion
              inclusion: Select whether the applicant completed this section to your satisfaction
```

I think the convention is to use the word select.

David-Feetenby: Changed 👍

thomasleese (approved):

PR #694, opened 2022-11-07
[1107] Only show assessors in the filters and assign pages

gpeng:
We currently show all staff users in the filters and assign assessor/reviewer pages. We need to be able to create support users that don't appear in this list.

### Changes

* Add a boolean `assessor?` field to the staff model. Default to false
* Add a scope `Staff.assessors` that returns assessor == true
* Use the scope on these pages

I've not validated that assignees are assessors as yet as we need to think that through but this will fix the UI for now.

### DIFF
--- a/app/models/staff.rb
+++ b/app/models/staff.rb
@@ -3,6 +3,7 @@
 # Table name: staff
 #
 #  id                     :bigint           not null, primary key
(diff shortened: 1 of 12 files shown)


steventux (approved):
:100:

PR #696, opened 2022-11-07
[1109] Whingy image validation

gpeng:
Update upload validation to be more permissive.

* Support uppercase extensions
* Support jpeg as an extension
* Add translations for the error message

## Review

I'm not sure that the mismatch validation is actually working. We can review that in another PR as we need to get this fix in before we launch ideally.


config/locales/teacher_interface.en.yml
```diff
@@ -185,8 +185,14 @@ en:
           attributes:
             original_attachment:
               blank: Select a file to upload
+              invalid_content_type: we can only accept files in PDF, JPG, PNG, DOCX or DOC format
+              mismatch_content_type: The file type and file name do not match. For example, if the upload is called example.doc, it must be a DOC file.
+              file_size_too_big: files cannot be larger than 50MB
```
tvararu, line 190:
```suggestion
              invalid_content_type: Files must be in PDF, JPG, PNG, DOCX or DOC format
              mismatch_content_type: Files must have matching file type and file names. For example, if the upload is called example.doc, it must be a DOC file
              file_size_too_big: Files must not be larger than 50MB
```

config/locales/teacher_interface.en.yml
```diff
@@ -185,8 +185,14 @@ en:
...
             translated_attachment:
               blank: Select a translation to upload
+              invalid_content_type: we can only accept files in PDF, JPG, PNG, DOCX or DOC format
+              mismatch_content_type: The file type and file name do not match. For example, if the upload is called example.doc, it must be a DOC file.
+              file_size_too_big: files cannot be larger than 50MB
```
tvararu, line 195:
```suggestion
              invalid_content_type: Files must be in PDF, JPG, PNG, DOCX or DOC format
              mismatch_content_type: Files must have matching file type and file names. For example, if the upload is called example.doc, it must be a DOC file
              file_size_too_big: Files must not be larger than 50MB
```

tvararu (approved):
Just some content suggestions based on https://design-system.service.gov.uk/components/error-message/#use-instructions-and-descriptions

PR #704, opened 2022-11-08
Update support country page layout

gpeng:
We need to be able to see which regions are enabled for applications.

* Layout in a table for better legibility
* Add eligibility inactive tags
* Add application form tags

Screenshot 2022-11-08 at 12 17 45


thomasleese (approved):

thomasleese:
You could be a designer!

PR #715, opened 2022-11-09
Avoid duplicate assessment section recorded events

thomasleese:
If the state is the same, we could avoid recording a assessment section recorded event since it would say the same thing.

[Trello Card](https://trello.com/c/mJFRfgzr/1119-a-few-more-snags-%F0%9F%90%9F)

### DIFF
--- a/app/services/update_assessment_section.rb
+++ b/app/services/update_assessment_section.rb
@@ -28,13 +28,16 @@ class UpdateAssessmentSection
   attr_reader :assessment_section, :user, :params
 
   def create_timeline_event(old_state:)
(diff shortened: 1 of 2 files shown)


thomasleese:
I'm raising this PR as a potential solution to duplicate events, however I don't think it's necessarily a good idea, would be interested in thoughts:

- The timeline is no longer accurate.
- We lose a record of when notes are changed (i.e. the form is resubmitted) but the state is the same.

I did think about doing something more intelligent where in the case of the state staying the same, we find the existing timeline event and update the time on it, however again this feels like we're messing with an audit log.

PR #722, opened 2022-11-10
[1132] Remove anonymous support user

gpeng:
We have persona based logins in test and un/pw authentication in production now so we no longer need the basic auth anonymous support user functionality.

* Remove basic auth strategy
* Remove feature flag
* Fix system specs that used basic auth to login as a staff user


thomasleese (approved):
I think it would be worth writing some documentation on how to enable the personas without access to the support console (as would be when a user first runs the app).

thomasleese (approved):

PR #738, opened 2022-11-15
Verify age range and subjects before recommendation

thomasleese:
This allows the assessor to consider the age range and subjects as part of the award flow, after any potential FI requests have come through rather than before.

Depends on #737 

[Trello Card](https://trello.com/c/Ufpvs9DD/1154-move-subject-and-age-assessment-to-after-fi-received)

## Screenshot

Screenshot 2022-11-15 at 18 07 34
Screenshot 2022-11-15 at 18 07 43


gpeng:
Closing this for now as we're going to have a rethink when the assessors have had a bit more time with the current implementation.

PR #741, opened 2022-11-16
[1155] FI request decline page email update new email content

gpeng:
When an application is declined following an FI request we want to send a different email as we've already informed the user of the initial reasons with the FI request.

* Send an alternative email

NB: will update with screen shots and link when the review app deploys.


app/mailers/teacher_mailer.rb
```diff
@@ -31,6 +32,14 @@ def application_declined
     )
   end
 
+  def application_declined_following_further_information_request
```
thomasleese, line 35: This name, not having it's own locale key for the subject, will break the timeline entry component:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/be8908619a78fd1f5998312ac893f5b00a3e0b53/app/components/timeline_entry/component.rb#L81-L86

I don't mind this approach, but I wonder if it would be better if we passed in `further_information: true` or something in the `params` instead of having an entirely new view, and then wrote some `erb` in the view to customise the content. It feels to me like these emails are similar enough that would be a better approach, and means content common to both of them would stay in sync.

gpeng: I'll sort the key out. I did think of the approach you're suggesting. I decided it was better to make it explicit that they're treated differently. I don't really mind either way though.

thomasleese (approved):

gpeng:
This got a bit more complicated as I needed to update the preview too. I 'had a rethink' about passing the FI or not as a parameter to the `application_declined` mailer but it still needed the knowledge of that to be to be in a couple of places and would still require changes to the preview component/view and the service so I've gone with a helper to encapsulate that logic. Happy to discuss.

thomasleese (approved):
Ah yeah, that makes sense about the preview stuff, I think I'd personally still prefer a single view with a parameter, but I don't feel strongly, and happy for this to be merged in.

gpeng:
[FI request decline page & email update](https://trello.com/c/Y5x1Stwa/1155-fi-request-decline-page-email-update)

PR #742, opened 2022-11-16
Ensure emails are recorded as timeline events

thomasleese:
The observer wasn't being registered correctly so I've changed to a different approach which seems to work now. I've tested this by sending an email (both sync and async) in a Rails console and seeing that a timeline event is created for both of them.

## Screenshot

Screenshot 2022-11-16 at 10 49 04

### DIFF
--- a/app/lib/teacher_mailer_observer.rb
+++ b/app/lib/teacher_mailer_observer.rb
@@ -15,3 +15,5 @@ class TeacherMailerObserver
     )
   end
 end
+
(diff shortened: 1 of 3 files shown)


David-Feetenby (approved):

gpeng:
[Show FI request emails on the timeline](https://trello.com/c/ExDPHj8v/1161-show-fi-request-emails-on-the-timeline)

PR #744, opened 2022-11-17
1162 assessor interface roles

David-Feetenby:
This pr creates the data structure for assessor interface roles as well as adding the option to the invite ui.

Additionally, this adds the new award_decline boolean to the existing assessor personas whilst also creating a new persona for an admin.


David-Feetenby:
[Assessor interface roles](https://trello.com/c/7nteBPtG/1162-assessor-interface-roles)

app/views/staff/invitations/new.html.erb
```diff
@@ -8,6 +8,9 @@
   <% end -%>
 
   <%= f.govuk_text_field :name %>
-
+  <br>
```
gpeng, line 11: We don't need the br-ses. The form builder should take care of the spacing for us. We could probably put these in a checkbox group with the title 'Permissions' and get rid of the 'access' and 'permissions' from the two check box labels.

lib/tasks/example_data.rake
```diff
@@ -67,11 +67,28 @@ end
...
+    },
   ]
 end
 
+def admin
+  {
+    name: "Sally Admin",
+    email: "admin-sally@example.com",
+    support_console_access: true,
+    award_decline: true,
```
gpeng, line 88: Might be good to have this set to false as once we implement the authorisation checks we'll want a user that doesn't have award_decline.

spec/factories/staff.rb
```diff
@@ -53,8 +54,12 @@
       confirmed_at { Time.zone.now }
     end
 
-    trait :assessor do
-      assessor { true }
+    trait :award_decline do
```
gpeng, line 57: Shall we call these `with_award_decline` and `with_support_console_access`? Reads a bit better. Possibly even with a `_permission` suffix?

lib/tasks/example_data.rake
```diff
@@ -67,11 +67,28 @@ end
...
+    },
+    {
+      name: "Beryl Assessor",
+      email: "assessor-beryl@example.com",
+      award_decline: true,
+    },
   ]
 end
 
+def admin
```
gpeng, line 83: Can we show this user on the personas page so as we can log in as it?

gpeng (changes requested):
Looking good. All working fine. Just a few cosmetic things really. Could we also show the permissions a user has on the Staff index page?

app/views/staff/invitations/new.html.erb
David-Feetenby: The spacing was not correct  before the br tag. I'll put them in a checkbox group and maybe it will be correct without them, then.

Screenshot 2022-11-21 at 08 48 28

David-Feetenby:
> Looking good. All working fine. Just a few cosmetic things really. Could we also show the permissions a user has on the Staff index page?

done 👍

app/views/support_interface/staff/index.html.erb
```diff
@@ -20,6 +20,16 @@
       row.value { staff.created_at.to_s }
     end
 
+    summary_list.row do |row|
+      row.key { "Assessor" }
+      row.value { staff.with_award_decline_permission? ? "Yes" : "No" }
```
thomasleese, line 25: For being able to visually see this information quickly, I wonder if it's worth using [`govuk_tag`](https://govuk-components.netlify.app/components/tag/) similar to show we do below for accepted and not accepted.

spec/factories/staff.rb
thomasleese: I think the suggestion here was just to change the factory trait name rather than the database field, but actually I like the `_permission` prefix on the database field names but I don't think we need the `with_`, so something like:

```rb
# using the trait
create(:staff, :with_award_decline_permission)

# which would be equivalent to
create(:staff, award_decline_permission: true)
```

app/views/support_interface/staff/index.html.erb
```diff
@@ -19,7 +19,26 @@
...
       row.value { staff.created_at.to_s }
     end
+    
+      summary_list.row do |row|
+      row.key { "Assessor" }
+      row.value { if staff.award_decline_permission?
+          govuk_tag(text:"Yes", colour: "green")
+        else
+          govuk_tag(text:"No", colour: "red")
+        end }
```
thomasleese, line 29:
```suggestion
      row.value do
        if staff.award_decline_permission?
          govuk_tag(text: "Yes", colour: "green")
        else
          govuk_tag(text: "No", colour: "red")
        end
      end
```

I think the formatting on this code is a little off.

app/views/support_interface/staff/index.html.erb
```diff
@@ -19,7 +19,26 @@
...
 
+    summary_list.row do |row|
+      row.key { "Support console access" }
+      row.value { 
+        if staff.support_console_permission?
+          govuk_tag(text:"Yes", colour: "green")
+        else
+          govuk_tag(text:"No", colour: "red")
+        end
+      }
```
thomasleese, line 40:
```suggestion
      row.value do
        if staff.support_console_permission?
          govuk_tag(text:"Yes", colour: "green")
        else
          govuk_tag(text:"No", colour: "red")
        end
      end
```

lib/tasks/example_data.rake
```diff
@@ -14,12 +14,16 @@ namespace :example_data do
...
+    staff_members.each do |staff|
       FactoryBot.create(
         :staff,
         :confirmed,
-        name: assessor[:name],
-        email: assessor[:email],
+        name: staff[:name],
+        email: staff[:email],
+        award_decline_permission: staff[:award_decline_permission] || false,
+        support_console_permission: staff[:support_console_permission] || false,
```
thomasleese, line 26:
```suggestion
        **staff
```

This should work to automatically create a staff instance with the parameters in the `staff` hash.

spec/models/staff_spec.rb
```diff
@@ -65,13 +66,15 @@
     describe ".assessors" do
       subject { described_class.assessors }
 
-      context "when assessor == true" do
-        let(:assessor) { create(:staff, :assessor) }
+      context "when with_award_decline_permission == true" do
```
thomasleese, line 69:
```suggestion
      context "when award_decline_permission is true" do
```

Usually the contexts are written as words.

gpeng (approved):

lib/tasks/example_data.rake
David-Feetenby: I think this is done, if you want to review it

thomasleese (approved):

PR #749, opened 2022-11-21
[1143] Expire further information requests

gpeng:
Further information requests have a time limit after which they expire and the application is automatically declined. For most regions this is 6 weeks from application submission but for some it is 4 weeks.

* Add an expired state to FurtherInformationRequest (and a factory trait for this)
* Add a timeline event type and support for it on the timeline component
* Add `FurtherInformationRequestExpirer` service. This checks the date on the FI request and if it should be expired it:
  * updates the assessment state to declined (which creates an timeline entry and emails the user)
  * updates the application form state
  * creates a specific timeline entry for the expiry
  * sets the FI request to 'expired'
  * sets the FI request assessor note accordingly
* Add two jobs to retrieve FI requests to check and then run them through the service
* Schedule the 'outer' job to run daily at 2am
* Slight tweak to the decline email, we're going to just use the standard content but the 'thank you' bit has been moved so as it is in both versions
* Update the teacher interface application show to include a note from the assessor if they FI request has expired

There is also work on the ticket for reminder emails but I've split this off and opened another ticket for that.

### DIFF
--- a/app/views/teacher_interface/application_forms/show.html.erb
+++ b/app/views/teacher_interface/application_forms/show.html.erb
@@ -65,6 +65,10 @@
 <% elsif @view_object.application_form.declined? %>
   <h2 class="govuk-heading-l">Your QTS application has been declined</h2>
 
+  <%- if @view_object.show_further_information_request_expired_content? -%>
+    <%= govuk_inset_text(text:t(".further_information_request_expired")) %>
+  <%- end -%>
+  
   <% if (sections = @view_object.notes_from_assessors).present? %>
     <h3 class="govuk-heading-m">Notes from the assessor:</h3>
 
--- a/app/components/timeline_entry/component.rb
+++ b/app/components/timeline_entry/component.rb
@@ -74,6 +74,16 @@ module TimelineEntry
       }
     end
 
+    def further_information_request_expired_vars
+      {
+        further_information_request: timeline_event.further_information_request,
+        date_requested:
+          timeline_event.further_information_request.created_at.strftime(
+            "%e %B %Y at %l:%M %P",
+          ),
+      }
+    end
+
     def email_sent_vars
       {
         subject:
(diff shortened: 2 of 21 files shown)


gpeng:
[FURTHER INFO: Decline a timed out Further Information request](https://trello.com/c/alRkDJR9/1143-further-info-decline-a-timed-out-further-information-request)

app/views/teacher_interface/application_forms/show.html.erb
```diff
@@ -65,6 +65,10 @@
 <% elsif @view_object.application_form.declined? %>
   <h2 class="govuk-heading-l">Your QTS application has been declined</h2>
 
+  <%- if @view_object.show_further_information_request_expired_content? -%>
+    <%= govuk_inset_text(text:t("assessor_interface.assessment_sections.show.failure_reasons.further_information_request_expired")) %>
```
thomasleese, line 69: Since this key will never be selected by an assessor, or set in the `failure_reason`, I wonder if this string belongs somewhere else, specifically we could do something like:

```rb
t(".further_information_request_expired")
```

And then set the key to: `assessor_interface.application_forms.show.further_information_request_expired `.

thomasleese (commented):
Looks good, I've left some comments, but they're all pretty minor!

app/views/teacher_interface/application_forms/show.html.erb
gpeng: I wondered that too. If we do that then the possible failure reasons are now in two places which is potentially confusing too. We might have a need to store them somewhere else altogether as they're not presenting very well in the analytics. Let's discuss this later.

thomasleese: Yeah, happy to discuss it.

I think in this case I'd argue that this isn't a "failure reason" as we've defined them, as they're specifically related to the initial assessment stage and associated with assessment sections, rather than further information requests. The `failure_reason_key` on the further information request comes directly from the assessment section after all.

This is more of a "further information request declined reason", i.e. it's the `failure_reason_assessor_feedback` equivalent for an FI request which we currently don't have a concept of.

gpeng:
Thanks @thomasleese . I've addressed those comments.

thomasleese (approved):

PR #750, opened 2022-11-22
Add teaching authority online checker URL

thomasleese:
This adds a new field to regions and countries which we can use the set the online checker URL, relevant if the teaching authority has an online checker for teachers.

[Trello Card](https://trello.com/c/nDtRrHHk/1135-add-online-checker-url-field-to-case-management-via-support-console)

## Screenshots

Screenshot 2022-11-22 at 10 44 19

Screenshot 2022-11-23 at 08 13 52

Screenshot 2022-11-22 at 10 56 06


thomasleese:
> Should we add an error message to the locale. It looks ok inline in the screenshot but the summary will be weird.

Good point, I've added that in now.

PR #757, opened 2022-11-23
Add authorisation to support and assessor interface

thomasleese:
This uses the [Pundit](https://github.com/varvet/pundit) library to add authorisation to staff users. Specifically, we require that a staff user has the `support_console_permission` to access the support console, and that a staff user has the `award_decline_permission` to perform any actions in the assessor interface.

[Trello Card](https://trello.com/c/YV473OCw/1163-assessor-interface-authorization)


app/views/errors/forbidden.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= content_for :page_title, "Forbidden" %>
+
+<h1 class="govuk-heading-l">Forbidden</h1>
```
gpeng, line 3: This is a bit shouty. I realise it'll mostly be internal users that may see it but maybe we should review the content.

gpeng (approved):
Looks good and all seems to be working. I wonder if we should follow up with hiding some of the things, links etc that people can't do. Not sure that's a priority though at the minute. I didn't test with a no-permission staff user but it looks like it should be good on that front.

app/views/errors/forbidden.html.erb
thomasleese: Yep agree, I'll chat to the content designers about improving this, but will probably merge this in for now and then we can improve the content in a later PR.

thomasleese:
> Looks good and all seems to be working. I wonder if we should follow up with hiding some of the things, links etc that people can't do. Not sure that's a priority though at the minute. I didn't test with a no-permission staff user but it looks like it should be good on that front.

Yeah, agree that we should follow up with hiding links, etc.

PR #761, opened 2022-11-23
[1183] Further information request reminder emails

gpeng:
Send a reminder email at 2 weeks, 1 week and 2 days before the expiry date. Expiry can be 6 weeks from request or 4 weeks for `FOUR_WEEK_COUNTRY_CODES` countries.

NB: this is currently on top of the expiry PR. I'll sort that out when I merge the other one.

* Add the `SendFurtherInformationRequestReminder` service
* Add `ReminderEmail` to allow us to check what emails have already been sent
* Add timeline support
* Job to call that ^ and job to queue up those <
* Refactor the expired service as it also needs the FI expiry logic so I've extracted it into a concern

### DIFF
--- a/app/mailers/teacher_mailer.rb
+++ b/app/mailers/teacher_mailer.rb
@@ -6,8 +6,10 @@ class TeacherMailer < ApplicationMailer
                   application_declined
                   application_received
                   further_information_received
+                  further_information_reminder
                 ]
   before_action :set_further_information_requested, only: :application_declined
+  before_action :set_due_date, only: :further_information_reminder
   after_action :store_observer_metadata
 
   GOVUK_NOTIFY_TEMPLATE_ID =
@@ -56,6 +58,14 @@ class TeacherMailer < ApplicationMailer
     )
   end
 
+  def further_information_reminder
+    view_mail(
+      GOVUK_NOTIFY_TEMPLATE_ID,
+      to: params[:teacher].email,
+      subject: I18n.t("mailer.teacher.further_information_reminder.subject"),
+    )
+  end
+
   private
 
   def set_name
@@ -64,6 +74,10 @@ class TeacherMailer < ApplicationMailer
 
   def set_reference
     @reference = params[:teacher].application_form.reference
+  end
+
+  def set_due_date
+    @due_date = params[:due_date]
   end
 
   def set_further_information_requested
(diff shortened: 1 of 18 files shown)


app/mailers/teacher_mailer.rb
```diff
@@ -56,6 +57,15 @@ def further_information_requested
     )
   end
 
+  def further_information_reminder(due_date:)
```
gpeng, line 60: I think I prefer passing this argument in here rather than as an argument to `.with(` but it's inconsistent with the `further_information_request` argument. I'll probably change it to be consistent 🤷‍♂️

gpeng:
[FURTHER INFO: FI request reminder emails](https://trello.com/c/lBIS3Moa/1183-further-info-fi-request-reminder-emails)

app/mailers/teacher_mailer.rb
thomasleese: My understanding is that we have to use `.with` if we want to use `deliver_later` as the parameters get serialised and then the email sending job calls `TeacherMailer.with(deserialised_params).further_information_reminder`.

app/services/further_information_request_reminder.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def create_timeline_event
+    TimelineEvent.create!(
+      application_form:,
+      creator_name: "Expirer",
+      mailer_action_name: "further_information_reminder",
+      event_type: "email_sent",
+    )
+  end
```
thomasleese, line 61: We shouldn't need to do this as https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/app/lib/teacher_mailer_observer.rb will create this for us.

db/migrate/20221122162438_create_reminder_emails.rb
```diff
@@ -0,0 +1,8 @@
+class CreateReminderEmails < ActiveRecord::Migration[7.0]
+  def change
+    create_table :reminder_emails do |t|
+      t.references :further_information_request
```
thomasleese, line 4:
```suggestion
      t.references :further_information_request, null: false, foreign_key: true
```

I'd suggest this to ensure data consistency (I'm not really sure why it's not the default 😆).

thomasleese (commented):
Generally looks good, I've left a few comments.

app/mailers/teacher_mailer.rb
gpeng: Ah right. That makes sense. I did not know that. I'll change it.

thomasleese (approved):

thomasleese (approved):

PR #768, opened 2022-11-25
Store filter parameters in session

thomasleese:
Rather than in query parameters and needing to remember to pass them around, we can store these in the session which means the filters a user has selected will always remain the same regardless of how they navigate around the site.

[Trello Card](https://trello.com/c/rOuiwQaH/1199-stickier-filters-in-case-management)

### DIFF
--- a/app/components/application_form_search_result/component.rb
+++ b/app/components/application_form_search_result/component.rb
@@ -1,9 +1,8 @@
 module ApplicationFormSearchResult
   class Component < ViewComponent::Base
-    def initialize(application_form:, search_params:)
+    def initialize(application_form:)
       super
       @application_form = application_form
-      @search_params = search_params
     end
 
     def full_name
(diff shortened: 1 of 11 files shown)


gpeng (approved):
This works and the code is fine. I'm not sure of the approach of storing them in the session though. It's going to be confusing in multiple tabs and we lose the ability to bookmark or share URLs. I also find it jarring that the filter is always there even if I've been all over the place since using it. It makes sense to me if I'm going 'back' to a list I was working from. All of this is a lot 'just me' and I understand the requirement. I've approved but would be interested in some UR (and/or discussion).

thomasleese:
> This works and the code is fine. I'm not sure of the approach of storing them in the session though. It's going to be confusing in multiple tabs and we lose the ability to bookmark or share URLs. I also find it jarring that the filter is always there even if I've been all over the place since using it. It makes sense to me if I'm going 'back' to a list I was working from. All of this is a lot 'just me' and I understand the requirement. I've approved but would be interested in some UR (and/or discussion).

Yeah, generally I'd prefer not to store this sort of thing in the session either, but it sounds like the way the assessors use it this makes more sense for them. Happy to discuss further though.

thomasleese:
Product is happy with this approach, so I'll merge it now - if it's problematic we can change it in the future.

PR #779, opened 2022-11-28
Improve application form back links

thomasleese:
This improves how back links work to ensure that the user is always taken to the right place. We will now store a history of where the user has been in the journey in the session and use that to determine the user should go when they click on a back link. For the most part, this is the previous entry in the stack, but from some pages (notably check your answers) we send the user back to the first "origin" page - which in most cases will be the overview task list.

This implements the journeys as described in this diagram: https://drive.google.com/file/d/1wAerFpEZVQpsu3otSc-Ib3HsDsM_bxXh/view?ts=6384846c

I'm anticipating a follow up PR from this which replaces the `?next=` params with the history stack.

[Trello Card](https://trello.com/c/zJbHelII/1208-back-link-redo)

### DIFF
--- a/app/controllers/concerns/history_trackable.rb
+++ b/app/controllers/concerns/history_trackable.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module HistoryTrackable
+  extend ActiveSupport::Concern
+
+  included { before_action :push_self, if: -> { request.get? } }
+
+  def push_self
+    history_stack.push_self(request, origin: false)
+  end
+
+  def push_self_as_origin
+    history_stack.push_self(request, origin: true)
+  end
+
+  def push_self_as_origin_and_reset
+    history_stack.push_self(request, origin: true, reset: true)
+  end
+
+  private
+
+  def history_stack
+    @history_stack ||= HistoryStack.new(session:)
+  end
+end
--- a/app/controllers/history_controller.rb
+++ b/app/controllers/history_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class HistoryController < ApplicationController
+  def back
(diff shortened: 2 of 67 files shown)


gpeng (commented):
I think this is close. I've found a few instances on the review app were pages aren't being included in the stack which I think is probably just a case of pushing them. (I went through the first few assessor sections in the review app) I do think that highlights a bit of an issue with this approach though in that we're going to have to keep on top of making sure to include new pages in the back links and that's quite complicated to understand. We'll need some docs if we go with this route but I wonder we should push everything (in application controller or rack or somewhere) and then exclude if needed. Also maybe we could define the origin routes centrally somewhere. The amount of control we get doing it explicitly is good though so maybe that's better. Happy to discuss.

thomasleese:
> I think this is close. I've found a few instances on the review app were pages aren't being included in the stack which I think is probably just a case of pushing them. (I went through the first few assessment sections in the review app) I do think that highlights a bit of an issue with this approach though in that we're going to have to keep on top of making sure to include new pages in the back links and that's quite complicated to understand. We'll need some docs if we go with this route but I wonder we should push everything (in application controller or rack or somewhere) and then exclude if needed. Also maybe we could define the origin routes centrally somewhere. The amount of control we get doing it explicitly is good though so maybe that's better. Happy to discuss.

I actually left the assessor interface out of this PR and focussed just on the teacher interface. We can use the same code in the assessor interface, but I was going to leave it to a separate PR.

I do like the idea of automatically adding everything to the stack, but it's get complicated I think where we would usually only want `GET` requests to be on the stack, however I can take a look at what can be done automatically.

thomasleese:
I've replaced this with #787 as I believe part of the problem was coming from still having a `next` param in the query parameters.

PR #781, opened 2022-11-29
[1209] Split notes from decline reasons - Part 1

gpeng:
We currently store assessment failure reasons and the associated assessor notes as a JSON field on the `AssessmentSection` record. This makes it tricky to analyse and to filter PII from.

We are going to separate these things out into a set of associated records.

As we already have existing data and this functionality is running in production we'll split it into 3 or 4 PRs and deploy them sequentially.

* Start saving the new records
* Copy existing data into the new records
* Update the functionally to use the new records
* Delete the old data and fix the temporary naming of the new association (`assessment_section_failure_reasons` -> `selected_failure_reasons` or something like that)

This is the first bit to start saving the new records to `AssessmentSectionFailureReason`

### DIFF
--- a/app/models/assessment_section.rb
+++ b/app/models/assessment_section.rb
@@ -23,6 +23,7 @@
 #
 class AssessmentSection < ApplicationRecord
   belongs_to :assessment
+  has_many :assessment_section_failure_reasons
 
   enum :key,
        {
--- a/app/models/assessment_section_failure_reason.rb
+++ b/app/models/assessment_section_failure_reason.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: assessment_section_failure_reasons
+#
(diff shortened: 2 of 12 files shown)


app/models/assessment_section.rb
```diff
@@ -23,6 +23,7 @@
 #
 class AssessmentSection < ApplicationRecord
   belongs_to :assessment
+  has_many :assessment_section_failure_reasons
```
thomasleese, line 26: I wonder if we should rename `failure_reasons` to `available_failure_reason_keys` at some point.

thomasleese (approved):

app/models/assessment_section.rb
gpeng: Yeah. We should have a think about the naming. I'm still not really into 'failure_` either but that's probably stuck now. I'll update them to something in the last PR when the old implementation is gone which will free up `selected_failure_reasons`.

PR #787, opened 2022-11-30
Improve application form back links

thomasleese:
This improves how back links work to ensure that the user is always taken to the right place. We will now store a history of where the user has been in the journey in the session and use that to determine the user should go when they click on a back link. For the most part, this is the previous entry in the stack, but from some pages (notably check your answers) we send the user back to the first "origin" page - which in most cases will be the overview task list.

This implements the journeys as described in this diagram: https://drive.google.com/file/d/1wAerFpEZVQpsu3otSc-Ib3HsDsM_bxXh/view?ts=6384846c

[Trello Card](https://trello.com/c/zJbHelII/1208-back-link-redo)

### DIFF
--- a/app/components/check_your_answers_summary/component.rb
+++ b/app/components/check_your_answers_summary/component.rb
@@ -24,7 +24,7 @@ module CheckYourAnswersSummary
     end
 
     def delete_link_to
-      path_with_next(@delete_link_to) if changeable && @delete_link_to
+      @delete_link_to if changeable && @delete_link_to
     end
 
     private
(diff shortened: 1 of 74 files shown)


gpeng:
I think this is better and seems to be working well. What do you think about splitting the act of pushing and defining the origin pages? So instead of skipping this stuff:

```
    skip_before_action :push_self
     before_action :push_self_as_origin_and_reset, only: :show
     before_action :push_self_as_check, only: :edit
```

We'd just have `push_self` on all gets but then have a different macro thing for defining an origin/check e.g.

```
define_back_link_origins :show
define_back_link_checks :edit
```

The push_self method in the module would add the necessary params based on the definitions. Might be a bit cleaner in the controllers and a bit more explicit?

I think we could still do with some docs about when to define an origin/check and what that means too.

thomasleese:
@gpeng I really like that suggestion, I've amended the PR so it works like that and added some documentation.

gpeng (approved):
Looks great. I've had a good click around and it seems to be working really well too.

PR #797, opened 2022-12-02
Calculate more working days

thomasleese:
This adds more working days calculations to help with analytics and SLA calculations:

- submission date to recommendation date
- submission date to initial assessment date
- initial assessment to recommendation
- FI received date to recommendation

[Trello Card](https://trello.com/c/GhqRqMET/1222-record-working-days-since-assessor-assigned)


app/jobs/update_working_days_job.rb
```diff
@@ -0,0 +1,102 @@
...
+    @calendar ||= Business::Calendar.load("england_and_wales")
+  end
+
+  def today
+    @today ||= Time.zone.now
+  end
+
+  def update_application_forms_working_days_since_submission
+    ApplicationForm
+      .where.not(submitted_at: nil)
```
gpeng, line 24: can we use the `.submitted` scope here?

gpeng: Actually, I see now, this would include other post submission states too. I think we need to exclude completed applications don't we or this will keep updating forever? Maybe we need some scopes that wrap the sort of 'ongoing' and 'completed' states.

app/jobs/update_working_days_job.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+class UpdateWorkingDaysJob < ApplicationJob
+  def perform
```
gpeng, line 4: I wonder if we should use the fan out sort of pattern for this so as one failure doesn't block the whole lot. Maybe it's not worth the extra complexity here.

thomasleese: Yeah, I thought about it and figured it made sense to keep it simple for now and build a more complex solution if we start seeing problems.

app/jobs/update_working_days_job.rb
thomasleese: We can't exclude completed ones, as it's possible (although unlikely) that an application form will go from submitted state to awarded state all within one day, and therefore the job wouldn't have sight of it.

You're right that the job does more work than it necessarily needs to (once an application is completed the `submitted_at` date is unlikely to change) but I wanted the job to be able to respond to us needing to manually change any dates without then needing to re-run this job. The idea was the job filters the queries only on what's required to perform the calculation (i.e. the date fields). I actually wondered about _not_ creating the scopes I did for assessments, and keeping it all explicit in this code?

gpeng: Fair enough 👍

gpeng (approved):

PR #799, opened 2022-12-02
[1209] Refactor failure reason definitions

gpeng:
We use these in a number of places in the service for various things.

* Move them to one place... `FailureReasons`. 

We do also group these in some of the factory classes when building an assessment and failure reasons. I haven't moved those here as I don't need that currently for the work to move to using `AssessmentSectionFailureReason`. There's a reasonable argument for moving the section grouping into `FailureReasons` too but we can do that in a future PR.


thomasleese (approved):
Looks good. I wonder about including https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/d21cde6cc475d513906c0312e4fae7e6cedf778a/app/lib/further_information_request_items_factory.rb#L51-L65 in the class, but I'm not 100% sure about it.

gpeng:
Aye. I think that and the reasons for each section are probably good candidates to move. I think I'll merge this as it will help with the `selected_failure_reasons` thing I'm doing and then I'll card up some further work for when that's done.

PR #802, opened 2022-12-05
Add docs on rolling back an application

thomasleese:
If an application has a further information request and we want to re-send the further information request, this explains how to do that.


docs/application-fixes.md
```diff
@@ -29,3 +29,31 @@ CreateDQTTRNRequest.call(application_form: application_form)
...
+# find the application form and user who is performing the change
+application_form = ApplicationForm.find(...)
+user = Staff.find_by(name: ...)
+
+# set some useful variables
+assessment = application_form.assessment
+further_information_request = assessment.further_information_requests.first
+
+# delete the FI requests and reset the assessment state
+TimelineEvent.where(further_information_request:).destroy_all
```
gpeng, line 53: Could we do soft deletion on the FI request to avoid this? It probably doesn't matter in the application we're doing at the mo but changing the application timeline doesn't seem like the right thing to do generally.

gpeng: Discussed offline. This is what we have to do 'now'. If/when we expand FI to allow soft deletion or multiple requests we can update.

gpeng (approved):

PR #812, opened 2022-12-07
Add Content for new Status Sanction fields

David-Feetenby:
This commit adds the sanction and status information to the relevant views


app/controllers/support_interface/countries_controller.rb
```diff
@@ -71,6 +71,8 @@ def country_params
       :teaching_authority_websites_string,
       :teaching_authority_certificate,
       :teaching_authority_other,
+      :teaching_authority_sanction_information,
+      :teaching_authority_status_information,
```
thomasleese, line 75: Does this need deploying separately first? It sounds like perhaps without it the support interface isn't working correctly?

app/controllers/teacher_interface/documents_controller.rb
```diff
@@ -50,5 +50,17 @@ def update
...
+
+    def load_teaching_authority_status_information
+      @teaching_authority_status_information =
+        application_form.region.teaching_authority_status_information
+    end
+
+    def load_teaching_authority_sanction_information
+      @teaching_authority_sanction_information =
+        application_form.region.teaching_authority_sanction_information
+    end
```
thomasleese, line 64: It looks like these aren't called anywhere.

app/views/teacher_interface/registration_number/edit.html.erb
```diff
@@ -7,14 +7,30 @@
   <%= f.govuk_text_field :registration_number, label: { size: "l", tag: "h1" } %>
 
   <% if @teaching_authority_other.present? %>
-    <%= raw GovukMarkdown.render(@teaching_authority_other) %>
+    <%= raw GovukMarkdown.render(@application_form.region.teaching_authority_other) %>
+  <% end %>
```
thomasleese, line 11: We should no longer be rendering the `teaching_authority_other` on this page.

app/views/teacher_interface/registration_number/edit.html.erb
```diff
@@ -7,14 +7,30 @@
...
       contact you for more information as part of your application.
     </p>
+
+    <% if @application_form.region.status_check_online? %>
+      <%= raw GovukMarkdown.render(@application_form.region.teaching_authority_status_information) %>
+    <% end %>
+
+    <% if @application_form.region.sanction_check_online? %>
+      <%= raw GovukMarkdown.render(@application_form.region.teaching_authority_sanction_information) %>
+    <% end %>
```
thomasleese, line 33: It looks like the code here is duplicating the lines above, but actually we don't need to render the content within the summary/detail component.

David-Feetenby:
[Fix "teaching authority other" field](https://trello.com/c/roDCVI5L/1219-fix-teaching-authority-other-field)

PR #816, opened 2022-12-07
Sign in teachers using a one time password

thomasleese:
This implements a "one-time password" authentication mechanism for teachers based on the same code used in https://github.com/DFE-Digital/refer-serious-misconduct.

Depends on #815 

[Trello Card](https://trello.com/c/nLrL1b0P/1235-update-auth-to-use-otp-instead-of-not-so-magic-links)

## Screenshots

Screenshot 2022-12-09 at 11 53 37

Screenshot 2022-12-10 at 22 29 57

Screenshot 2022-12-10 at 22 30 46

Screenshot 2022-12-10 at 22 32 25


app/forms/teachers/otp_form.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+class Teachers::OtpForm
+  MAX_GUESSES = 5
+  EXPIRY_IN_MINUTES = 30.minutes
```
gpeng, line 5: The email says 15 minutes

app/lib/devise/strategies/otp_authenticatable.rb
```diff
@@ -0,0 +1,53 @@
...
+    super && http_auth_hash[:otp].present?
+  end
+
+  def valid_for_params_auth?
+    super && params_auth_hash[:otp].present?
+  end
+
+  def authenticate!
+    resource = mapping.to.find_by(email:)
+    unless resource && resource.secret_key.present?
```
gpeng, line 22: Could we extract these predicates here and on 27 so we get:

```
fail!(:otp_invalid) and return if secret_key_absent? || otp_invalid?
```

app/views/teachers/mailer/otp.text.erb
```diff
@@ -0,0 +1,11 @@
+Use this code to confirm your email address:
+
+<%= @otp %>
+
+The code will expire after 15 minutes.
```
gpeng, line 5: 15 minutes or 30?

gpeng (changes requested):
Looks good and works really well. So much better than those horrific links.

I spotted a couple of bits of content that need updating.

Screenshot 2022-12-09 at 13 44 21

Screenshot 2022-12-09 at 13 41 37

There's a validation message missing too. Not sure we even want to validate fort that here 🤷 

Screenshot 2022-12-09 at 13 42 43

I also got a 500 page when I tried to enter a code that I'd already used. I didn't test the expiry as yet but reuse definitely errors.

gpeng:
Just spotted you've addressed some of that content stuff in the next PR. I'll check again on that one when this is merged.

thomasleese:
Thanks @gpeng I've addressed those comments.

> I also got a 500 page when I tried to enter a code that I'd already used. I didn't test the expiry as yet but reuse definitely errors.

Do you know if you took any other steps? I've got a test for this scenario and can't reproduce the problem myself: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/5ce576166a198a124d23564cac447e0ce1026545/spec/system/teacher_interface/authentication_spec.rb#L94-L107

> Do we need a 'resend the code' link on the 'enter your code' page.

It might be a good idea, would be interested to see what the UCD team think. There is a back link that takes them to the sign in page again, and the failure page has a link allowing them to try again.

> Also I wonder if we should include the link back to that page in the email although I realise that will get us back into horrible link land.

Yeah, I'm not sure about this one, I'm happy to add it, but I suppose then users might think it's a link they have to click to sign in.

gpeng (approved):

gpeng:
> Do you know if you took any other steps? I've got a test for this scenario and can't reproduce the problem myself:

I can't recreate it now either 🤷

PR #817, opened 2022-12-07
[1209] Split notes from decline reasons part 3

gpeng:
We've switched to using `AssessmentSection#assessment_section_failure_reasons` instead of the string array `selected_failure_reasons`.

Currently we write to both. This PR switches everything over to use the `AssessmentSection#assessment_section_failure_reasons` so as we can remove drop `selected_failure_reasons`.

To do this I dropped the column locally and then made these changes until everything was passing again. I've put the column back in for now and will drop it in a subsequent PR.  Following that we'll likely rename  `assessment_section_failure_reasons` too.


app/view_objects/teacher_interface/application_form_show_view_object.rb
```diff
@@ -30,14 +30,19 @@ def notes_from_assessors
...
+      next nil if section.assessment_section_failure_reasons.empty?
 
       failure_reasons =
-        section.selected_failure_reasons.map do |key, assessor_feedback|
-          is_decline = FailureReasons.decline?(failure_reason: key)
-          assessor_feedback = "" unless is_decline
+        section.assessment_section_failure_reasons.map do |failure_reason|
+          is_decline =
+            FailureReasons.decline?(failure_reason: failure_reason.key)
+          # assessor_feedback = "" unless is_decline
```
thomasleese, line 39: I think we can remove this line.

thomasleese (approved):
Just one minor comment, but it's good to go!

PR #832, opened 2022-12-13
[1269] Add santise.sql

gpeng:
* Add script to scrub PII from a database instance

This will be used to create a sanitised backup that we can use to create production like environments. These mostly match the analytics_pii.yml definitions.

I've run this locally and it all seems to be working. I'll run on the review app and add a comment when that's deployed.

### DIFF
--- a/db/scripts/sanitise.sql
+++ b/db/scripts/sanitise.sql
@@ -78,8 +78,7 @@ SET
 DELETE FROM "sessions";
 
 -- Staff
--- Keep logins?
--- no update required 
+DELETE FROM staff;
 
 -- Teacher
 UPDATE "teachers"
--- a/db/scripts/sanitise.sql
+++ b/db/scripts/sanitise.sql
@@ -1,0 +1,108 @@
+-- ActiveStorage
+
+DELETE FROM "active_storage_attachments";
+DELETE FROM "active_storage_blobs";
+DELETE FROM "active_storage_variant_records";
+
+-- ApplicationForm
+
+UPDATE "application_forms"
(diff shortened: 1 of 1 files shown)


db/scripts/sanitise.sql
```diff
@@ -0,0 +1,101 @@
...
+-- no update required
+
+-- ReminderEmail
+-- no update required
+
+-- Session
+DELETE FROM "sessions";
+
+-- Staff
+-- Keep logins?
```
gpeng, line 76: This will mean that we can all login on preprod and the assessor personas will still work if we run this over and example data setup. Wondering if we'd consider this PI and if we're happy taking this out of the production space and into preprod? It's really only staff emails that would be 'exposed' but there are hashed passwords in there too. I'm sort of leaning towards wiping them. Thoughts welcomed...

gpeng: Actually given we'll likely run this in other environments too I think we should definitely wipe them. I'll update.

db/scripts/sanitise.sql
```diff
@@ -0,0 +1,107 @@
...
+-- Assessment 
+-- no update required
+
+-- Countries
+-- no update required
+
+-- Documents
+-- no update required
+
+-- DQTTrnRequest
```
thomasleese, line 36:
```suggestion
-- DQTTRNRequest
```

😆

gpeng: I'm so angry right now...

thomasleese (approved):

PR #849, opened 2022-12-16
Add work experience question to eligibility checker

thomasleese:
This adds a new question to the eligibility checker, depending on whether a feature flag is active, which asks the teacher about their work experience.

I haven't changed anything about how we build the performance dashboard because we won't know the exact date/time of when the feature flag will be turned on, and therefore we can't update it yet to query based on that.

Depends on #845 

[Trello Card](https://trello.com/c/wpdkO3eO/1245-build-new-work-experience-question-in-checker)

## Screenshots

Screenshot 2022-12-16 at 12 27 51
Screenshot 2022-12-16 at 12 27 58
Screenshot 2022-12-16 at 12 51 29
Screenshot 2022-12-16 at 13 15 17
Screenshot 2022-12-16 at 13 15 31


app/models/eligibility_check.rb
```diff
@@ -22,6 +23,10 @@ class EligibilityCheck < ApplicationRecord
   belongs_to :region, optional: true
   has_one :application
 
+  enum :work_experience,
+       %i[under_9_months 9_to_20_months over_20_months],
```
gpeng, line 27: Should we store this as a string as we do elsewhere?

thomasleese: I've amended the commit so they're stored as strings now, at the time I was thinking there was no major benefit to storing them as strings and it makes the database bigger, but I've just thought we probably want them as strings in analytics so makes sense, and it matches our other enums.

gpeng (approved):

PR #852, opened 2022-12-19
Remove migration for feature flags

thomasleese:
It's not possible to run this migration anymore as the model it references no longer exists. It's already run in all the environments so we can safely delete it.

Should resolve https://sentry.io/organizations/dfe-teacher-services/issues/3822120387/.


thomasleese:
I'm going to merge this in as it'll have no effect in production and it's breaking the review apps.

PR #875, opened 2022-12-28
Update work history for new regulations

thomasleese:
This implements a new version of the work history flow which is enabled when the new regulations feature flag is active. I've built as a separate controller as there's a little overlap between this and the previous one, but using the same model to keep the database structure clearer.

[Trello Card](https://trello.com/c/JCjmoumv/1255-build-work-history-in-application-form)

## Screenshots

Screenshot 2022-12-28 at 12 32 39
Screenshot 2022-12-28 at 12 32 45
Screenshot 2022-12-28 at 12 32 51
Screenshot 2022-12-28 at 12 32 56
Screenshot 2022-12-28 at 12 33 01
Screenshot 2022-12-28 at 12 33 25
Screenshot 2022-12-28 at 12 33 29
Screenshot 2022-12-28 at 12 33 42
Screenshot 2022-12-28 at 12 33 48
Screenshot 2022-12-28 at 12 33 52

### DIFF
--- a/app/controllers/teacher_interface/new_regs/work_histories_controller.rb
+++ b/app/controllers/teacher_interface/new_regs/work_histories_controller.rb
@@ -1,0 +1,283 @@
+# frozen_string_literal: true
+
+module TeacherInterface
+  module NewRegs
+    class WorkHistoriesController < BaseController
+      include HandleApplicationFormSection
+      include HistoryTrackable
+
+      before_action :redirect_unless_application_form_is_draft
+      before_action :load_application_form
+
+      before_action :load_months_count, only: %i[add_another submit_add_another]
+
+      skip_before_action :track_history, only: :index
+
+      define_history_check :check_collection
+      define_history_check :check_member, identifier: :check_member_identifier
+
+      def index
+        if application_form.work_history_status_completed?
+          redirect_to %i[
+                        check
+                        teacher_interface
+                        application_form
+                        new_regs
+                        work_histories
+                      ]
+        elsif application_form.work_histories.empty?
+          redirect_to %i[
+                        new
+                        teacher_interface
+                        application_form
+                        new_regs
+                        work_history
+                      ]
+        else
+          redirect_to %i[
+                        add_another
+                        teacher_interface
+                        application_form
+                        new_regs
+                        work_histories
+                      ]
+        end
+      end
+
+      def check_collection
+        @work_histories = application_form.work_histories.ordered
+        @came_from_add_another =
+          history_stack.last_entry&.fetch(:path) ==
+            add_another_teacher_interface_application_form_new_regs_work_histories_path
+      end
+
+      def new
+        @work_history = WorkHistory.new(application_form:)
+
+        @form = WorkHistorySchoolForm.new(work_history: @work_history)
+      end
+
+      def create
+        @work_history = WorkHistory.new(application_form:)
+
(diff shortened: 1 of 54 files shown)


app/controllers/teacher_interface/new_regs/work_histories_controller.rb
```diff
@@ -0,0 +1,283 @@
+# frozen_string_literal: true
+
+module TeacherInterface
+  module NewRegs
```
gpeng, line 4: `NewRegulations`? Just thinking we don't generally abbreviate except for the pesky acronym ones.

app/lib/work_history_duration.rb
```diff
@@ -0,0 +1,76 @@
...
+  AVERAGE_WEEKS_PER_MONTH = 4.34
+  HOURS_PER_FULL_TIME_MONTH = 130.0
+
+  attr_reader :application_form
+
+  def work_histories
+    @work_histories ||=
+      application_form
+        .work_histories
+        .where.not(start_date: nil)
```
gpeng, line 26: Could these be a scope on WorkHistory?

app/lib/application_form_status_updater.rb
```diff
@@ -88,19 +88,29 @@ def subjects_status
...
-           !work_histories.empty? &&
-             work_histories.all? do |work_history|
-               work_history_complete?(work_history)
-             end
-         )
-      return :completed
-    end
+    all_work_histories_complete =
+      work_histories.all? do |work_history|
+        work_history_complete?(work_history)
```
gpeng, line 93: Would this make more sense on the `WorkHistory` model? I realise it was already here.

app/controllers/teacher_interface/new_regs/work_histories_controller.rb
gpeng: I see this is in a few places. Given they'll be the only regs in a couple of months I can live with the abbreviation.

thomasleese: I wasn't particular happy with the module anyway, but it seemed like the only option. I'm happy to change it, or leave it, either way.

app/lib/application_form_status_updater.rb
thomasleese: This was actually originally where we had it before implementing the `ApplicationFormStatusUpdater`, I think I prefer it in this class as we want to make sure we're only using the cached `work_history_status` field and not interrogating individual work histories, but I'm happy either way.

app/controllers/teacher_interface/new_regs/work_histories_controller.rb
gpeng: Let's leave it. I'll do a card for removing the old regs 'in the future ™️'  and then this will just be regs 🤣

app/lib/work_history_duration.rb
thomasleese: I'm not sure, it feels quite specific to this class to me, in the sense that the only values we _need_ to calculate the duration are the `start_date` and the `hours_per_week`. I could add a `has_start_date` and `has_hours_per_week` scope, but that doesn't feel worth it to me, and having a `can_calculate_duration` scope feels like leaking responsibilities of the duration class to the model. What do you think?

gpeng (approved):

PR #888, opened 2023-01-03
Ensure resolved duplicates can be awarded

thomasleese:
There's an issue currently where application forms which have been flagged as potential duplicates cannot be awarded as they're not in the right state for the `AwardQTS` service is called.


gpeng (commented):
Can a resolved duplicate ever come back as anything other than awarded from the DQT API? I'm just wondering if it would be better to allow `AwardQTS` to award applications in `potential_duplicate_in_dqt` state. It seems more representative of what happens and would avoid the complexity in the job.

thomasleese:
> Can a resolved duplicate ever come back as anything other than awarded from the DQT API?

I'm not entirely sure. The reason I went with this approach was to guard against the potential of the duplicate being resolved, but then no TRN comes back, so at least the application form will be in the right state. As you say though, this might never happen in real life, in which case I prefer the approach you've suggested.

gpeng:
Fair point. I wonder if we should guard against awarding an application without a TRN in the `AwardQTS` service rather than based on the state, or maybe in addition to the state. That would give us an exception to work from then if we do find the API is resolving duplicates without actually awarding them.

thomasleese:
> Fair point. I wonder if we should guard against awarding an application without a TRN in the `AwardQTS` service rather than based on the state, or maybe in addition to the state. That would give us an exception to work from then if we do find the API is resolving duplicates without actually awarding them.

I've amended the commit and reworked it to be as you suggest, this should make it clear if there's ever a case where a potential duplicate hasn't been found, but the API hasn't responded with a TRN.

gpeng (approved):

PR #900, opened 2023-01-05
Add 'eligibility_enabled' to countries model

David-Feetenby:
Add a new field to the countries table
This field will be used for turning off and on the eligible checker for whole countries.

There is no interface for this in this pr as this pr is 1 part of : https://trello.com/c/rYd9rkdi/1314-add-nine-new-countries-to-the-support-console

### DIFF
--- a/db/migrate/20230105155624_add_eligibility_enabled_to_countries.rb
+++ b/db/migrate/20230105155624_add_eligibility_enabled_to_countries.rb
@@ -1,0 +1,9 @@
+class AddEligibilityEnabledToCountries < ActiveRecord::Migration[7.0]
+  def change
+    add_column :countries,
+               :eligibility_enabled,
+               :boolean,
+               default: true,
+               null: false
+  end
+end
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -4,6 +4,7 @@
 #
 #  id                                      :bigint           not null, primary key
 #  code                                    :string           not null
+#  eligibility_enabled                     :boolean          default(TRUE), not null
 #  teaching_authority_address              :text             default(""), not null
 #  teaching_authority_certificate          :text             default(""), not null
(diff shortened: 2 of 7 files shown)


lib/tasks/add_countries.rake
```diff
@@ -0,0 +1,37 @@
...
+      if regions.blank?
+        country
+          .regions
+          .find_or_create_by!(name: "")
+          .update!(legacy: false, application_form_enabled: false)
+      else
+        regions.each do |region|
+          next if country.regions.where(name: region[:name]).any?
+
+          country.regions.create!(region)
```
thomasleese, line 18:
```suggestion
          country.regions.create!(legacy: false, application_form_enabled: false, **region)
```

Perhaps we want to default to this as well like the "national" region in the lines above?

lib/tasks/add_countries.rake
```diff
@@ -0,0 +1,37 @@
+namespace :add_countries do
+  desc "Add countries to support interface"
+  task generate: :environment do
```
thomasleese, line 3: I suggest we name the Rake task specific to this one time, since we're going to run it once and then never again. Something like `countries:create_feb_1` perhaps?

db/migrate/20230105155624_add_eligibility_enabled_to_countries.rb
```diff
@@ -0,0 +1,5 @@
+class AddEligibilityEnabledToCountries < ActiveRecord::Migration[7.0]
+  def change
+    add_column :countries, :eligibility_enabled, :boolean
```
thomasleese, line 3: We should also add this field to the support console so someone can tick the box to enable the countries on the 1st Feb.

David-Feetenby:
[Add nine new countries to the support console](https://trello.com/c/rYd9rkdi/1314-add-nine-new-countries-to-the-support-console)

PR #903, opened 2023-01-05
View FI details after award or decline

steventux:
## Trello 

https://trello.com/c/RWlWJnk5/1310-view-fi-docs-details-after-award-decline

## Context and changes made

Assessors should be able to see FI details if the application has been awared or declined.
We use the `passed` value on the `FurtherInformationRequest` record to determine whether the FI details are read only or editable.

Not sure whether we intend to render the check-answers component information in the read-only view (see screenshot)
 

## Edit page in read only state

image

### DIFF
--- a/app/view_objects/assessor_interface/application_forms_show_view_object.rb
+++ b/app/view_objects/assessor_interface/application_forms_show_view_object.rb
@@ -51,10 +51,7 @@ class AssessorInterface::ApplicationFormsShowViewObject
       further_information_request = further_information_requests[index]
 
       if further_information_request.received? &&
-           (
-             further_information_request.passed.nil? ||
-               assessment.request_further_information?
-           )
+           assessment.request_further_information?
         url_helpers.edit_assessor_interface_application_form_assessment_further_information_request_path(
           application_form,
           assessment,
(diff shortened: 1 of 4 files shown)


thomasleese:
I've had a look at the PR and it definitely seems to be on the right track. I wonder about whether we can find a way of re-using the `edit` controller method here rather than needing a new one, since effectively we just want to show the assessors a read-only version of the `edit` page (i.e. the form at the bottom isn't there).

I think an advantage of this is we don't need to introduce a new view which will need to be mostly kept in sync with the view used in the `edit` page, but it is a little odd to have `/edit` in the URL when it's read-only.

> How we render the link to the edit/review FI pages doesn't directly relate to the application form state, the conditionals were and still are related to the state of the assessment and further information request.

I think this is okay, since that's the most relevant state for determining whether there's a link.

> Do we need greater protection around access for the (read-only) review page added in this PR? I've tried to stay away from Pundit policies as mentioned in the Trello story but it feels like it's missing some checks similar to those which control how the link is rendered.

Since the review page has no form on it, I don't think we do need anything more, we've already decided that any user will have the ability to _view_ information about an application.

steventux:
Thanks @thomasleese will amend

steventux:
@thomasleese have removed the `review` route, action etc. This makes the PR quite simple - the only question I have is around the check-answers component rendering - do we want this (see updated screenshot on PR).

thomasleese (approved):
I think this looks good, and I think we do want the "Check your answers" component to be displayed since it should include the response from the applicant.

PR #922, opened 2023-01-11
Add expiration logic to Requestable

thomasleese:
This moves the expiration logic out of the `FurtherInformationRequestExpirable` concern and in to the `Requestable` concern since we'll need this logic in both further information requests and reference requests.


app/models/further_information_request.rb
```diff
@@ -28,7 +28,17 @@ class FurtherInformationRequest < ApplicationRecord
...
   has_many :reminder_emails
 
+  delegate :application_form, to: :assessment
+
   def failed
     passed == false
   end
+
+  def expires_after
+    if %w[AU CA GI NZ US].include?(application_form.country.code)
```
steventux, line 38: This array could be a constant, not a blocker.

app/models/concerns/requestable.rb
```diff
@@ -16,4 +16,12 @@ module Requestable
       update!(state: "received", received_at: Time.zone.now)
     end
   end
+
+  def expires_after
+    6.weeks
```
steventux, line 21: Is 6 weeks the default expiry period? (ie. if not one of the countries which expires after 4 weeks in `FurtherInformationRequest`)
Perhaps it could be shared via a constant or method if so. 
Not a blocker.

steventux (approved):
:+1: 2 minor points, not blocking. Good as is.

app/models/concerns/requestable.rb
thomasleese: I'm actually thinking I might take this method out of `Requestable` and define it explicitly on `FurtherInformationRequest` and `ReferenceRequest`, since it's probably not the responsibility of the concern to have a default.

steventux (approved):
:rocket:

PR #923, opened 2023-01-11
Add emails related to references

thomasleese:
This adds two new emails which we send in relation to references, one to the referee and one to the teacher.

I've also refactored the mailers to make better use of helpers for common code.

Depends on #921 and #922

[Trello Card](https://trello.com/c/ipsDiRoR/1340-send-automatic-emails-to-referees)

### DIFF
--- a/app/helpers/application_form_helper.rb
+++ b/app/helpers/application_form_helper.rb
@@ -1,6 +1,13 @@
+# frozen_string_literal: true
+
 module ApplicationFormHelper
   def application_form_full_name(application_form)
-    "#{application_form.given_names} #{application_form.family_name}"
+    if application_form.given_names.present? ||
+         application_form.family_name.present?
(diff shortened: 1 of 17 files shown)


app/helpers/application_form_helper.rb
```diff
@@ -1,6 +1,13 @@
+# frozen_string_literal: true
+
 module ApplicationFormHelper
   def application_form_full_name(application_form)
-    "#{application_form.given_names} #{application_form.family_name}"
+    if application_form.given_names.present? ||
+         application_form.family_name.present?
+      "#{application_form.given_names} #{application_form.family_name}"
```
steventux, line 7: Minor thing: Does this need to be stripped of leading and trailing whitespace? It's possible one of the two names is not present.

steventux (approved):
One minor comment, otherwise looks good to me. :rocket:

PR #925, opened 2023-01-11
Allow countries to skip eligibility questions

thomasleese:
This adds a new boolean field to countries allowing some countries to skip the eligibility questions. At the moment we only know of Scotland and Northern Ireland where this will be the case, but I've implemented this as a field on the country model following the same pattern for other customisations like this.

## Screenshots

Screenshot 2023-01-11 at 12 12 00
Screenshot 2023-01-11 at 12 12 12
Screenshot 2023-01-11 at 12 12 16


steventux (approved):
:+1:

PR #927, opened 2023-01-11
Allow regions to skip work history

thomasleese:
This adds a new boolean field to regions allowing some regions to skip the work history spoke of the application form. At the moment we only know of Scotland and Northern Ireland where this will be the case, but I've implemented this as a field on the country model following the same pattern for other customisations like this.

## Screenshots

Screenshot 2023-01-11 at 13 27 48
Screenshot 2023-01-11 at 13 28 00
Screenshot 2023-01-11 at 13 37 48


steventux (approved):
:+1:

PR #931, opened 2023-01-11
Provide 'qualifications information' on a per region or country basis

steventux:
## Trello

https://trello.com/c/Jyzrctlo/1332-what-qualifications-we-cannot-accept-by-country

## Context / Changes

We'd like to provide tailored content on a per region or per country basis which is appended to the **Proof of qualifications** section visible in the eligibility screener. The content should be editable via the support UI for each country or region where applicable.

- Add `qualifications_information` to `countries` and `regions` tables
- Add form fields in Support UI to capture qualifications information
- Present qualifications information on support preview screens and relevant eligibility screener page

#### Support UI

image

#### Rendered in content
 
image

#### Rendered in application form (qualifications section)

image

### DIFF
--- a/app/controllers/support_interface/countries_controller.rb
+++ b/app/controllers/support_interface/countries_controller.rb
@@ -67,6 +67,7 @@ class SupportInterface::CountriesController < SupportInterface::BaseController
     params.require(:country).permit(
       :eligibility_enabled,
       :eligibility_skip_questions,
+      :qualifications_information,
       :teaching_authority_name,
       :teaching_authority_address,
       :teaching_authority_emails_string,
--- a/app/controllers/support_interface/regions_controller.rb
+++ b/app/controllers/support_interface/regions_controller.rb
@@ -41,6 +41,7 @@ class SupportInterface::RegionsController < SupportInterface::BaseController
       :legacy,
       :application_form_enabled,
       :application_form_skip_work_history,
+      :qualifications_information,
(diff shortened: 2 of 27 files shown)


thomasleese (commented):
Overall looks good, I think we also need to show this information in the qualification form of the application form: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/app/views/teacher_interface/qualifications/_form.html.erb

app/forms/teacher_interface/qualification_form.rb
```diff
@@ -55,6 +55,15 @@ def update_model
...
 
+    def qualifications_information
+      region = qualification.application_form.region
+
+      (
+        region.qualifications_information.presence ||
+          region.country.qualifications_information
+      )
+    end
+
```
thomasleese, line 66: I'm not sure about this going in the form, generally we've tried to keep the forms only responsible for saving to the database and validation. We have started introducing "view objects" which might be a better place to put this, although for something as small as this I'd be tempted to put it directly in to the view.

steventux: Thanks @thomasleese I introduced a view object as this was easier to unit test variations. Wasn't 100% sure it should be `QualificationViewObject` but that seems to fit with the controller & action.

thomasleese:
I'm going to merge this in as it's been product reviewed.

PR #932, opened 2023-01-11
Always show qualification question

thomasleese:
This changes the logic behind the "skips question" field on countries to show the qualification question, but none other than that. This was missed in 0466c7e13fbe04216cfabc317cca184e129e86db where this field was added.

I did consider renaming the field, but I thought it was okay to keep it for now and I've changed the label in the support console (the field didn't say which questions it skips in the first place). If we do want to rename it, we'll need to do it in steps anyway so it'll come in a separate PR.


steventux (commented):
:rocket:

steventux (approved):

PR #935, opened 2023-01-12
Add induction required form where necessary

thomasleese:
This adds a new "induction required" form to the check professional standing assessor spoke if the application form is under the new regulations and didn't require work history (and therefore has no way of determining the induction required status other than asking the assessor).

This is only needed for Scotland and Northern Ireland at the moment but I've implemented it in a generic way that means we could apply it to other countries in the future if we needed to.

[Trello Card](https://trello.com/c/ZDTTxczS/1344-scotland-ni-assessor)

### Scotland

Screenshot 2023-01-12 at 10 58 49

### Northern Ireland

Screenshot 2023-01-12 at 11 09 50


steventux (approved):
:rocket: :100:

PR #938, opened 2023-01-13
Add CF CLI ACTION

enformatique11:


thomasleese:
I think #990 fixes this but feel free to reopen if I've got that wrong.

PR #946, opened 2023-01-15
Allow work history without contact information 

thomasleese:
If the user is going through the reduced evidence journey then they can enter their work history without needing to provide the contact information.

[Trello Card](https://trello.com/c/HnLsFSKb/1379-reduce-evidence-journey-application-form-work-history)


steventux (approved):
:rocket:

PR #956, opened 2023-01-17
Send submission emails to teaching authorities

thomasleese:
This adds a new email which is sent to teaching authorities when an application is submitted and the teaching authority requires an email.

[Trello Card](https://trello.com/c/DNbG6bWE/1363-notify-a-teaching-authority-that-we-will-need-a-lops)

## Screenshots

Screenshot 2023-01-18 at 07 30 47
Screenshot 2023-01-18 at 07 31 23

### DIFF
--- a/config/initializers/date_formats.rb
+++ b/config/initializers/date_formats.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+Date::DATE_FORMATS[:long_ordinal_uk] = "%-d %B %Y"
--- a/app/controllers/support_interface/regions_controller.rb
+++ b/app/controllers/support_interface/regions_controller.rb
@@ -55,6 +55,7 @@ class SupportInterface::RegionsController < SupportInterface::BaseController
       :teaching_authority_other,
       :teaching_authority_online_checker_url,
       :teaching_authority_provides_written_statement,
+      :teaching_authority_requires_submission_email,
     )
   end
 end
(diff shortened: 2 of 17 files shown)


app/helpers/dates_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module DatesHelper
+  def format_date(date)
```
steventux, line 4: Is there a reason why we're not adding a custom format to `Date::DATE_FORMATS` and calling `Date#to_fs(:my_custom_format)` ? 
This works fine as is but it's not how most of our date formats are defined in [other projects](https://github.com/DFE-Digital/find-a-lost-trn/blob/main/config/initializers/date_formats.rb).

steventux (approved):
A couple of trivial comments. Fine to merge as is.

app/helpers/dates_helper.rb
thomasleese: No particular reason, I didn't know that existed! I'll amend it to use that.

config/initializers/date_formats.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+Date::DATE_FORMATS[:long_ordinal_uk] = "%-d %B %Y"
```
steventux, line 3: :+1: :100:

steventux (approved):

PR #973, opened 2023-01-19
English language proficiency assessor flow

steventux:
## Trello

https://trello.com/c/hW8QvsYR/1383-build-english-language-proficiency-assessor-flow

## Context & changes

Assessor checks for the _English language proficiency (ELP)_ section differ from the other section checks if the teacher has indicated they have exemption via citizenship or qualification. 
In this case the assessor confirms ELP via the personal information section...

#### Personal information section when exemption via citizenship

image

#### Qualifications section if the exemption is via the qualification

image

#### ELP section for both cases directing the assessor to the relevant section eg.

image

#### Confirming ELP section when proof is MOI

image

#### Confirming ELP section when proof is provider

image


app/forms/assessor_interface/updates_english_language_status.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module AssessorInterface::UpdatesEnglishLanguageStatus
```
thomasleese, line 3: Could we put this inside a `concerns` directory?

thomasleese (commented):
I've had a quick look and it all looks pretty good, I've left a small comment, but nothing major.

app/lib/assessment_factory.rb
```diff
@@ -121,6 +122,40 @@ def age_range_subjects_section
     AssessmentSection.new(key: "age_range_subjects", checks:, failure_reasons:)
   end
 
+  def english_language_proficiency_section
```
thomasleese, line 125: We'll need to add a conditional to make sure this isn't shown for all applications:

```rb
if application_form.created_under_new_regulations? &&
         FeatureFlags::FeatureFlag.active?(:application_english_language)
```

We only want it to show for new applications (as old ones aren't asked about their English language proficiency) and if the feature is enabled.

thomasleese (commented):
All looks good, I've left one comment I missed before.

thomasleese (approved):

PR #975, opened 2023-01-19
Remove further information statuses (part 2 of 2)

thomasleese:
These statuses will be covered by the new waiting on and received statuses, and so we won't need dedicated statuses for further information.

See also #974 

## Screenshot

Screenshot 2023-01-19 at 13 28 40


steventux (approved):
:100:

PR #976, opened 2023-01-19
Record potential duplicates in the DQT TRN requests

thomasleese:
This adds a new column to the DQT TRN Request which will store whether or not the request returned with a potential duplicate in DQT. We'll be storing this so the overall status of an application can be determined dynamically from this information.


steventux (approved):
:rocket:

PR #987, opened 2023-01-22
Determine application form status dynamically

thomasleese:
This changes the status of the application form to be dynamic, generated based on the status of other fields, which means we can apply business logic according to certain rules. In a follow up PR I'm going to rename this field from `state` to `status` so it matches the "section status" fields we also determine dynamically.

Depends on #976, #974 and #975.


David-Feetenby (approved):

steventux (approved):
:+1: Looks good, easy to review by commits - all makes sense.

PR #995, opened 2023-01-23
Update sanitise.sql

gpeng:
We create a sanitised back up with db/scripts/santise.sql. It was erroring on one of the statements and setting the feature flags fully to production and consequently opening the service on restored environments.

* Fix the teacher update SQL. There is no `unconfirmed_email_address` field on that table so that bit was failing
* Add a comment about ReferenceRequest. No sanitisation needed on that table
* Set the `service_open` flag to `false`. Not strictly 'sanitisation' but it avoids us unintentionally opening a restored environment

### DIFF
--- a/db/scripts/sanitise.sql
+++ b/db/scripts/sanitise.sql
@@ -41,7 +41,10 @@ DELETE FROM "dqt_trn_requests";
 -- no update required
 
 -- Features
--- no update required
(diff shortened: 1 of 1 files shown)


steventux (approved):
Do we need `like` or `=` for the `service_open` flag? Either way works, not a blocker.

thomasleese (approved):

PR #996, opened 2023-01-23
Alias application form state to status (part 1 of 2)

thomasleese:
This allows us to rename the column in the next commit in a backwards compatible way. I'd like to rename it to be clearer that "status" is something dynamically generated based on the "state" of other objects.


steventux (approved):
:+1:

PR #997, opened 2023-01-23
Rename application form state to status (part 2 of 2)

thomasleese:
This renames the state field to status to make it clearer that it's generated dynamically. I'd like to rename it to be clearer that "status" is something dynamically generated based on the "state" of other objects.

Depends on #996 


steventux (approved):
:100:

PR #1000, opened 2023-01-24
Add maintenance page

NickGraham101:
### Context

We need to be able to put up a maintenance page containing static content.

### Changes proposed in this PR

- Static content for maintenance page 
- Make recipes that reroute traffic inside GovUK PaaS.  Rerouting traffic via Frontdoor caused down time.

### Guidance for review

- The maintenance page can be put up and taken down in dev with the following shell commands: `make dev enable-maintenance` and `make dev disable-maintenance`
- The recipes can't be tested in prod, review the variables used to ensure the correct routes are changed


gpeng (approved):
This looks good to me and seems to be working fine on dev. We need to update the content and add some docs but we can do that in a separate ticket when the content is ready 👍

NickGraham101:
[Deployable 'maintenance' state](https://trello.com/c/kuTEZAe9/1430-deployable-maintenance-state)

saliceti (approved):

PR #1040, opened 2023-01-30
Ensure we associate example applications with providers from seed data

steventux:
This will make sure we only use the providers from seed data when generating example data instead of creating dummy providers.

### DIFF
--- a/spec/factories/application_forms.rb
+++ b/spec/factories/application_forms.rb
@@ -256,7 +256,9 @@ FactoryBot.define do
 
     trait :with_english_language_provider do
       english_language_proof_method { "provider" }
-      association :english_language_provider
+      english_language_provider do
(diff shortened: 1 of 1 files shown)


spec/factories/application_forms.rb
```diff
@@ -260,6 +271,12 @@
       english_language_provider_reference { "reference" }
     end
 
+    trait :with_english_language_provider_from_seed_data do
+      english_language_proof_method { "provider" }
+      english_language_provider { providers_from_seed_data.sample }
```
thomasleese, line 276:
```suggestion
      english_language_provider { EnglishLanguageProvider.all.sample }
```

Your code works, but I wonder if it's enough that we can do this. It's unlikely that we'd ever have any more providers since there's no means to create them other than in the Rails console or running the seeds.

thomasleese (approved):

PR #1077, opened 2023-02-03
Add assessor comments to decline QTS

David-Feetenby:
Add Assessor's notes before final decline step

Add a free text field for assessor's notes on the final decline step so that declined applicants can see more info

This is validated when there are no failure reasons, so that in the event of an empty decline (I.E. Rapid decline/FI requests) it is mandatory. Otherwise, it should be optional.

### DIFF
--- a/app/controllers/assessor_interface/assessment_recommendation_declines_controller.rb
+++ b/app/controllers/assessor_interface/assessment_recommendation_declines_controller.rb
@@ -8,20 +8,27 @@ module AssessorInterface
     before_action :load_assessment_and_application_form
 
     def edit
-      @form = AssessmentDeclarationForm.new
+      @form = AssessmentDeclarationDeclineForm.new
     end
 
     def update
       @form =
-        AssessmentDeclarationForm.new(
+        AssessmentDeclarationDeclineForm.new(
           declaration:
             params.dig(
-              :assessor_interface_assessment_declaration_form,
+              :assessor_interface_assessment_declaration_decline_form,
               :declaration,
             ),
+          recommendation_assessor_note:
+            params.dig(
+              :assessor_interface_assessment_declaration_decline_form,
+              :recommendation_assessor_note,
+            ),
+          session:,
+          assessment:,
         )
 
-      if @form.valid?
+      if @form.save
         redirect_to [
                       :preview,
                       :assessor_interface,
@@ -59,6 +66,10 @@ module AssessorInterface
         if @form.confirmation
           ActiveRecord::Base.transaction do
             assessment.decline!
+            assessment.update!(
+              recommendation_assessor_note:
+                session[:recommendation_assessor_note],
+            )
             DeclineQTS.call(application_form:, user: current_staff)
           end
 
--- a/app/forms/assessor_interface/assessment_declaration_decline_form.rb
+++ b/app/forms/assessor_interface/assessment_declaration_decline_form.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class AssessorInterface::AssessmentDeclarationDeclineForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attr_accessor :session, :assessment
+
+  attribute :declaration, :boolean
+  attribute :recommendation_assessor_note, :string
+
+  validates :declaration, presence: true
+  validates :recommendation_assessor_note,
+            presence: true,
+            if: -> { assessment.selected_failure_reasons_empty? }
+
+  def save
+    return false unless valid?
+
+    session[:recommendation_assessor_note] = recommendation_assessor_note
+    true
+  end
+end
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -8,6 +8,7 @@
 #  age_range_note                            :text             default(""), not null
 #  induction_required                        :boolean
 #  recommendation                            :string           default("unknown"), not null
+#  recommendation_assessor_note              :text             default(""), not null
 #  recommended_at                            :datetime
 #  started_at                                :datetime
 #  subjects                                  :text             default([]), not null, is an Array
@@ -120,6 +121,10 @@ class Assessment < ApplicationRecord
       end
       recommendations << "decline" if can_decline?
     end
+  end
+
+  def selected_failure_reasons_empty?
+    sections.all? { |section| section.selected_failure_reasons.empty? }
   end
 
   private
--- a/app/views/assessor_interface/assessment_recommendation_declines/edit.html.erb
+++ b/app/views/assessor_interface/assessment_recommendation_declines/edit.html.erb
@@ -6,38 +6,38 @@
 <p class="govuk-body-l">You’re about to decline this QTS application from <%= application_form_full_name(@application_form) %> for the following reasons:</p>
 
 <h2 class="govuk-heading-l">Your decline reasons and notes</h2>
+<% unless @assessment.further_information_requests.exists? %>
+  <% @assessment.sections.each do |section| %>
+    <% if section.selected_failure_reasons.present? %>
+      <h3 class="govuk-heading-m"><%= t(".assessment_section.#{section.key}") %></h3>
+      <ul class="govuk-list">
+        <% section.selected_failure_reasons.each do |failure_reason| %>
+          <li>
+            <h4 class="govuk-heading-s"><%= t("assessor_interface.assessment_sections.show.failure_reasons.#{failure_reason.key}") %></h4>
 
-<% @assessment.sections.each do |section| %>
-  <% if section.selected_failure_reasons.present? %>
-    <h3 class="govuk-heading-m"><%= t(".assessment_section.#{section.key}") %></h3>
-    <ul class="govuk-list">
-      <% section.selected_failure_reasons.each do |failure_reason| %>
-        <li>
-          <h4 class="govuk-heading-s"><%= t("assessor_interface.assessment_sections.show.failure_reasons.#{failure_reason.key}") %></h4>
-
-          <% if FailureReasons.decline?(failure_reason: failure_reason.key) %>
-            <p class="govuk-body govuk-!-margin-bottom-2">Your note to the applicant:</p>
-          <% else %>
-            <p class="govuk-body govuk-!-margin-bottom-2">Your note (the applicant won’t see this):</p>
-          <% end %>
-
-          <%= govuk_inset_text do %>
-            <%= simple_format failure_reason.assessor_feedback %>
-          <% end %>
-        </li>
-      <% end %>
-    </ul>
+            <% if FailureReasons.decline?(failure_reason: failure_reason.key) %>
+              <p class="govuk-body govuk-!-margin-bottom-2">Your note to the applicant:</p>
+            <% else %>
+              <p class="govuk-body govuk-!-margin-bottom-2">Your note (the applicant won’t see this):</p>
+            <% end %>
+            <%= govuk_inset_text do %>
+              <%= simple_format failure_reason.assessor_feedback %>
+            <% end %>
+          </li>
+        <% end %>
+      </ul>
+    <% end %>
   <% end %>
+<% else %>
+  <p class="govuk-hint">Use this space to tell the applicant why they’ve been declined.</p>
+  <p class="govuk-hint">This is the last communication they’ll receive about this application so make sure your comments are clear.</p>
 <% end %>
 
 <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
 
-<p class="govuk-body"><%= t(".instruction") %></p>
-<p class="govuk-body">If you need to check something, use the ‘Back’ button.</p>
-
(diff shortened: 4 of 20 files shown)


app/controllers/assessor_interface/assessment_recommendation_declines_controller.rb
```diff
@@ -59,6 +66,11 @@ def update_confirm
         if @form.confirmation
           ActiveRecord::Base.transaction do
             assessment.decline!
+            assessment.update!(
+              recommendation: "decline",
```
thomasleese, line 70: This is unnecessary as the line above will have already done this.

app/forms/assessor_interface/assessment_declaration_decline_form.rb
```diff
@@ -0,0 +1,28 @@
...
+  include ActiveModel::Attributes
+
+  attr_accessor :session, :assessment
+  attribute :declaration, :boolean
+  attribute :recommendation_assessor_note, :string
+
+  validates :declaration, presence: true
+  validates :recommendation_assessor_note,
+            presence: true,
+            if: -> { selected_failure_reasons_empty(assessment) }
```
thomasleese, line 14: I wonder if this should be a method on the `assessment` model so it can be tested on its own. Perhaps `assessment.selected_failure_reasons_empty?`.

app/views/assessor_interface/assessment_recommendation_declines/edit.html.erb
```diff
@@ -37,7 +36,7 @@
 
 <%= form_with model: @form, url: [:assessor_interface, @application_form, @assessment, :assessment_recommendation_decline], method: :put do |f| %>
   <%= f.govuk_error_summary %>
-
+  <%= f.govuk_text_area :recommendation_assessor_note, label: { text: "Assessor's notes" } %>
```
thomasleese, line 39: Ideally we should put the label in `config/locales/helpers.yml` so all the strings are in the locale files.

app/views/teacher_interface/application_forms/show.html.erb
```diff
@@ -88,7 +88,12 @@
       </ul>
     <% end %>
   <% end %>
-
+   <% if @view_object.assessment.recommendation_assessor_note.present?%>
```
thomasleese, line 91:
```suggestion
  <% if @view_object.assessment.recommendation_assessor_note.present? %>
```

app/views/teacher_interface/application_forms/show.html.erb
```diff
@@ -88,7 +88,12 @@
...
+    <p class="govuk-body">
+     <%="#{@view_object.assessment.recommendation_assessor_note}"%>
```
thomasleese, line 93:
```suggestion
      <%= @view_object.assessment.recommendation_assessor_note %>
```

You don't need to interpolate it in to a string if it's the only thing being output.

db/migrate/20230203133927_add_recommendation_note_to_assessment.rb
```diff
@@ -0,0 +1,5 @@
+class AddRecommendationNoteToAssessment < ActiveRecord::Migration[7.0]
+  def change
+    add_column :assessments, :recommendation_assessor_note, :text
```
thomasleese, line 3:
```suggestion
    add_column :assessments, :recommendation_assessor_note, :text, null: false, default: ""
```

At the moment the field can either be `""` or it can be `nil` and they can both represent the absence of the note. I like to require that strings are not nullable but default to a blank string so there's only one possible value which represents the absence of a note.

spec/factories/assessments.rb
```diff
@@ -55,6 +56,11 @@
       end
     end
 
+    trait :decline_with_no_failure_Reasons do
```
thomasleese, line 59:
```suggestion
    trait :decline do
```

I think this can just be `decline` on its own and if a user wants additional failure reasons they can add them. I'd also suggest putting this next to the `award` trait so they're easier to find together.

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+require "rails_helper"
+RSpec.describe AssessorInterface::AssessmentDeclarationDeclineForm,
+               type: :model do
+  let(:assessment) { create(:assessment, :started) }
+  let(:session) { {} }
+  subject(:form) { described_class.new(session:, assessment:) }
+  it { is_expected.to validate_presence_of(:declaration) }
+
+  let(:assessment) { create(:assessment, :decline_with_no_failure_Reasons) }
```
thomasleese, line 11: This `assessment` overwrites the `assessment` definition on line 6 meaning both exceptions are run with it. You'll probably want to define two `context` blocks for a declined assessment and a started one. Then you can also test that the form _doesn't_ validate the presence of `recommendation_assessor_note` when the assessment has failure reasons.

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  it { is_expected.to validate_presence_of(:recommendation_assessor_note) }
+
+  describe "#save" do
+    context "when valid" do
+      it "saves the form" do
+        form.declaration = true
+        form.recommendation_assessor_note = "This is a note"
+        expect(form.save).to be true
```
thomasleese, line 19: If you define the subject as:

```rb
subject(:save) { form.save }
```

You can change this line to

```suggestion
        it { is_expected.to be true }
```

Which would be more typical Spec.

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  it { is_expected.to validate_presence_of(:declaration) }
+
+  let(:assessment) { create(:assessment, :decline_with_no_failure_Reasons) }
+  it { is_expected.to validate_presence_of(:recommendation_assessor_note) }
+
+  describe "#save" do
+    context "when valid" do
+      it "saves the form" do
+        form.declaration = true
+        form.recommendation_assessor_note = "This is a note"
```
thomasleese, line 18: Worth moving these outside the expectation (`it`) as they're the setup for the test, not part of the test itself.

thomasleese (commented):
General idea looks good, I've left a few comments on the code structure/style.

app/forms/assessor_interface/assessment_declaration_decline_form.rb
David-Feetenby: I've move it and think i'm testing adequately

app/forms/assessor_interface/assessment_declaration_decline_form.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class AssessorInterface::AssessmentDeclarationDeclineForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attr_accessor :session, :assessment
+  attribute :recommendation_assessor_note, :string
+  attribute :declaration, :boolean
+  attribute :recommendation_assessor_note, :string
```
thomasleese, line 10: It looks like this attribute is defined twice.

app/models/assessment.rb
```diff
@@ -122,6 +123,12 @@ def available_recommendations
     end
   end
 
+  def selected_failure_reasons_empty?(assessment)
```
thomasleese, line 126: We don’t need to pass an assessment in to this method as it’s already available to us here as `self`. You can also implicitly call methods within the same class, something like:

```
sections.all? { |section| section.selected_failure_reasons.empty? }
```

db/schema.rb
```diff
@@ -137,6 +137,7 @@
     t.integer "working_days_submission_to_started"
     t.integer "working_days_since_started"
     t.boolean "induction_required"
+    t.text "recommendation_assessor_note"
```
thomasleese, line 140: This doesn’t match the migration, it needs to be defined with the same `null` and `default` arguments.

spec/factories/assessments.rb
```diff
@@ -55,6 +56,11 @@
       end
     end
 
+    trait :decline do
```
thomasleese, line 59: I think it would be good if this was defined directly below the award trait, they are very similar and it’ll make it easier to find the traits when looking through this file.

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  context "when recommendation is started" do
+    let(:assessment) { create(:assessment, :started) }
+    it { is_expected.to validate_presence_of(:declaration) }
+  end
+
+  context "when recommendation is declined with no failure reasons" do
+    let(:assessment) { create(:assessment, :decline) }
+    it { is_expected.to validate_presence_of(:recommendation_assessor_note) }
+
+    describe "#save" do
```
thomasleese, line 25: Typically you structure a spec file with the describe blocks at the top level, as all the tests relate to the same subject (the thing we’re describing) and then within each describe block there are contexts.

If you take a look at https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/spec/forms/assessor_interface/age_range_subjects_form_spec.rb for example we’ve got two describe blocks, the validations and the save method, I think it would be good to follow that structure in this test so it’s consistent but also makes it easier to find expectations in the file.

spec/models/assessment_spec.rb
```diff
@@ -258,4 +259,23 @@
...
     end
   end
+
+  describe "#selected_failure_reasons_empty?" do
+    subject(:selected_failure_reasons_empty?) do
+      assessment.selected_failure_reasons_empty?(assessment)
+    end
+
+    context "with a decline with no failure reasons" do
+      before { create(:assessment, :decline) }
```
thomasleese, line 269: This line isn’t changing the test as it’s not overwriting the `assessment` variable defined near the top of the file, and therefore not changing the `assessment` in the subject.

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
thomasleese: I think you're still missing a `describe "validations"` block to group together all the expectations related to validation. Ideally they'd go together so it's easier to see what's being tested in this spec file.

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it { is_expected.to validate_presence_of(:recommendation_assessor_note) }
+  end
+
+  context "when recommendation is started" do
+    let(:assessment) { create(:assessment, :started) }
+    it { is_expected.to validate_presence_of(:declaration) }
+  end
+
+  context "when recommendation is declined with failure reasons" do
+    before(:each) do
```
thomasleese, line 35:
```suggestion
    before do
```

It's not a problem, but as it's the default value typically it's not specified.

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      it { is_expected.to validate_presence_of(:recommendation_assessor_note) }
+    end
+
+    context "when recommendation is started" do
+      let(:assessment) { create(:assessment, :started) }
+      it { is_expected.to validate_presence_of(:declaration) }
+    end
+
+    context "when recommendation is declined with failure reasons" do
+      before(:each) do
```
thomasleese, line 29:
```suggestion
      before do
```

spec/forms/assessor_interface/assessment_declaration_decline_form_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        is_expected.not_to validate_presence_of(:recommendation_assessor_note)
+      end
+    end
+  end
+
+  describe "#save" do
+    let(:assessment) { create(:assessment, :decline) }
+    context "when valid" do
+      subject(:save) { form.save }
+      it { is_expected.to be true }
```
thomasleese, line 44: We should probably also test here that it changes the `session`, you could use `.to change` for that, something like:

```rb
it "sets the session" do
  expect { save }.to change { session[:recommendation_assessor_note] }.from(...).to(...)
end
```

thomasleese (commented):
Looks pretty close, just got a couple of small comments.

app/views/assessor_interface/assessment_recommendation_declines/edit.html.erb
```diff
@@ -6,38 +6,38 @@
...
-            <p class="govuk-body govuk-!-margin-bottom-2">Your note (the applicant won’t see this):</p>
-          <% end %>
-
-          <%= govuk_inset_text do %>
-            <%= simple_format failure_reason.assessor_feedback %>
-          <% end %>
-        </li>
-      <% end %>
-    </ul>
+<%unless @assessment.further_information_requests.exists?%>
```
thomasleese, line 9:
```suggestion
<% unless @assessment.further_information_requests.exists? %>
```

thomasleese (approved):

PR #1106, opened 2023-02-08
Assess references that have been received

steventux:
[Trello](https://trello.com/c/FA4eVtHf/1499-assess-references-that-have-been-received)

## Context and changes

_As an assessor that is looking at the reference that have been received about an applicants work history
I need to be able to view the information that a referee has submitted
so that I can  decide whether to award or decline an application_

Adds a _verify reference requests_ and _review reference request_ pages.

I'm not 100% sure what the happy path from verifying all references should be so for the moment this PR takes the user to the application overview. This probably isn't right.

## Verifying reference requests

image

## Reviewing a reference request

image

### DIFF
--- a/app/controllers/assessor_interface/reference_requests_controller.rb
+++ b/app/controllers/assessor_interface/reference_requests_controller.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module AssessorInterface
+  class ReferenceRequestsController < BaseController
+    before_action :authorize_note
+
+    def edit
+      @form = RequestableForm.new(requestable: reference_request)
+    end
+
+    def update
+      @form =
+        RequestableForm.new(
+          requestable: reference_request,
+          **reference_request_form_params,
+        )
+
+      if @form.save
+        redirect_to [
+                      :assessor_interface,
+                      assessment.application_form,
+                      assessment,
+                      :verify_references,
+                    ]
+      else
+        render :edit, status: :unprocessable_entity
+      end
+    end
+
+    private
+
+    def reference_request_form_params
+      params.fetch(:assessor_interface_requestable_form, {}).permit(:passed)
+    end
+
+    def reference_request
+      @reference_request ||=
+        ReferenceRequest
+          .includes(:work_history, assessment: :application_form)
+          .where(
+            assessment_id: params[:assessment_id],
+            assessment: {
+              application_form_id: params[:application_form_id],
+            },
+          )
+          .find(params[:id])
+    end
+
+    delegate :assessment, to: :reference_request
+  end
+end
--- a/app/controllers/assessor_interface/verify_references_controller.rb
+++ b/app/controllers/assessor_interface/verify_references_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module AssessorInterface
+  class VerifyReferencesController < BaseController
+    before_action :authorize_note
+
+    def index
+      @view_object = VerifyReferencesViewObject.new(assessment:)
+      @form = VerifyReferencesForm.new(assessment:)
+    end
+
+    def update
+      @view_object = VerifyReferencesViewObject.new(assessment:)
+      @form = VerifyReferencesForm.new(assessment:, **verify_references_params)
+
+      if @form.save
+        redirect_to [:assessor_interface, assessment.application_form]
+      else
+        render :index, status: :unprocessable_entity
+      end
+    end
+
+    def assessment
+      @assessment ||= Assessment.find(params[:assessment_id])
+    end
+
+    def verify_references_params
+      params.fetch(:assessor_interface_verify_references_form, {}).permit(
+        :references_verified,
+      )
+    end
+  end
+end
--- a/app/forms/assessor_interface/verify_references_form.rb
+++ b/app/forms/assessor_interface/verify_references_form.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class AssessorInterface::VerifyReferencesForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attr_accessor :assessment
+  attribute :references_verified, :boolean
+  validates :references_verified, inclusion: [true, false]
+
+  delegate :application_form, to: :assessment
+
+  def save
+    return false if invalid?
+
+    assessment.update!(references_verified:)
+
+    true
+  end
+end
--- a/app/models/concerns/reviewable.rb
+++ b/app/models/concerns/reviewable.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Reviewable
+  extend ActiveSupport::Concern
+
+  included do
+    belongs_to :assessment
+
+    validates :reviewed_at, presence: true, unless: -> { passed.nil? }
+
+    def reviewed!(passed)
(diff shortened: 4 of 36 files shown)


app/models/concerns/requestable.rb
```diff
@@ -7,7 +7,13 @@ module Requestable
     belongs_to :assessment
 
     enum :state,
-         { requested: "requested", received: "received", expired: "expired" },
+         {
+           requested: "requested",
+           received: "received",
+           expired: "expired",
+           accepted: "accepted",
+           rejected: "rejected",
```
thomasleese, line 15: I was thinking about how to model this, and I wondered if we needed to capture this as two separate states:

- `state`: `requested`, `received`, `expired`
- `review_state`: `unknown`, `accepted`, `rejected`

My reasoning is that if the state is `accepted` that means it must also be `received`, and we've got logic in a few places that checks if the request has been `received?` but this'll be `false` if the state is `accepted` (or `rejected`).

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/141a8080aec38d347acba5dac5cc281afbb84435/app/lib/application_form_status_updater.rb#L126-L128

We could update the various places where we check for `received?` to also include `accepted?` and `rejected?` but I'd be worried about forgetting to do that in a certain place, or just generally it feels right that something could be `received?` and `accepted?` at the same time.

With further information requests, we model this as a nullable boolean `passed`, and I wonder if we can simply move this out of further information request and in to the `Requestable` concern so all requestables have the same mechanism.

steventux: Ah OK good call. I was going on the assumption that `accepted` and `rejected` were implicitly `received` but it's not as explicit as having 2 different state mechanisms. I'll look at integrating with `Requestable`.

app/models/application_form.rb
```diff
@@ -34,6 +34,7 @@
 #  received_reference                            :boolean          default(FALSE), not null
 #  reduced_evidence_accepted                     :boolean          default(FALSE), not null
 #  reference                                     :string(31)       not null
+#  references_verified                           :boolean          default(FALSE)
```
thomasleese, line 37: I think this is better placed on the assessment model actually, to keep things separate we've put most things related to what the assessors do on the assessment.

app/controllers/assessor_interface/verify_references_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def update
+      @view_object = VerifyReferencesViewObject.new(assessment:)
+      @form =
+        VerifyReferencesForm.new(application_form:, **verify_references_params)
+
+      if @form.save
+        redirect_to [:assessor_interface, application_form]
+      else
+        render :index
```
thomasleese, line 20:
```suggestion
        render :index, status: :unprocessable_entity
```

app/forms/assessor_interface/verify_references_form.rb
```diff
@@ -0,0 +1,22 @@
...
+  validates :verify_references, inclusion: [true, false]
+
+  delegate :assessment, to: :application_form
+
+  def save
+    return false if invalid?
+
+    if verify_references == true
+      application_form.update!(references_verified: true)
+    end
```
thomasleese, line 18:
```suggestion
    application_form.update!(references_verified: verify_references)
```

Do we need to allow assessors to say "No" after having previously said "Yes"?

app/views/assessor_interface/verify_references/index.html.erb
```diff
@@ -0,0 +1,64 @@
...
+      label: { text: "Yes, I can progress this application to award or decline" },
+    ) %>
+  <%= f.govuk_radio_button(
+    :verify_references,
+    :false,
+    label: { text: "No, I cannot progress this application yet" },
+  ) %>
+  <% end %>
+
+  <%= f.govuk_submit "Continue", prevent_double_click: false do %>
```
thomasleese, line 61:
```suggestion
 <%= f.govuk_submit prevent_double_click: false do %>
```

The default label is "Continue".

app/views/assessor_interface/verify_references/index.html.erb
```diff
@@ -0,0 +1,64 @@
...
+  <li>the reference from the current or most recent role has been received and is valid (if required)</li>
+  <li>the references that you’ve checked and marked as valid add up to the required number of months for this application to progress</li>
+</ul>
+
+<%= form_with(
+  model: @form,
+  url: [:assessor_interface, @form.application_form, @form.assessment, :update_verify_references],
+  method: :put
+) do |f| %>
+  <%= f.govuk_error_summary %>
```
thomasleese, line 40: We haven't been consistent about this in the app, but I think we should move this to the top of the file so that the error message, if there is one, appears above all the other content.

app/models/concerns/reviewable.rb
```diff
@@ -0,0 +1,21 @@
...
+  included do
+    belongs_to :assessment
+
+    validates :reviewed_at, presence: true, unless: -> { passed.nil? }
+
+    def reviewed!(passed)
+      update!(passed:, reviewed_at: Time.zone.now)
+    end
+
+    def review_state
```
thomasleese, line 15: I wonder if this could simply be called `status` since I think we can use it as the status in all situations where we want to render the status tag next to this reviewable item.

app/controllers/assessor_interface/reference_requests_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+            assessment: {
+              application_form_id: params[:application_form_id],
+            },
+          )
+          .find(params[:id])
+    end
+
+    def assessment
+      reference_request.assessment
+    end
```
thomasleese, line 53:
```suggestion
    delegate :reference_request, to: :assessment
```

app/forms/assessor_interface/reference_request_form.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class AssessorInterface::ReferenceRequestForm
```
thomasleese, line 3: Not a big deal, and we can always do it later, but could we make this form generic? If it accepted a `reviewable` rather than a `reference_request` we could apply it to the requestables we'll be reviewing later on.

thomasleese (commented):
Generally looks good, I've left a few comments, some of them we don't necessarily have to change now.

app/forms/assessor_interface/verify_references_form.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_accessor :assessment
+  attribute :verify_references, :boolean
+  validates :verify_references, inclusion: [true, false]
+
+  delegate :application_form, to: :assessment
+
+  def save
+    return false if invalid?
+
+    assessment.update!(references_verified:)
```
thomasleese, line 16:
```suggestion
    assessment.update!(references_verified: verify_references)
```

I think the field in the form doesn't match the field in the database, although maybe we should rename the one in the form?

steventux: Ah! :man_facepalming: I've amended to use the AR field name throughout.

thomasleese (approved):

thomasleese (approved):

thomasleese:
This has been product reviewed so I'll merge this in.

PR #1107, opened 2023-02-09
Revert "Allow assessing sections before professional standing"

David-Feetenby:
This reverts commit aa62992c1ae6b7d1a6618ef502c9ec6bfd80e102.


thomasleese (approved):
Code looks fine, I think it would be good to explain in the commit message why we're reverting this and linking to the decline flow with an assessor note.

PR #1150, opened 2023-02-20
Update seeds.rb

David-Feetenby:
Update seeds.rb so that the seeded data is up to date


public/location-autocomplete-canonical-list.json
```diff
@@ -56,7 +56,7 @@
   ["Croatia", "country:HR"],
   ["Cuba", "country:CU"],
   ["Curaçao", "territory:CW"],
-  ["Cyprus (European Union)", "country:CY"],
+  ["Xprus (European Union)", "country:CY"],
```
thomasleese, line 59: I don't think this change is quite right!

thomasleese (approved):

PR #1178, opened 2023-02-27
Reminder notifications for outstanding references

steventux:
[Trello](https://trello.com/c/xTeS3ggL/1593-reminder-notifications-to-outstanding-references)

## Context

Referees have 6 weeks to provide references.
Reminders sent at frequency: 2 weeks and 4 weeks

## Changes

- Makes `ReminderEmail` use a polymorphic association to requestables (ie. `FurtherInformationRequest` and `ReferenceRequest`)
- Adds service class and job to send reference request reminders
- Schedules triggering of reference request reminder job

### DIFF
--- a/db/migrate/20230227152913_change_reminder_emails_to_polymorphic_association.rb
+++ b/db/migrate/20230227152913_change_reminder_emails_to_polymorphic_association.rb
@@ -1,0 +1,20 @@
+class ChangeReminderEmailsToPolymorphicAssociation < ActiveRecord::Migration[
+  7.0
+]
+  def change
+    change_table :reminder_emails, bulk: true do |t|
+      t.remove_foreign_key :further_information_requests
+      t.remove_index [:further_information_request_id]
+      t.rename :further_information_request_id, :requestable_id
+      t.string :requestable_type,
+               null: false,
+               default: "FurtherInformationRequest"
+      t.index %i[requestable_type requestable_id]
+    end
+
+    change_column_default :reminder_emails,
+                          :requestable_type,
+                          from: "FurtherInformationRequest",
+                          to: ""
+  end
+end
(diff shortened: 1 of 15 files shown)


db/migrate/20230227152913_change_reminder_emails_to_polymorphic_association.rb
```diff
@@ -0,0 +1,15 @@
...
+  7.0
+]
+  def change
+    change_table :reminder_emails, bulk: true do |t|
+      t.remove_foreign_key :further_information_requests
+      t.remove_index [:further_information_request_id]
+      t.rename :further_information_request_id, :requestable_id
+      t.string :requestable_type,
+               null: false,
+               default: "FurtherInformationRequest"
```
thomasleese, line 11: Does the migration work if we then immediately after this line change the default to `nil`? I don't think it matters much, but feels a bit odd to leave the default here I think.

thomasleese (commented):
One small comment, otherwise looks good to me!

db/migrate/20230227152913_change_reminder_emails_to_polymorphic_association.rb
steventux: Good idea, I'll give it a go.

thomasleese (approved):

db/migrate/20230227152913_change_reminder_emails_to_polymorphic_association.rb
steventux: It works if the `change_column_default` method is called outside of the `change_table` block ie. we issue 2 `ALTER` statements to ensure the default is applied to existing records then reset to an empty string. I've amended the commit.

PR #1202, opened 2023-03-07
Build triage / preliminary checks

steventux:
[Trello](https://trello.com/c/bU8bW4GT/1599-build-triage-preliminary-checks)

### Pre-assessment task item

image

### Preliminary check form

image

### Case note added

image

This PR provides the pre-assessment preliminary check task and the relevant form and case notes.

There are some LOPless requirements in the Trello story which I will follow up with on another PR.


app/controllers/assessor_interface/preliminary_checks_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      if @form.save
+        CreatePreliminaryCheckNote.call(
+          application_form:,
+          author: current_staff,
+          preliminary_check_complete: @form.preliminary_check_complete,
+        )
+
+        redirect_to assessor_interface_application_form_path(application_form)
+      else
+        render :edit
```
thomasleese, line 26:
```suggestion
        render :edit, status: :unprocessable_entity
```

I don't think it matters much but it ensures we use the right HTTP codes.

app/lib/application_form_status_updater.rb
```diff
@@ -84,6 +84,9 @@ def new_status
         "waiting_on"
       elsif received_further_information || received_qualification
         "received"
+      elsif application_form.region.requires_preliminary_check &&
```
thomasleese, line 87: I think we should copy the value of `requires_preliminary_check` to a field on the application form when creating one rather than checking the value of the region. This should avoid problems where an application form already exists and we then enable/disable value on the region. This is similar to what we do for [some other fields](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/cf7bed35f5e803945945b2aa8be8bbafb5e0d012/app/lib/application_form_factory.rb#L15-L19).

app/views/assessor_interface/preliminary_checks/edit.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          :true,
+          link_errors: true,
+          label: { text: "Yes" },
+          ) %>
+    <%= f.govuk_radio_button(
+          :preliminary_check_complete,
+          :false,
+          label: { text: "No, I’ll decline it in the appropriate section" },
+          ) %>
+  <% end %>
```
thomasleese, line 37:
```rb
  <%= f.govuk_collection_radio_buttons :preliminary_check_complete,
                                       [OpenStruct.new(value: :true), OpenStruct.new(value: :false)],
                                       :value,
                                       legend: { size: "s" } %>
```

An alternative way of rendering the radio button would be this, combined with putting the strings in the locale file (for example [the exemption options](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/cf7bed35f5e803945945b2aa8be8bbafb5e0d012/config/locales/helpers.en.yml#L134-L137)) which keeps the view specific to the structure (and in theory makes it easier for a content designer to make changes to the content via the locale file).

I don't mind too strongly, just a suggestion as I've been trying to move us towards using the locale files more.

thomasleese (commented):
Looks pretty good to me, I've left a few comments.

app/lib/application_form_status_updater.rb
steventux: Makes sense, it would certainly produce odd side-effects. Have appended commits to do this, can squash these down if needed.

thomasleese (approved):

PR #1206, opened 2023-03-08
Convert months to years and months

thomasleese:
For the assessors, we want to convert any number of months above 24 to years so it's easier to read.

[Trello Card](https://trello.com/c/xM4ycfKg/1675-app-form-work-history-convert-months-to-years-in-work-history-count)

## Screenshots

Screenshot 2023-03-08 at 11 54 53

### DIFF
--- a/app/views/assessor_interface/assessment_sections/_work_history_summary.html.erb
+++ b/app/views/assessor_interface/assessment_sections/_work_history_summary.html.erb
@@ -11,7 +11,17 @@
 <% end %>
 
 <% if application_form.created_under_new_regulations? %>
+  <%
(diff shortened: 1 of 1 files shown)


steventux (approved):
Potential for moving _the math_  to a helper if we're gonna use this content pattern elsewhere. Not needed here though.

thomasleese:
> Potential for moving _the math_ to a helper if we're gonna use this content pattern elsewhere. Not needed here though.

Yeah, I think that's fair, I did consider it but felt perhaps it was unnecessary for just this like you say. If we were showing the same to the applicant then definitely agree it should be moved elsewhere.

David-Feetenby (approved):

PR #1226, opened 2023-03-13
Notification for LoPS applications which do not pass through preliminary check

steventux:
[Trello](https://trello.com/c/eQKqXYST/1684-notification-after-preliminary-check-passed)

We shouldn't merge this until the preliminary check feature is in use.

This covers the Hong Kong scenario of a country which doesn't require a preliminary check but has LoPS. The email advises the teacher to contact their teaching authority to obtain the certificate/letter.

### DIFF
--- a/app/services/submit_application_form.rb
+++ b/app/services/submit_application_form.rb
@@ -33,7 +33,8 @@ class SubmitApplicationForm
       .application_received
       .deliver_later
 
-    if application_form.teaching_authority_provides_written_statement
+    if !application_form.requires_preliminary_check &&
+         application_form.teaching_authority_provides_written_statement
       TeacherMailer
         .with(teacher: application_form.teacher)
         .initial_checks_passed
--- a/app/services/submit_application_form.rb
+++ b/app/services/submit_application_form.rb
@@ -33,6 +33,13 @@ class SubmitApplicationForm
       .application_received
       .deliver_later
 
+    if application_form.teaching_authority_provides_written_statement
+      TeacherMailer
+        .with(teacher: application_form.teacher)
+        .initial_checks_passed
+        .deliver_later
+    end
+
     if region.teaching_authority_requires_submission_email
       TeachingAuthorityMailer
         .with(application_form:)
--- a/spec/services/submit_application_form_spec.rb
+++ b/spec/services/submit_application_form_spec.rb
@@ -185,6 +185,7 @@ RSpec.describe SubmitApplicationForm do
     context "when teaching authority provides the written statement" do
       before do
         application_form.update!(
+          requires_preliminary_check: false,
           teaching_authority_provides_written_statement: true,
         )
       end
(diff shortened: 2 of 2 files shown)


app/mailers/teacher_mailer.rb
```diff
@@ -71,6 +71,21 @@ def further_information_reminder
     )
   end
 
+  def initial_checks_passed
+    @teaching_authority_certificate =
+      region.teaching_authority_certificate.presence ||
+        region.country.teaching_authority_certificate
+    @teaching_authority_name =
+      region.teaching_authority_name.presence ||
+        region.country.teaching_authority_name
```
thomasleese, line 80: There are some helpers which can do this for us, we could use them directly in the view:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/2938ba3a3916335f944b6ce288a4bceba5bacafb/app/helpers/region_helper.rb#L15-L18

app/views/teacher_mailer/initial_checks_passed.text.erb
```diff
@@ -0,0 +1,25 @@
+Dear <%= application_form_full_name(@application_form) %>
+
+We’ve performed our initial checks on your application.
+
+<%= render "shared/teacher_mailer/reference_number" %>
+
+You should now request your <%= @teaching_authority_certificate %> from <%= @teaching_authority_name %>. Contact them directly and instruct them to send the document to ApplyQTS.Verification@education.gov.uk.
```
thomasleese, line 7: The verification email is defined in the locale file:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/2938ba3a3916335f944b6ce288a4bceba5bacafb/config/locales/en.yml#L16

thomasleese (commented):
Look good, just some comments on the view, but nothing major.

app/views/teacher_mailer/initial_checks_passed.text.erb
thomasleese: Easy to forget! :D

app/services/submit_application_form.rb
```diff
@@ -33,6 +33,13 @@ def call
       .application_received
       .deliver_later
 
+    if application_form.teaching_authority_provides_written_statement
```
thomasleese, line 36:
```suggestion
    if !application_form.requires_preliminary_check && application_form.teaching_authority_provides_written_statement
```

If I understand the logic correctly, should we only be sending it to users where we don't need a preliminary check?

thomasleese: In fact what's the benefit of sending this email given we _don't_ tell this users to not request the document yet:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/32256edb49453e674da3e5bd45abffd1fee29db0/app/views/shared/eligible_region_content_components/_proof_of_recognition.html.erb#L71-L75

steventux: I think this is a sensible condition to add. I originally thought that the two were mutually exclusive but, for example, Ghanaian applicants would receive this email twice without the additional clause - thanks for spotting this.

PR #1277, opened 2023-03-28
Fix background job timeout error reporting

David-Feetenby:
This pr filters out sidekiq errors if they are a faraday error and either ActiveStorage::AnalyzeJob or an ActiveStorage::PurgeJobJob and tells sentry to ignore those errors if the job is still due to be retries

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -13,6 +13,15 @@ def filter_record_not_unique_exception_messages!(event, hint)
   end
 end
 
+def ignore_sidekiq_errors_if_retry(event)
+  job = event&.transaction&.context&.job
+  return false if job.nil?
+  event&.exception.is_a?(Faraday::Error) &&
+    (
+      job.is_a?(ActiveStorage::AnalyzeJob) || job.is_a?(ActiveStorage::PurgeJob)
+    ) && job.retry?
+end
+
 Sentry.init do |config|
   config.breadcrumbs_logger = %i[active_support_logger http_logger]
 
@@ -25,6 +34,7 @@ Sentry.init do |config|
 
   config.before_send =
     lambda do |event, hint|
+      return nil if ignore_sidekiq_errors_if_retry(event)
       filter_record_not_unique_exception_messages!(event, hint)
       filter.filter(event.to_hash)
     end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -84,9 +84,9 @@ ActiveRecord::Schema[7.0].define(version: 2023_03_24_134509) do
     t.bigint "english_language_provider_id"
     t.text "english_language_provider_reference", default: "", null: false
     t.datetime "awarded_at"
-    t.boolean "reduced_evidence_accepted", default: false, null: false
     t.boolean "teaching_authority_provides_written_statement", default: false, null: false
     t.boolean "written_statement_confirmation", default: false, null: false
+    t.boolean "reduced_evidence_accepted", default: false, null: false
     t.boolean "english_language_provider_other", default: false, null: false
     t.datetime "declined_at"
     t.boolean "waiting_on_professional_standing", default: false, null: false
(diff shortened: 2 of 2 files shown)


config/initializers/sentry.rb
```diff
@@ -13,6 +13,17 @@ def filter_record_not_unique_exception_messages!(event, hint)
   end
 end
 
+def ignore_sidekiq_errors_if_retry(event)
+  # Check if the Sidekiq job is still retryable
+  if event&.exception.is_a?(Faraday::Error) &&
+       (
+         event&.transaction&.context&.job.is_a?(ActiveStorage::AnalyzeJob) ||
+           event&.transaction&.context&.job.is_a?(ActiveStorage::PurgeJob)
+       ) && event&.transaction&.context&.job&.retry?
```
thomasleese, line 22: Could we assign `event.transaction.context.job` to a variable so it's easier to read, perhaps something like:

```rb
return unless event&.exception.is_a?(Faraday::Error)

job = event&.transaction&.context&.job
return if job.nil?

if (job.is_a?(ActiveStorage::AnalyzeJob) || job.is_a?(ActiveStorage::PurgeJob) && job.retry?
  ...
end
```

config/initializers/sentry.rb
```diff
@@ -23,6 +34,8 @@ def filter_record_not_unique_exception_messages!(event, hint)
       Rails.application.config.filter_parameters,
     )
 
+  config.before_send = method(:ignore_sidekiq_errors_if_retry)
```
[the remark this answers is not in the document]
David-Feetenby: Cheers, i didn't know that, i'll find a way

config/initializers/sentry.rb
```diff
@@ -13,6 +13,17 @@ def filter_record_not_unique_exception_messages!(event, hint)
...
 end
 
+def ignore_sidekiq_errors_if_retry(event)
+  job = event&.transaction&.context&.job
+  if event&.exception.is_a?(Faraday::Error) &&
+       (
+         job.is_a?(ActiveStorage::AnalyzeJob) ||
+           job.is_a?(ActiveStorage::PurgeJob)
+       ) && job.retry?
+    nil
```
thomasleese, line 23: I don't think this'll quite work, it'll return `nil` from the `ignore_sidekiq_errors_if_retry` method, but later on where we use this method we don't look at the return value, and the crucial bit is that we return `nil` from the `before_send` callback. We probably want to change this method to return a boolean, and then do something like:

```
return nil if ignore_sidekiq_errors_if_retry(event)
```

Assuming you've decided that returning `true` from `ignore_sidekiq_errors_if_retry` means we want to ignore the event.

config/initializers/sentry.rb
```diff
@@ -13,6 +13,17 @@ def filter_record_not_unique_exception_messages!(event, hint)
...
 
+def ignore_sidekiq_errors_if_retry(event)
+  job = event&.transaction&.context&.job
+  if event&.exception.is_a?(Faraday::Error) &&
+       (
+         job.is_a?(ActiveStorage::AnalyzeJob) ||
+           job.is_a?(ActiveStorage::PurgeJob)
+       ) && job.retry?
+    true
+  end
```
thomasleese, line 24: This would work, but whenever you've got something like this, you can remove the `if` statement. For example:

```rb
if <expression>
  return true
end
```

Can just become:

```
return <expression>
```

Since the expression itself will evaluate to true or false.

config/initializers/sentry.rb
```diff
@@ -13,6 +13,14 @@ def filter_record_not_unique_exception_messages!(event, hint)
   end
 end
 
+def ignore_sidekiq_errors_if_retry(event)
+  job = event&.transaction&.context&.job
```
thomasleese, line 17: `job` can be `nil` here (due to the safe navigation operator `&.`), so I think we should add a line below this:

```rb
return false if job.nil?
```

(false because we _don't_ want to ignore this event)

An alternative would be to use the safe navigation operator wherever we access `job`, but I think my suggestion above is more typical Ruby.

thomasleese (approved):

PR #1284, opened 2023-03-30
Handle upload errors

steventux:
[Trello](https://trello.com/c/H2i437GH/1740-handle-timeout-and-not-found-errors-in-upload-controllers)

ActionController::Live module changes the "process" method so that it runs inside a spawned thread.
The "process" method will also handle all filters (before/around/after action hooks)
Usually the authentication happens in a before action filter and if the user is not authenticated Devise will throw :warden
but since this is running in a spawn thread, the Warden middleware doesn't have the chance to catch this symbol and handle it properly.

See https://github.com/rosenfeld/devise-and-streaming

This PR adds some additional authentication methods in a module to
handle actions which could call `send_blob_stream` or other streaming
methods after a Devise `authenticate_teacher|staff!` before hook.
For the reasons explained above this won't handle an unauthenticated
user gracefully, hence the defensive code.

### DIFF
--- a/app/controllers/teacher_interface/uploads_controller.rb
+++ b/app/controllers/teacher_interface/uploads_controller.rb
@@ -5,6 +5,10 @@ module TeacherInterface
     include ActiveStorage::Streaming
     include HandleApplicationFormSection
     include HistoryTrackable
+    include StreamedResponseAuthenticatable
+
+    skip_before_action :authenticate_teacher!, only: :show
+    before_action -> { authenticate_or_redirect(:teacher) }, only: :show
 
     before_action :redirect_unless_draft_or_further_information
     before_action :load_application_form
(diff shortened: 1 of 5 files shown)


app/controllers/teacher_interface/uploads_controller.rb
```diff
@@ -5,10 +5,12 @@ class UploadsController < BaseController
...
     include HandleApplicationFormSection
     include HistoryTrackable
+    include StreamedResponseAuthenticatable
+
+    skip_before_action :authenticate_teacher!, only: :show
+    before_action -> { authenticate_or_redirect(:teacher) }, only: :show
 
     before_action :redirect_unless_draft_or_further_information
-    before_action :load_application_form
-    before_action :load_document
```
thomasleese, line 11: Did we mean to remove these lines? I think we use them in the other views.

steventux: Oops no, amended.

thomasleese (approved):

PR #1322, opened 2023-04-13
Test Defender for Azure malware scan results

steventux:
We have enabled a free trial of Defender for Azure on a per storage resource basis. Defender can run a malware scan on a storage account.
The malware scan is triggered on blob upload, the scan results are saved in a blob index tag.

There are a couple of commits in this PR which persist the result of the scan in `Upload#malware_scan_result`, the scan result is fetched in a background job once the upload is saved, the result isn't available via `ActiveStorage`, hence the additional API call.

We may be able to implement https://trello.com/c/5hRzWDXO/1769-store-virus-scan-state-on-the-upload-model this way.

### DIFF
--- a/app/forms/teacher_interface/upload_form.rb
+++ b/app/forms/teacher_interface/upload_form.rb
@@ -24,16 +24,24 @@ module TeacherInterface
     def update_model
       unless skippable?
         if original_attachment.present?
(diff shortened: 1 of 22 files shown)


steventux:
Closing as spike is complete

PR #1323, opened 2023-04-13
Add initial Terraform code for AKS

thomasleese:
This sets up the project ready to start work on migrating the infrastructure to AKS.

Depends on https://github.com/DFE-Digital/terraform-modules/pull/13.


.github/workflows/actions/deploy-environment/action.yml
```diff
@@ -14,6 +14,10 @@ inputs:
   pr_id:
     description: "Pull Request number"
     required: false
+  platform:
+    description: "PaaS or AKS"
```
saliceti, line 18: Can we use lowercase in the description as it's what is expected in the code?

terraform/aks/terraform.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  required_version = "~> 1.4"
```
saliceti, line 2: From experience it's easier to pin the version exactly (=) for terraform as well as the providers to avoid surprises later on

.github/workflows/actions/deploy-environment/action.yml
thomasleese: Yeah, no problem!

terraform/aks/terraform.tf
thomasleese: Yeah, no problem, I've pinned the versions of Terraform and the provider.

saliceti (approved):

PR #1332, opened 2023-04-18
Rename internal initial assessment status to in progress

David-Feetenby:
This pr is the 1/3 relating to story https://trello.com/c/8Uzg5EM5/1749-rename-internal-initial-assessment-status-to-in-progress

### DIFF
--- a/app/lib/filters/status.rb
+++ b/app/lib/filters/status.rb
@@ -3,13 +3,20 @@
 class Filters::Status < Filters::Base
   def apply
     return scope if statuses.empty?
-
     scope.where(status: statuses)
   end
 
   private
 
   def statuses
-    Array(params[:statuses]).reject(&:blank?)
+    Array(params[:statuses])
+      .reject(&:blank?)
+      .flat_map do |status|
+        if status == "assessment_in_progress"
+          %w[initial_assessment assessment_in_progress]
+        else
+          status
+        end
+      end
   end
 end
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -122,6 +122,7 @@ class ApplicationForm < ApplicationRecord
          draft: "draft",
          submitted: "submitted",
          preliminary_check: "preliminary_check",
+         assessment_in_progress: "assessment_in_progress",
          initial_assessment: "initial_assessment",
          waiting_on: "waiting_on",
          received: "received",
@@ -166,6 +167,7 @@ class ApplicationForm < ApplicationRecord
               overdue
               awarded_pending_checks
               potential_duplicate_in_dqt
+              assessment_in_progress
             ],
           ).or(awarded.where("awarded_at >= ?", 90.days.ago)).or(
             declined.where("declined_at >= ?", 90.days.ago),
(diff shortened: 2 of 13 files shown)


app/models/application_form.rb
```diff
@@ -130,6 +130,7 @@ class ApplicationForm < ApplicationRecord
          awarded: "awarded",
          declined: "declined",
          potential_duplicate_in_dqt: "potential_duplicate_in_dqt",
+         assessment_in_progress: "assessment_in_progress",
```
thomasleese, line 133: Just a minor point, but could we put this next to `initial_assessment` as the order of these vaguely matches the order the application would go through.

app/view_objects/assessor_interface/application_forms_index_view_object.rb
```diff
@@ -46,6 +46,7 @@ def status_filter_options
       awarded
       declined
       potential_duplicate_in_dqt
+      assessment_in_progress
```
thomasleese, line 49: Adding this here means we'll end up with two "Assessment in progress" entries in the filter which would be confusing for assessors, I think instead we should replace the `initial_assessment` entry with `assessment_in_progress` and make sure the status filter will handle that correctly.

app/lib/filters/status.rb
```diff
@@ -3,13 +3,20 @@
...
   end
 
   private
 
   def statuses
-    Array(params[:statuses]).reject(&:blank?)
+    Array(params[:statuses])
+      .reject(&:blank?)
+      .flat_map do |status|
+        if %w[initial_assessment assessment_in_progress].include?(status)
```
thomasleese, line 15:
```suggestion
        if status == "assessment_in_progress"
```

I don't think this will ever come through as `initial_assessment` anymore as we've removed it from the list of statuses: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1332/files#diff-e5bd83cce46d654bc07bb75cb1ba83151390557e0dd03c02798ab6b3368b8960R41

app/lib/filters/status.rb
```diff
@@ -3,13 +3,20 @@
...
+          %w[assessment_in_progress]
```
thomasleese, line 16:
```suggestion
          %w[initial_assessment assessment_in_progress]
```

If the user selects the `assessment_in_progress` status, we want it to also include any existing applications with the `initial_assessment` status.

David-Feetenby: Would that not just write them to a flat string of the two combined?

thomasleese: Yes, it would, because we want to then filter the database like:

```rb
where(status: %w[initial_assessment assessment_in_progress])
```

thomasleese: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/a0932a42e8e956dab49dd3ad7220f681b56cc8a4/app/lib/filters/status.rb#L7

This is where we do the filter on the database. During this migration, we'll have applications in both the `initial_assessment` and `assessment_in_progress` status, which we want to consider the same, so if a user chooses to filter on "In progress" applications, we'll need to show them applications in both status.

thomasleese (approved):

David-Feetenby:
[Rename internal "initial assessment" status to "in progress"](https://trello.com/c/8Uzg5EM5/1749-rename-internal-initial-assessment-status-to-in-progress)

PR #1341, opened 2023-04-19
Limit Slack notifications about Github workflow failures

steventux:
We want these for build, rspec and deploy_nonprod but only when the main branch is pushed, otherwise an rspec failure on a PR branch will notify.


thomasleese:
I'm going to merge this as @steventux is not working today.

PR #1344, opened 2023-04-20
Migrate initial assessment status

David-Feetenby:
Chang applications application form status to assessment in progress where it is currently initial assessment.

### DIFF
--- a/lib/tasks/update_application_form_status.rake
+++ b/lib/tasks/update_application_form_status.rake
@@ -1,0 +1,8 @@
+desc "Update application status"
+task update_application_form_status_initial_assessment: :environment do
+  # fetch all application forms with 'initial assessment' status
+  forms = ApplicationForm.where(status: "initial_assessment")
+
+  # update the status for each form
+  forms.update_all(state: "assessment_in_progress")
+end


lib/tasks/update_application_form_status.rake
```diff
@@ -0,0 +1,10 @@
+namespace :db do
+  desc "Update application status"
+  task update_status: :environment do
+    # fetch all application forms with 'initial assessment' status
+    forms = ApplicationForm.where(status: "initial_assessment")
+
+    # update the status for each form
+    forms.each { |form| form.update(status: "assessment_in_progress") }
```
thomasleese, line 8: This would work, but it would be quite slow as it would require a query to do the database for each application form. If we do something like `forms.update_all(state: "assessment_in_progress")` that'll do the update in one single query and let the database handle that for us.

lib/tasks/update_application_form_status.rake
```diff
@@ -0,0 +1,7 @@
+namespace :db do
+  desc "Update application status"
+  task update_status: :environment do
```
thomasleese, line 3:
```suggestion
  task update_application_form_status_initial_assessment: :environment do
```

Although it's a temporary task, I think it's still worth giving it a clearer name. Worth making the filename match the task too I think.

lib/tasks/update_application_form_status.rake
```diff
@@ -0,0 +1,7 @@
+namespace :db do
```
thomasleese, line 1: I don't think we need a namespace for this task as it's only one on its own.

lib/tasks/update_application_form_status.rake
```diff
@@ -0,0 +1,7 @@
...
+  desc "Update application status"
+  task update_status: :environment do
+    # update all application forms with 'initial assessment' status
+    forms.update_all(state: "assessment_in_progress")
```
thomasleese, line 5: It doesn't look like a `forms` variable is defined here so I'm not sure the task will work at the moment.

thomasleese (approved):
It looks good to me!

David-Feetenby:
[Rename internal "initial assessment" status to "in progress"](https://trello.com/c/8Uzg5EM5/1749-rename-internal-initial-assessment-status-to-in-progress)

PR #1357, opened 2023-04-26
Virus scan existing uploads

steventux:
[Trello](https://trello.com/c/itCez4ET/1768-virus-scan-existing-uploads)

As part of introducing malware scanning for ActiveStorage uploads we need to scan any blobs/files in our production storage account.
We can initiate a malware scan by re-sending the blob data for each unscanned upload.
This PR introduces a rate-limited job which works with a default batch size of 1000 uploads. The job finds uploads with a pending malware scan status and resends the blob data for each.
After a delay of 1 minute the scan result is fetched for the same upload in another background job, this allows for the malware scan to complete before we attempt to update the `Upload#malware_scan_result` field.

### DIFF
--- a/app/jobs/resend_stored_blob_data_job.rb
+++ b/app/jobs/resend_stored_blob_data_job.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class ResendStoredBlobDataJob < ApplicationJob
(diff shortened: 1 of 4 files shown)


thomasleese (approved):
Looks good!

PR #1361, opened 2023-04-27
Support AKS secrets

thomasleese:
This changes how we fetch the secrets to support both the old infrastructure on the PaaS and the new infrastructure on AKS.


nickdevdfe (approved):
changes appear correct

PR #1381, opened 2023-05-03
Replace initial assessment label

David-Feetenby:
Replace initial assessment timeline events with new assessment in progress label


steventux (commented):
I think we should omit the change to `db/schema.rb`

steventux (approved):

PR #1419, opened 2023-05-24
Help users submit the FI they have uploaded

David-Feetenby:
Add content to stop confusion between completed/submitted

### DIFF
--- a/config/locales/teacher_interface.en.yml
+++ b/config/locales/teacher_interface.en.yml
@@ -33,6 +33,7 @@ en:
         heading: English language proficiency
     further_information_request:
       show:
+        check: Check your response
         failure_reason:
           age_range: Tell us more about the age range you can teach
           applicant_already_dqt: Additional personal information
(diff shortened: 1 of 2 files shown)


config/locales/teacher_interface.en.yml
```diff
@@ -13,6 +13,7 @@ en:
           professional_standing_request_expired: Your application has been declined as we did not receive your %{certificate_name} from %{teaching_authority_name} within the 90-day period.
         draft:
           check: Check your application
+          check_further_information: Check your response
```
thomasleese, line 16: Since we're only using this for further information, we should put it in the `teacher_interface. further_information_request.show` key.

In the view, this will mean that we only need to write `t(".check")` as we're in the `teacher_interface/further_information_requests/show` view, that'll get automatically converted to `t("teacher_interface.further_information_request.show.check")`.

thomasleese (approved):

PR #1423, opened 2023-05-24
Check for applicant in DQT when application is submitted

steventux:
https://trello.com/c/GXIBUN1U/1849-spike-upfront-duplication-check

When an application is submitted, query the DQT API with the applicant's details and save the first matched record on the application.

image


thomasleese (approved):
Looks good to me!

thomasleese (approved):

PR #1430, opened 2023-05-26
Add decline reason for work history

David-Feetenby:
Add  new decline reason for insufficient work history duration

### DIFF
--- a/app/lib/failure_reasons.rb
+++ b/app/lib/failure_reasons.rb
@@ -29,6 +29,7 @@ class FailureReasons
       "teaching_qualifications_from_ineligible_country",
     TEACHING_QUALIFICATIONS_NOT_AT_REQUIRED_LEVEL =
       "teaching_qualifications_not_at_required_level",
+    WORK_HISTORY_DURATION = "work_history_duration",
   ].freeze
 
   FURTHER_INFORMATIONABLE = [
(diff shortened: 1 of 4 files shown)


app/lib/failure_reasons.rb
```diff
@@ -29,6 +29,8 @@ class FailureReasons
       "teaching_qualifications_from_ineligible_country",
     TEACHING_QUALIFICATIONS_NOT_AT_REQUIRED_LEVEL =
       "teaching_qualifications_not_at_required_level",
+    TEACHING_QUALIFICATIONS_COMPLETED_IN_ELIGIBLE_COUNTRY =
```
thomasleese, line 32: I'm not sure that we need this line, it'll mean that we end up defining `TEACHING_QUALIFICATIONS_COMPLETED_IN_ELIGIBLE_COUNTRY` as `work_history_duration`.

thomasleese (commented):
Just a minor comment, otherwise it looks good!

app/lib/failure_reasons.rb
David-Feetenby: Not sure why i left that in 😆

app/lib/failure_reasons.rb
```diff
@@ -29,6 +29,8 @@ class FailureReasons
       "teaching_qualifications_from_ineligible_country",
     TEACHING_QUALIFICATIONS_NOT_AT_REQUIRED_LEVEL =
       "teaching_qualifications_not_at_required_level",
+    WORK_HISTORY_DURATION =
+      WORK_HISTORY_DURATION = "work_history_duration",
```
thomasleese, line 33:
```suggestion
    WORK_HISTORY_DURATION =
      "work_history_duration",
```

The variable only needs to defined once :)

thomasleese (approved):

PR #1439, opened 2023-06-01
Work history being entered from before applicants graduated/qualified

David-Feetenby:
Add content and banner logic to help stop unneeded work history

### DIFF
--- a/app/helpers/application_form_helper.rb
+++ b/app/helpers/application_form_helper.rb
@@ -108,4 +108,16 @@ module ApplicationFormHelper
       { key: { text: key }, value: { text: value }, actions: actions || [] }
     end
   end
+
+  def application_form_display_work_history_before_teaching_qualification_banner?(
+    application_form
+  )
+    earliest_certificate_date =
+      application_form.teaching_qualification.certificate_date
+    earliest_work_history_date =
+      application_form.work_histories.pluck(:start_date).compact.min
+
+    earliest_certificate_date.present? && earliest_work_history_date.present? &&
+      earliest_work_history_date < earliest_certificate_date
+  end
 end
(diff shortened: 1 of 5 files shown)


spec/system/teacher_interface/check_answers_spec.rb
```diff
@@ -67,6 +67,70 @@
...
+      )
+    end
+  end
+
+  it "should not display the banner if both earliest_certificate_date and earliest_work_history_date are blank" do
+    @application_form.qualifications = []
+    @application_form.work_histories = []
+
+    #reload the page
+    visit current_path
```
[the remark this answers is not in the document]
David-Feetenby: The test seems to fail if i don't specify it. Not sure why, but currently in the process of refactoring them anyway so hopefully will just get superceded.

app/helpers/application_form_helper.rb
```diff
@@ -108,4 +108,14 @@ def application_form_summary_rows(
       { key: { text: key }, value: { text: value }, actions: actions || [] }
     end
   end
+
+  def display_banner?(application_form)
```
thomasleese, line 112: Could we clarify the name of this? Something that explains which banner is being displayed would be ideal.

spec/system/teacher_interface/check_answers_spec.rb
```diff
@@ -67,10 +67,74 @@
     end
   end
 
+  it "display banner" do
```
thomasleese, line 70: Would be good to explain which banner we're testing here, but the test itself looks good!

app/helpers/application_form_helper.rb
```diff
@@ -108,4 +108,14 @@ def application_form_summary_rows(
       { key: { text: key }, value: { text: value }, actions: actions || [] }
     end
   end
+
+  def display_work_history_banner?(application_form)
```
thomasleese, line 112:
```suggestion
  def application_form_display_work_history_before_teaching_qualification_banner?(application_form)
```

The helpers are global methods across all the views, so for a convention we've prefixed them with the model they're related to `application_form`.

thomasleese (approved):

PR #1433, opened 2023-05-31
Add production AKS environment

thomasleese:
This adds the test, preproduction and production environments for the AKS platform, allowing us to deploy the entire application pipeline in AKS.

PR #1443, opened 2023-06-02
AFQTS DNS migration changes - frontdoor and dns only

nickdevdfe:
Migration changes for frontdoor and dns has been updated, tested and deployed for the core part only.
Config's have been updated for the environments only  and not deployed.


terraform/custom_domains/environment_domains/main.tf
```diff
@@ -0,0 +1,59 @@
+module "domains" {
+  source              = "git::https://github.com/DFE-Digital/terraform-modules.git//domains/environment_domains?ref=0.5.4"
```
saliceti, line 2: Can we use `ref=stable` to take advantage of module improvements automatically?

terraform/custom_domains/environment_domains/terraform.tf
```diff
@@ -0,0 +1,25 @@
+terraform {
+
+  required_version = "= 1.3.5"
```
saliceti, line 3: Can we align the version to the app terraform? https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/terraform/aks/terraform.tf#L2

terraform/custom_domains/environment_domains/terraform.tf
```diff
@@ -0,0 +1,25 @@
...
+
+provider "azurerm" {
+  features {}
+
+  skip_provider_registration = true
+}
+
+provider "azurerm" {
+  features {}
+  version         = "3.45.0"
```
saliceti, line 22: This is redundant with the required_providers configuration and [deprecated](https://developer.hashicorp.com/terraform/language/providers/configuration#version-deprecated)

terraform/custom_domains/environment_domains/terraform.tf
nickdevdfe: This has been moved to another PR, it can be tested and updated there. https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1447

terraform/custom_domains/environment_domains/terraform.tf
```diff
@@ -0,0 +1,25 @@
+terraform {
+
+  required_version = "= 1.3.5"
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "3.45.0"
```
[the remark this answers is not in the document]
nickdevdfe: This has been moved to another PR, it can be tested and updated there. https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1447

terraform/custom_domains/infrastructure/terraform.tf
```diff
@@ -0,0 +1,22 @@
+terraform {
+
+  required_version = "= 1.3.5"
```
saliceti, line 3: Can we use the same version as https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/terraform/aks/terraform.tf#L2

terraform/custom_domains/infrastructure/terraform.tf
```diff
@@ -0,0 +1,22 @@
...
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "3.43.0"
```
saliceti, line 7: Can we use the same version as https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/terraform/aks/terraform.tf#L9

Makefile
```diff
@@ -7,45 +7,52 @@ SERVICE_SHORT=afqts
 help: ## Show this help
 	@grep -E '^[a-zA-Z\.\-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
+
+.PHONY: full
```
saliceti, line 11: Can this be removed?

saliceti (approved):

Makefile
```diff
@@ -8,44 +8,44 @@ help: ## Show this help
...
 	$(eval REGION=UK South)
 	$(eval STORAGE_ACCOUNT_SUFFIX=sa)
 	$(eval KEY_VAULT_SECRET_NAME=APPLICATION)
 
 .PHONY: dev
-dev: paas
+dev: paas  ## Sets the paas environment variables for dev PAAS deployment
 	$(eval DEPLOY_ENV=dev)
-	$(eval AZURE_SUBSCRIPTION=s165-teachingqualificationsservice-development)
+	$(eval AZURE_SUBSCRIPTION=s189-teachingqualificationsservice-development)
```
thomasleese, line 26: Why do we change the subscription for the PaaS environment?

Makefile
```diff
@@ -228,3 +228,45 @@ disable-maintenance: read-tf-config ## make dev disable-maintenance / make produ
...
 	cf unmap-route apply-for-qts-unavailable london.cloudapps.digital --hostname apply-for-qts-in-england-${DEPLOY_ENV}
 	cf delete apply-for-qts-unavailable -r -f
+
+validate-domain-resources: set-what-if domain-azure-resources # make publish validate-domain-resources AUTO_APPROVE=1
+
+deploy-domain-resources: check-auto-approve domain-azure-resources # make publish deploy-domain-resources AUTO_APPROVE=1
+
+.PHONY: afqts_domain
+afqts_domain:   ## runs a script to config variables for setting up dns
+	$(eval include global_config/afqts_domain.sh)
```
thomasleese, line 238:
```suggestion
.PHONY: domain
domain:   ## runs a script to config variables for setting up dns
	$(eval include global_config/domain.sh)
```

```suggestion
.PHONY: domain
domain:   ## runs a script to config variables for setting up dns
```

There's no need to include a `afqts_` prefix in this Makefile since we're already within the Apply for QTS project. It's worth noting that we only call it `afqts` where Azure limits the length of our names, and it's not commonly used as an acronym on this project otherwise.

global_config/afqts_domain.sh
```diff
@@ -0,0 +1,10 @@
+CONFIG=production
+CONFIG_SHORT=pd
+AZURE_SUBSCRIPTION=s189-teacher-services-cloud-production
+RESOURCE_NAME_PREFIX=s189p01
+ENV_TAG=Prod
+DNS_ZONE=afqts
+RESOURCE_GROUP_NAME=${RESOURCE_NAME_PREFIX}-${DNS_ZONE}domains-rg
+KEYVAULT_NAME=${RESOURCE_NAME_PREFIX}-${DNS_ZONE}domains-kv
+STORAGE_ACCOUNT_NAME=${RESOURCE_NAME_PREFIX}${DNS_ZONE}domainstf
+DOMAINS_ID=afqts
```
thomasleese, line 10: This looks very similar to the production global configuration we're going to add for AKS (https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1433/files) but with some additional fields. Can we use the same config?

Especially the variables `DOMAINS_ID` and `DNS_ZONE`, these would never change so we can define them as static variables in the `Makefile`?

thomasleese (changes requested):
Could we add some additional information to the commit explaining what this change is for?

I understand that I might not be the target audience of this PR, but as a developer of the application it's very useful to understand the infrastructure and how it's deployed. The commit message doesn't explain _why_ we're making the change here so that would be helpful to add, especially if we want to look back at some previous commit messages in the future to understand why a change was made.

Makefile
```diff
@@ -92,7 +92,7 @@ read-deployment-config:
 	$(eval API_APP_NAME="apply-for-qts-in-england-$(DEPLOY_ENV)")
 
 .PHONY: set-azure-account
-set-azure-account: ${environment}
+set-azure-account:
```
thomasleese, line 95: Nice change!

Makefile
```diff
@@ -8,13 +8,13 @@ help: ## Show this help
 	@grep -E '^[a-zA-Z\.\-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 .PHONY: paas
-paas:
+paas:  ## Sets the paas environment variables for over all PAAS deployment
```
thomasleese, line 11: I intentionally left this `paas` task undocumented as it's not useful on its own (all the individual environment targets have `paas` as a dependency). With the `##` comment here it ends up being listed in `make help` which is a little confusing if you've not used the `Makefile` before.

We could have something like this, but I would question the benefit of a comment, over for example changing the target to be more descriptive.

```suggestion
paas:  # Set the PaaS platform variables
```

A better option might:

```suggestion
set-paas-platform:
```

Makefile
nickdevdfe: This has been corrected.

Makefile
nickdevdfe: script name shorten as suggested

terraform/custom_domains/infrastructure/terraform.tf
```diff
@@ -0,0 +1,22 @@
...
+}
+
+provider "azurerm" {
+  features {}
+
+  skip_provider_registration = true
+  #   subscription_id            = try(local.azure_credentials.subscriptionId, null)
+  #   client_id                  = try(local.azure_credentials.clientId, null)
+  #   client_secret              = try(local.azure_credentials.clientSecret, null)
+  #   tenant_id                  = try(local.azure_credentials.tenantId, null)
```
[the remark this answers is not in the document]
nickdevdfe: This was originally left in for a github actions update - but that has been moved to a Trello ticket, so can be removed

global_config/afqts_domain.sh
nickdevdfe: Reviewed with Thomas, Will clean up in upcoming PR's

PR #1388, opened 2023-05-10
Add deployment pipeline for AKS

thomasleese:
This adds the ability to deploy the `dev` and `review` environment on AKS using GitHub Actions.

[Trello Card](https://trello.com/c/bqMVGldA/1803-add-deployment-pipeline-for-aks)

PR #1445, opened 2023-06-02
Remove PaaS review environment

thomasleese:
This removes the old PaaS-based `review` environment which is no longer being used as review apps are now deployed to AKS since #1388.

I've also renamed `pr_id` to `PULL_REQUEST_NUMBER` to be clearer on the variable name.


nickdevdfe (approved):
Reviewed code it has been tested locally.

PR #1454, opened 2023-06-06
Add apk upgrade ssl to docker file

David-Feetenby:
Snyk found vulnerability with:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,8 @@ RUN apk add --update --no-cache tzdata && \
     cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
     echo "Europe/London" > /etc/timezone
 
-#Upgrade openssl to latest version
-RUN apk upgrade openssl
-
-#Upgrade libssl3 to latest version
-RUN apk upgrade libssl3
+# Upgrade ssl and crypto libraries to latest version
+RUN apk upgrade openssl libssl3 libcrypto3
 
 # build-base: dependencies for bundle
 # yarn: node package manager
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -12,6 +12,12 @@ RUN apk add --update --no-cache tzdata && \
     cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
     echo "Europe/London" > /etc/timezone
 
+#Upgrade openssl to latest version
+RUN apk upgrade openssl
+
+#Upgrade libssl3 to latest version
+RUN apk upgrade libssl3
```
nickdevdfe, line 19: I believe this should fix it
RUN apk add --no-cache libssl3=3.1.1-r1

thomasleese: Looking in the build logs, it does seem to be upgrading: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/actions/runs/5187084105/jobs/9349009662?pr=1454

```
#12 [builder  5/14] RUN apk upgrade libssl3
#0 0.591 (1/1) Upgrading libssl3 (3.1.0-r4 -> 3.1.1-r1)
#0 0.681 OK: 26 MiB in 37 packages
#12 DONE 0.8s
```

I'm also not sure we'd want to pin the version as really we want version 3.1.1-r1 or above.

nickdevdfe: When ruby-alpine updates, we'll remove this command.

thomasleese (approved):

Dockerfile
steventux: I think the issue is that we are only upgrading the relevant libs on the builder image and not the production image.
See https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1454/commits/85d95655e88213eae52e07759eb9d6775325d954

PR #1463, opened 2023-06-07
[SPIKE] Authenticate assessors using azure AD

steventux:
https://trello.com/c/feUgIKnD/1889-spike-authenticate-assessors-using-azure-ad


app/controllers/staff/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,20 @@
+class Staff::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def azure_activedirectory_v2
+    auth = request.env["omniauth.auth"]
+    email = auth["info"]["email"]
+    provider = auth["provider"]
+    uid = auth["uid"]
+
+    @staff = Staff.find_by(email:, provider: "azure_activedirectory_v2")
```
thomasleese, line 8:
```suggestion
    @staff = Staff.find_by(email:)
```

I think this would allow us to "migrate" the accounts over by getting assessors to sign in via ActiveDirectory and then it would link their account up using the email address - is this an approach we might want to take?

thomasleese:
Replaced by #1521

PR #1485, opened 2023-06-15
Refactor task lists

thomasleese:
This refactors how we render task lists to move them in to a common component with a specific structure, which then allows us to be clearer in the view objects.

This is necessary as a precursor to some upcoming work related to preliminary checks and consent which means the task list for assessors will be more dynamic.


richardpattinson (approved):
Looks great, we can merge with the dependent PR.

PR #1491, opened 2023-06-19
Don't run production smoke test with credentials

thomasleese:
The production service should be open so we don't need to run the smoke test against it with credentials.


thomasleese:
[Apply for QTS: Add StatusCake](https://trello.com/c/EvuDihrh/429-apply-for-qts-add-statuscake)

nickdevdfe (approved):
Reviewed and approved.

PR #1497, opened 2023-06-20
Fix typo in monitoring configuration

thomasleese:
The interpolation should use the `$` rather than `#`.


nickdevdfe (approved):
Reviewed and approved.

PR #1500, opened 2023-06-21
Revert "Remove tasks jobs and services for resending blob data"

thomasleese:
This reverts commit 02148817c58b5fb7dd4aeaeb78ae966da41aa532.

Sometimes it seems like the virus scanner gets stuck and we end up with a number of pending uploads, these tasks will make it possible to resend them so they get scanned again.


richardpattinson (approved):
Im happy to revert the previous PR.

PR #1501, opened 2023-06-21
Fix flaky eligible region content test

thomasleese:
This improves the reliability of this test by fixing the country code to _not_ be Northern Ireland, which has it's own content for this page.

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/actions/runs/5308639745/jobs/9608334123?pr=1488 is an example of where this test fails.


richardpattinson (approved):
Lgtm!

PR #1502, opened 2023-06-21
Fix flaky update assessment section test

thomasleese:
This fixes the flaky test by not hard coding the ID of the staff user, so we don't end up with two staff users with the same ID.

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/actions/runs/5334743623/jobs/9666943566?pr=1500 is an example of this test failing.


richardpattinson (approved):
Nice! Ive not run into this one yet but removing the hard coding seems good.

PR #1505, opened 2023-06-22
Don't expire old tokens on sign in

thomasleese:
This is to better handle Outlook's "safe links" feature where it navigates to the page before the user does, and so in certain situations we end up treating the user as already signed in when they come to do it.

A better fix might be to look at the user-agent and not record the sign in if it's coming from Outlook, but I want to see if this change fixes the immediate problem.

[Trello Card](https://trello.com/c/DnyC5BTd/1978-invalid-or-expired-login-link-using-outlook)


richardpattinson (approved):
Nice change, approved =)

AbigailMcP:
[Base AKS config + review apps](https://trello.com/c/cSNsmW8m/63-base-aks-config-review-apps)

PR #1506, opened 2023-06-22
[1975] Add task to update status of waiting on application forms

richardpattinson:
Expected:
When a work reference becomes overdue
The 'verify reference requests' has a status of overdue;

And the application’s status is moved to overdue

Actual:

The application status is showing as waiting on (see attachment for screenshots).

It's done when

The application status is overdue, when the work reference and verify reference requests are in a status of overdue

Any applications that have overdue work references (and not actioned) have a status of overdue.

In this PR there is a fix to ensure when we update a timeline event that status will be updated. There is also a rake task to backfill any application forms that have incorrect statuses.

### DIFF
--- a/lib/tasks/application_forms.rake
+++ b/lib/tasks/application_forms.rake
@@ -61,4 +61,17 @@ namespace :application_forms do
       puts "#{application_form.reference}: #{application_form.status}"
     end
   end
+
+  desc "Update status of application forms in waiting on to overdue"
+  task update_waiting_on_applications: :environment do
+    user = "Expirer"
+    assessment_ids = ReferenceRequest.expired.select(:assessment_id)
+    ApplicationForm
+      .waiting_on
+      .joins(:assessment)
+      .where(assessment: { id: assessment_ids })
+      .each do |application_form|
+        ApplicationFormStatusUpdater.call(application_form:, user:)
+      end
+  end
 end
--- a/spec/services/expire_requestable_spec.rb
+++ b/spec/services/expire_requestable_spec.rb
@@ -18,6 +18,14 @@ RSpec.describe ExpireRequestable do
           creator: user,
         )
       end
+
+      it "calls the status updater" do
+        expect(ApplicationFormStatusUpdater).to receive(:call).with(
+          application_form: requestable.application_form,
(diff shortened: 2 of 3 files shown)


lib/tasks/application_forms.rake
```diff
@@ -61,4 +61,12 @@ namespace :application_forms do
       puts "#{application_form.reference}: #{application_form.status}"
     end
   end
+
+  desc "Update status of application forms in waiting on to overdue"
+  task :update_waiting_on_applications, :environment do
+    user = "Status updater"
```
thomasleese, line 67:
```suggestion
    user = "Expirer"
```

Since we're fixing the `ExpireRequestable` service, I think we should show that as the user which changed the status.

lib/tasks/application_forms.rake
```diff
@@ -61,4 +61,12 @@ namespace :application_forms do
...
+    ApplicationForm.waiting_on.each do |application_form|
```
thomasleese, line 68: I think we could also filter on `ApplicationForm`s which are `waiting_on` _and_ have an expired requestable, reducing the number of application forms we need to update the status of.

spec/services/expire_requestable_spec.rb
```diff
@@ -18,6 +18,11 @@
           creator: user,
         )
       end
+
+      it "calls the status updater" do
+        expect(ApplicationFormStatusUpdater).to receive(:call).at_least(:once)
```
thomasleese, line 23:
```suggestion
        expect(ApplicationFormStatusUpdater).to receive(:call).with(application_form: requestable.application_form, user:).at_least(:once)
```

To just check we're passing the right arguments in.

thomasleese (commented):
Looks good, just got a few comments.

spec/services/expire_requestable_spec.rb
richardpattinson: Should be in now

lib/tasks/application_forms.rake
richardpattinson: Updated =)

thomasleese (approved):

PR #1499, opened 2023-06-21
Improve preliminary check flow

thomasleese:
This implements the new preliminary check flow based on cut-down assessment sections rather than its own view.

Depends on #1485 

[Trello Card](https://trello.com/c/sMOGMUWl/1971-revised-preliminary-checks-flow)

PR #1512, opened 2023-06-27
Remove old preliminary check code

thomasleese:
This removes code related to the old preliminary check process which has since been replaced with #1499.


richardpattinson (approved):
Looks great; no changes requested.

PR #1513, opened 2023-06-27
Improve preliminary check status

thomasleese:
This ensures that the status is always set correctly, and improves the readability of the code. It's necessary if an assessors reviews the preliminary assessment section but then chooses not to decline immediately, this will currently result in the status changing to "assessment in progress" but I think we should leave it as "preliminary check".


richardpattinson (approved):
Looks great to me.

PR #1521, opened 2023-06-29
[1991] Implement authentication via azure ad

richardpattinson:
This PR carries on the spike work: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1463. 

We currently use devise with a username and password stored in the service DB to authenticate internal users.

Now that we all have Microsoft AD accounts we will use that instead. This will give us centralised user management and 2FA etc

This PR takes the work from the spike and adds a feature flag for authentication via azure, the logic to make it run and some content changes to reflect the new sign in flow.


config/feature_flags.yml
```diff
@@ -1,4 +1,14 @@
 feature_flags:
+  enable_login_with_azure:
```
thomasleese, line 2:
```suggestion
  sign_in_with_active_directory:
```

We don't need `enable` as the flag itself is enabling something :)

config/feature_flags.yml
```diff
@@ -1,4 +1,14 @@
...
+    author: Richard Pattinson
+    description: Allow users to use the login with azure feature.
```
thomasleese, line 4:
```suggestion
    description: Allow users to sign in using accounts in Active Directory.
```

The functionality is part of Azure, but it's specifically signing in with accounts stored in Active Directory.

config/initializers/devise.rb
```diff
@@ -370,7 +370,11 @@
   # When using the :trackable module, set to true to consider magic link tokens
   # generated before the user's current sign in time to be expired. In other words,
   # each time you sign in, all existing magic links will be considered invalid.
-  # config.passwordless_expire_old_tokens_on_sign_in = true
+  config.passwordless_expire_old_tokens_on_sign_in = true
```
thomasleese, line 373:
```suggestion
  # config.passwordless_expire_old_tokens_on_sign_in = true
```

This should remain commented from https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1505.

db/migrate/20230607124047_add_oauth_attributes_to_staff.rb
```diff
@@ -0,0 +1,8 @@
+class AddOauthAttributesToStaff < ActiveRecord::Migration[7.0]
+  def change
+    change_table :staff, bulk: true do |t|
+      t.string :provider
```
thomasleese, line 4: I'm wondering if there's any benefit to storing these? We don't actually use them for anything, especially `provider` which we know there will only ever be one.

spec/system/assessor_interface/authentication_spec.rb
```diff
@@ -4,6 +4,7 @@
 
 RSpec.describe "Assessor authentication", type: :system do
   it "allows signing in and signing out" do
+    FeatureFlags::FeatureFlag.deactivate(:enable_login_with_azure)
```
thomasleese, line 7: Could we move these in to a `given_sign_in_with_active_directory_is_deactive` method, just to match the convention of these system tests.

app/views/staff/sessions/new.html.erb
```diff
@@ -1,18 +1,24 @@
...
+      <% end %>
     <% end %>
+
+    <%= f.govuk_submit %>
   <% end %>
+<% end %>
 
-  <%= f.govuk_submit %>
+<% if FeatureFlags::FeatureFlag.active?(:enable_login_with_azure) %>
+  <%= govuk_button_to "Sign in with Active Directory", omniauth_authorize_path(:staff, :azure_activedirectory_v2), method: :post, data: { turbo: "false" } , id: "azure_ad_login"%>
```
thomasleese, line 21:
```suggestion
  <%= govuk_button_to "Sign in with Active Directory", omniauth_authorize_path(:staff, :azure_activedirectory_v2), method: :post, id: "azure_ad_login"%>
```

Do we need the `turbo: "false"` bit?

thomasleese, line 21:
```suggestion
  <%= govuk_button_to "Sign in with Active Directory", omniauth_authorize_path(:staff, :azure_activedirectory_v2), method: :post, data: { turbo: "false" } , id: "button-sign-in-with-active-directory"%>
```

We should be clear that it's a button if we're using an ID.

thomasleese (commented):
Generally looks good, got a number of small comments.

app/views/staff/sessions/new.html.erb
richardpattinson: I think what the turbo: false is doing is disabiling turbolinks from packaging this request as an ajax request. Im unsure if it should go.

db/migrate/20230607124047_add_oauth_attributes_to_staff.rb
richardpattinson: Hmmm; I'm not sure if we could need these going forward. Will we ever want to look up a provider for a staff member? Keeping it on there seems low cost.

.env.test
```diff
@@ -0,0 +1,7 @@
+BIGQUERY_DISABLE=false
+BIGQUERY_API_JSON_KEY={ replaceMe: 'I should be a copy of a BigQuery JSON key' }
+BIGQUERY_DATASET=events_local
+BIGQUERY_PROJECT_ID=apply-for-qts-in-england
+BIGQUERY_TABLE_NAME=events
+MICROSOFT_OAUTH_CLIENT_ID=test
+MICROSOFT_OAUTH_CLIENT_SECRET=test
```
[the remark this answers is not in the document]
richardpattinson: How should this be rolled into the config file?

db/migrate/20230607124047_add_oauth_attributes_to_staff.rb
thomasleese: There will only ever be one provider, `azure_ad` so we can't definitely get rid of `provider` I reckon. Keeping `uid` seems more reasonable, but I'd be tempted to rename it to `azure_ad_uid` or something because `uid` on its own is a bit unclear what it is.

app/views/staff/sessions/new.html.erb
thomasleese: We don't use turbo links in this repo so I think we can get rid of it.

.env.test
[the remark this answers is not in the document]
thomasleese: I think we can just delete the file, in our tests we can set the environment variables using `ClimateControl` and we've made it optional here https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1521/files#diff-cb6e8126296dbc007e7625eee863de5c3213ed949b9999ea3418775f65826531R375-R377

thomasleese (commented):
Just got a few comments, it also looks like you've got a conflict with the schema in `main`, you might want to rebase to fix that.

app/controllers/staff/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Staff::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def azure_activedirectory_v2
+    auth = request.env["omniauth.auth"]
+    email = auth["info"]["email"]
+    provider = auth["provider"]
+    azure_ad_uid = auth["azure_ad_uid"]
+
+    @staff = Staff.find_by_email(email)
+
+    if @staff&.update(provider:, azure_ad_uid:)
```
thomasleese, line 10:
```suggestion
    if @staff&.update(azure_ad_uid:)
```

I think this will always fail as the `provider` is no longer there.

db/schema.rb
```diff
@@ -137,7 +137,7 @@
     t.integer "age_range_min"
     t.integer "age_range_max"
     t.text "subjects", default: [], null: false, array: true
-    t.datetime "recommended_at", precision: nil
+    t.datetime "recommended_at"
```
thomasleese, line 140: Looks like there are some unintended changes in this file, not entirely sure where they're coming from though. 🤔

richardpattinson: So clearly this is somethign to do with some configuration on my computer when i ran the migration; Im not sure how to undo the removals of precision: nil without rolling back my changes. Any idea how i can get my schema to be consistent with main?

app/controllers/staff/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,26 @@
+class Staff::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def azure_activedirectory_v2
+    auth = request.env["omniauth.auth"]
+    email = auth["info"]["email"]
+    azure_ad_uid = auth["azure_ad_uid"]
```
thomasleese, line 5:
```suggestion
    azure_ad_uid = auth["uid"]
```

Is this correct? In the original spike the key we get back from Omniauth is `uid`.

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1463/files#diff-ed2e5712248eb61bb5efb057efaa190974f0925392e36769f873e605a0d167d0R6

spec/system/support_interface/staff_spec.rb
```diff
@@ -10,6 +10,7 @@
   end
 
   it "allows inviting a user" do
```
thomasleese, line 12: Do we need a version of this test when Azure AD is disabled? In which case we would expect the user to need to provide a password so they can log in?

spec/system/support_interface/staff_spec.rb
```diff
@@ -27,11 +28,29 @@
...
+    and_i_fill_name
+    and_i_choose_support_console_permission
+    and_i_send_invitation
+    then_i_see_an_invitation_email
+    then_i_see_the_staff_index
+    then_i_see_the_invited_staff_user
+    then_i_sign_out
+
+    when_i_visit_the_invitation_email
+    then_i_am_taken_to_the_login_page
```
thomasleese, line 53:
```suggestion
    and_i_fill_password
    and_i_set_password
    then_i_see_the_staff_index

    then_i_see_the_accepted_staff_user
```

It looks like this is missing from this test ([taken from the one we've got currently](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/a17c21a393333c95f34195ae4d619d5408438fe4/spec/system/support_interface/staff_spec.rb#L29-L34)), we need to make sure that when a user is invited and we don't have Azure AD enabled that they can still successfully set a password and sign in.

thomasleese (approved):

PR #1530, opened 2023-06-30
Handle rolling back an automatic decline

thomasleese:
If an application is declined automatically because we haven't received a professional standing request, we may still want to be able to roll back the assessment.

[Trello Card](https://trello.com/c/qyDtUAXC/2029-update-rollback-service-to-handle-auto-decline-after-expired-requestable)


richardpattinson (approved):
Good changes all round

PR #1532, opened 2023-06-30
Check twice for duplicates in DQT

thomasleese:
Sometimes DQT won't find a duplicate the first time we do a search, but it seems to then be able to find the duplicate the next time (perhaps related to caching). While this is figured out, we can check twice for duplicates in DQT, 5 minutes after the first check.

[Trello Card](https://trello.com/c/v2FJ8kfJ/2030-dqt-duplicate-check-not-finding-duplicates)


richardpattinson (approved):
Looks great to me

PR #1535, opened 2023-07-03
Don't migrate uploads with errors

thomasleese:
Errors means that the file isn't in Azure anyway so there's nothing to migrate across, and saves us a little time in the process.


richardpattinson (approved):
Great work

PR #1537, opened 2023-07-04
Migrate dev DNS/AFD to s189 subscription

nickdevdfe:


Makefile
```diff
@@ -316,6 +315,9 @@ domains-apply: domains-init ## terraform apply for dns resources
 domains-destroy: domains-init ## terraform destroy for dns resources
 	terraform -chdir=terraform/custom_domains/environment_domains destroy -var-file workspace_variables/${DEPLOY_ENV}.tfvars.json
 
+domains-development:
+	$(eval DEPLOY_ENV=dev)
```
thomasleese, line 319: While we're here shall we add the other environments so they're ready to go?

thomasleese (approved):
Approved as it's definitely working for dev, and we can think about how to manage the domain environments better.

PR #1538, opened 2023-07-04
Remove Terraform workspace for PaaS dev environment

thomasleese:
We've switched the development environment over to AKS so we can remove the code related to the old PaaS environment.


nickdevdfe (approved):
Reviewed and approved.

PR #1544, opened 2023-07-06
Remove Terraform workspace for PaaS test environment

thomasleese:
We've switched the test environment over to AKS so we can remove the code related to the old PaaS environment.


nickdevdfe (approved):
reviewed and approved

PR #1578, opened 2023-07-19
Remove code related to the PaaS

thomasleese:
This removes code related to the PaaS as we've now completed the migration to AKS. I've removed the service unavailable page as they currently won't work, although we may need to bring them back in the future.

[Trello Card](https://trello.com/c/TjTM7Sxi/2087-switch-off-old-paas-infrastructure)


.github/workflows/lint.yml
```diff
@@ -52,13 +52,7 @@ jobs:
         run: terraform fmt -check -recursive
         working-directory: terraform
 
-      - name: Validate PaaS
-        run: |
-          terraform init -backend=false
-          terraform validate -no-color
-        working-directory: terraform/paas
-
-      - name: Validate AKS
```
richardpattinson, line 61: Why have this name change?

thomasleese: No particular reason other than we're only validating one thing now.

.gitignore
```diff
@@ -10,7 +10,6 @@
 .env.production.local
 .env.test.local
 app/assets/builds/*
-bin/fetch_config.rb
```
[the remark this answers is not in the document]
thomasleese: We no longer use this script, it's been removed from the `Makefile` so we don't need to ignore it as it won't exist.

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1578/commits/8a553e521c73172990a757081f370bf03d4ba634#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L102-L105

bin/restore-db
```diff
@@ -1,28 +0,0 @@
-#!/bin/bash
```
richardpattinson, line 1: What does this file do/ why do we no longer need it?

service_unavailable_page/web/mime.types
```diff
@@ -1,78 +0,0 @@
-types {
```
richardpattinson, line 1: Again; im unsure what this mime types file is doing; something to do with the service unavailable page?

service_unavailable_page/web/nginx.conf
```diff
@@ -1,35 +0,0 @@
-worker_processes 1;
```
richardpattinson, line 1: Same as above; what are these for?

bin/restore-db
thomasleese: It used to be able to restore the database of a PaaS environment, so it'll no longer work on AKS. We should have an alternative, but that's for a separate PR.

service_unavailable_page/web/public/internal/assets/images/govuk-mask-icon.svg
```diff
@@ -1,7 +0,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="132" height="97" viewBox="0 0 132 97" version="1.1">
```
richardpattinson, line 1: Why does the service unavailable page create so many files?

service_unavailable_page/web/mime.types
thomasleese: The service unavailable page was deployed to the PaaS, and therefore doesn't work on AKS. We probably do want an alternative for the future but I think it's confusing to keep code around which doesn't work, so until then we can remove it.

This particular file was a configuration file for Nginx, to make sure it served the files with the correct type for the browser.

service_unavailable_page/web/public/internal/index.html
```diff
@@ -1,167 +0,0 @@
...
-      Status - GOV.UK
-    </title>
-    <meta name="robots" content="noindex, nofollow" />
-    <meta
-      name="viewport"
-      content="width=device-width, initial-scale=1, viewport-fit=cover"
-    />
-    <meta name="theme-color" content="#0b0c0c" />
-    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-    <link
```
richardpattinson, line 16: Why does this page have so many images?

service_unavailable_page/web/public/internal/assets/images/govuk-mask-icon.svg
thomasleese: The service unavailable page was designed to be deployed without any dependency on our service, as the service would be down, so it needed its own copy of the GOV.UK Design system.

service_unavailable_page/web/public/internal/index.html
thomasleese: The service unavailable page was designed to be deployed without any dependency on our service, as the service would be down, so it needed its own copy of the GOV.UK Design system.

richardpattinson (commented):
Would like some clarity on some comments about files we are removing; Cannot see any issues otherwise. Once the comments are addressed I will approve

service_unavailable_page/web/public/internal/assets/images/govuk-mask-icon.svg
richardpattinson: Makes a lot of sense. cheers for the clarification

richardpattinson (approved):
Cheers for the clarification; This all seems like extraneous code now after the migration.

PR #1582, opened 2023-07-20
Remove third application form reminder

thomasleese:
We'd like to reduce the number of reminders we're sending applicants about unfinished applications as there is a concern that this will substantially increase the size of the backlog.

[Trello Card](https://trello.com/c/DqyQTmGm/2085-reduce-the-number-of-reminders-sent-out-for-draft-applications)


syed87 (approved):
That all looks good to me

PR #1586, opened 2023-07-21
Increase written statement expiration to 180 days

thomasleese:
We'd like to give applicants more time to receive their written statement as we're receiving a high number just after the deadline which is requiring us to rollback many applications.


syed87 (approved):
Looks Good To Me

PR #1591, opened 2023-07-24
Use Terrafile

thomasleese:
This switches us to using Terrafile as per the [teacher services cloud template](https://github.com/DFE-Digital/teacher-services-cloud/tree/main/templates/new_service). This means we can use the correct release tag of the Terraform modules according to each environment.

[Trello Card](https://trello.com/c/TjTM7Sxi/2087-switch-off-old-paas-infrastructure)


nickdevdfe (approved):
Reviewed and approved.

PR #1598, opened 2023-07-26
Switch to non-AKS environment references

thomasleese:
As we're no longer running two parallel environments (AKS and PaaS) we can rename the environment to remove the `_aks` suffix and stick with one consistent environment name.

[Trello Card](https://trello.com/c/uHFN31ny/2088-rename-github-aks-environments)


nickdevdfe (approved):
Tested, reviewed and approved.

PR #1609, opened 2023-08-01
Use development qualified teachers API in review

thomasleese:
The review environment is similar to the development environment except the apps are short lived, we should match up the environments so they correspond to the same one that the qualified teachers API is running in.


syed87 (approved):
looks good to me

PR #1630, opened 2023-08-09
Remove CountryCode#eligible_in_february_2023?

thomasleese:
We don't use this method anywhere so it's safe to be deleted.


syed87 (approved):
looks good to me

PR #1645, opened 2023-08-21
Rename manage applications permission

thomasleese:
Initially we thought this permission would get used across the board, but we now discovering that it's only used for one feature as different sets of users will need to use each feature.


richardpattinson (approved):
Nice job.

PR #1649, opened 2023-08-24
Improve content related to work experience

thomasleese:
This clarifies that the work experience doesn't have to have been in the UK, but it does need to have been after the teaching qualification.

[Trello Card](https://trello.com/c/TBnGHcDj/2204-add-content-about-teaching-experience-from-any-country-being-valid)


richardpattinson (approved):
Looks good to me

PR #1656, opened 2023-09-01
Only set work_history_checked attributes where necessary

thomasleese:
This attribute will only exist if the failure reason is able to be associated with a work history, so we should only set the attribute in that case.

This should fix: https://dfe-teacher-services.sentry.io/issues/4445860540/


thomasleese:
I've temporary disabled the required approvals to merge this in as it's causing issues for our users.

syed87 (commented):
looks good to me

PR #1660, opened 2023-09-04
Rename teaching authority information fields

thomasleese:
This renames the fields to be consistent with the `qualifications_information` field, and makes it clear that it's not specific to the teaching authority, more general information applicable for the region or country.

[Trello Card](https://trello.com/c/typ3Qxxo/2243-rename-teaching-authority-other-field)


richardpattinson (approved):
Looks good to me!

PR #1663, opened 2023-09-05
Remove preliminary check from region

thomasleese:
This removes the fields from regions as we only ever set this at a country-level, it should make the support console simpler and remove any confusion about where to set this.

[Trello Card](https://trello.com/c/vit18758/2245-remove-requires-preliminary-check-from-region)


thomasleese:
We've decided to switch so we'll remove it from the country table.

PR #1664, opened 2023-09-05
Remove preliminary check from country

thomasleese:
This removes the fields from countries and leaves it on regions, it should make the support console simpler and remove any confusion about where to set this.

[Trello card](https://trello.com/c/vit18758/2245-remove-requires-preliminary-check-from-region)


syed87 (approved):
looks good to me

PR #1669, opened 2023-09-08
Hide additional fields for countries with one region

thomasleese:
At the moment with countries with on region there are two places you can put this information, which is confusing. Instead we should hide it to encourage users to use the region.

I've also made a few changes to the pages to organise the fields and better match the designs.

[Trello Card](https://trello.com/c/Bbu2YRc6/2253-hide-duplicate-region-country-fields-if-there-is-only-one-region)


syed87 (approved):
looks good to me

PR #1676, opened 2023-09-12
2172 subject limited countries

syed87:
This adds a new subject_limited field to country to replace hard coded list. Countries can be added to list of subject limited countries by selecting new expanded option
Screenshot 2023-09-12 at 09 33 59

https://trello.com/c/azNtrW4F/2172-allow-customising-subject-limited-countries


app/controllers/support_interface/countries_controller.rb
```diff
@@ -61,6 +61,7 @@ def country_params
         :region_names,
         :sanction_information,
         :status_information,
+        :eligibility_route,
```
thomasleese, line 64: I think we can remove `eligibility_skip_questions` from this list as we're not passing it through to the controller anymore.

app/lib/country_code.rb
```diff
@@ -29,9 +29,8 @@ def european_economic_area?(code)
     end
 
     def secondary_education_teaching_qualification_required?(code)
```
thomasleese, line 31: I think we can remove this method now and replace the two places it's used with direct access to the `country`, to avoid an unnecessary database query in this method.

app/models/country.rb
```diff
@@ -34,8 +35,5 @@ class Country < ApplicationRecord
   CODES_ELIGIBLE_IN_FEBRUARY_2023 =
```
thomasleese, line 35: It looks like this is only used in the line below so can we get rid of it?

db/migrate/20230912081147_add_subject_limited_to_countries.rb
```diff
@@ -0,0 +1,5 @@
+class AddSubjectLimitedToCountries < ActiveRecord::Migration[7.0]
+  def change
+    add_column :countries, :subject_limited, :boolean, default: false
```
thomasleese, line 3:
```suggestion
    add_column :countries, :subject_limited, :boolean, default: false, null: false
```

I think we should make sure this field can't be `null`, as currently this field can have three values (`true`, `false` and `null`), but we only need to `true` or `false`.

db/seeds.rb
```diff
@@ -369,3 +369,12 @@
...
 end
+
+subject_limited_countries = %w[GH IN JM NG SG ZA ZW]
+
+subject_limited_countries.each do |country_code|
+  country = Country.find_by(code: country_code)
+  next unless country
+
+  country.update!(subject_limited: true)
+end
```
thomasleese, line 380:
```suggestion
Country.where(code: subject_limited_countries).update_all(subject_limited: true)
```

This avoids a database query for each country, and instead makes a single query that updates all the countries in one go.

spec/factories/countries.rb
```diff
@@ -22,16 +23,11 @@
     sequence :code, Country::CODES.cycle
 
     trait :requires_secondary_education_teaching_qualification do
```
thomasleese, line 25: I wonder if we could look at renaming these traits to match the new naming convention we're going with `subject_limited`.

spec/factories/countries.rb
```diff
@@ -22,16 +23,11 @@
...
+      subject_limited { true }
     end
 
     trait :doesnt_require_secondary_education_teaching_qualification do
-      sequence :code,
-               (
-                 Country::CODES -
-                   Country::CODES_REQUIRING_SECONDARY_EDUCATION_TEACHING_QUALIFICATION
-               ).cycle
+      subject_limited { true }
```
thomasleese, line 30: Should this be `false`? Also I wonder if we even need this trait since the default value for this field is `false`.

spec/lib/assessment_factory_spec.rb
```diff
@@ -282,6 +282,7 @@
       end
 
       describe "age range and subjects section" do
+        before { create(:country, code: "SG", subject_limited: true) }
```
thomasleese, line 285:
```suggestion
        before { create(:country, :requires_secondary_education_teaching_qualification, code: "SG") }
```

I think we should either use the trait, or remove the trait if we're not going to use it. If we rename the trait then we'll need to use the new name.

spec/system/support_interface/countries_spec.rb
```diff
@@ -59,18 +59,34 @@
   private
 
   def given_countries_exist
-    create(:region, :national, country: create(:country, code: "IE"))
-    create(:region, :national, country: create(:country, code: "PL"))
-    united_states = create(:country, code: "US")
+    create(
+      :region,
+      :national,
+      country: create(:country, code: "IE", subject_limited: false),
```
thomasleese, line 65:
```suggestion
      country: create(:country, code: "IE"),
```

Do we need to specify this field directly if it defaults to `false`?

thomasleese (commented):
Generally looks good! I've just got a few comments on the code.

app/lib/country_code.rb
syed87: I've made all the other changes suggested however I'm getting some errors with this I can't resolve.

app/models/application_form.rb
```diff
@@ -227,9 +227,7 @@ def english_language_exempt?
   end
 
   def secondary_education_teaching_qualification_required?
-    CountryCode.secondary_education_teaching_qualification_required?(
-      country.code,
-    )
+    country&.subject_limited || false
```
thomasleese, line 230:
```suggestion
    country.subject_limited
```

`country` won't be `nil` here so we don't need to access it safely.

db/migrate/20230921082209_change_subject_limited_to_not_null_in_countries.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeSubjectLimitedToNotNullInCountries < ActiveRecord::Migration[7.0]
+  def change
+    change_column_null :countries, :subject_limited, false
```
thomasleese, line 3: Rather than introducing a new migration, we can alter the existing one as it's not merged in to `main` yet and hasn't been run in any of our environments, that keeps the migrations clean and means there will never be a time where the column is `nullable`. It also means rolling back can happen in one step rather than two.

app/lib/country_code.rb
thomasleese: I was able to replace the `qualified_for_subject_required?` method with:

```rb
Country.exists?(code: country_code, subject_limited: true)
```

What errors were you getting?

thomasleese (commented):
Looks very close, just got a final couple of comments.

app/lib/country_code.rb
syed87: That's working for me without error now

app/forms/support_interface/country_form.rb
```diff
@@ -81,4 +87,31 @@ def diff_actions
...
+    elsif eligibility_skip_questions
+      "reduced"
+    else
+      "standard"
+    end
+  end
+
+  def eligibility_route=(value)
+    subject_limited_will_change!
+    eligibility_skip_questions_will_change!
```
thomasleese, line 103: Do we need these methods on this form? I know we needed it when the code was on the model directly, but I wonder if we need it on the form?

app/views/support_interface/countries/edit.html.erb
```diff
@@ -32,6 +33,11 @@
 
   <% if @country.regions.count > 1 %>
     <%= render "shared/support_interface/country_region_information_fields", f: %>
+  <% else %>
+    <%= f.hidden_field :qualifications_information, value: '' %>
+    <%= f.hidden_field :sanction_information, value: '' %>
+    <%= f.hidden_field :status_information, value: '' %>
+    <%= f.hidden_field :other_information, value: '' %>
```
thomasleese, line 40: We don't need these lines anymore as the form handles this now.

app/views/support_interface/countries/index.html.erb
```diff
@@ -16,11 +16,14 @@
               <%= govuk_tag(text: "Ineligible", colour: "red", classes: ["govuk-!-margin-left-2"]) %>
             <% end %>
 
-            <% if country.eligibility_skip_questions %>
-              <%= govuk_tag(text: "Reduced eligibility", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
-            <% else %>
-              <%= govuk_tag(text: "Standard eligibility", colour: "grey", classes: ["govuk-!-margin-left-2"]) %>
-            <% end %>
+            <% if country.subject_limited %>
+  <%= govuk_tag(text: "Subject limited", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
```
thomasleese, line 20:
```suggestion
  <%= govuk_tag(text: "Subject limited eligibility", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
```

To make it really clear what this means.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_09_12_144508) do
+ActiveRecord::Schema[7.0].define(version: 2023_09_21_082209) do
```
thomasleese, line 13:
```suggestion
ActiveRecord::Schema[7.0].define(version: 2023_09_21_081147) do
```

This is currently referring to the other migration that got deleted.

.DS_Store
thomasleese, line 1: I think this file can be deleted.

app/views/support_interface/countries/preview.html.erb
```diff
@@ -53,6 +53,19 @@
...
 
+<% if @country.eligibility_skip_questions_changed? || @country.subject_limited_changed? %>
+  <h3 class="govuk-heading-s">Which route will applicants take through the eligibility checker?</h3>
+  <p class="govuk-body">
+    <% if @country.subject_limited %>
+      <%= govuk_tag(text: "Subject limited", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
+    <% elsif @country.eligibility_skip_questions %>
+      <%= govuk_tag(text: "Reduced eligibility", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
+    <% else %>
+      <%= govuk_tag(text: "Standard eligibility", colour: "grey", classes: ["govuk-!-margin-left-2"]) %>
```
thomasleese, line 64:
```suggestion
      <%= govuk_tag(text: "Subject limited", colour: "yellow") %>
    <% elsif @country.eligibility_skip_questions %>
      <%= govuk_tag(text: "Reduced eligibility", colour: "yellow") %>
    <% else %>
      <%= govuk_tag(text: "Standard eligibility", colour: "grey") %>
```

I don't think the left margin is required here.

spec/models/eligibility_check_spec.rb
```diff
@@ -98,12 +98,14 @@
...
 
       context "when teach_children is false" do
+        before { create(:country, :subject_limited, code: "IN") }
         before { eligibility_check.teach_children = false }
 
         it { is_expected.to include(:teach_children_secondary) }
       end
 
       context "when qualified_for_subject is false" do
+        before { create(:country, :subject_limited, code: "IN") }
```
thomasleese, line 108: To avoid duplication, this line can go in the `"when filtering by subject"` context, and then it'll apply to both the `"when teach_children is false"` and `"when qualified_for_subject is false"` context.

thomasleese (commented):
Looks pretty good, just some small comments left.

db/schema.rb
syed87: switched to 2023_09_12_081147

app/forms/support_interface/country_form.rb
syed87: Doesn't seem to be needed. Everything works without these methods

db/schema.rb
thomasleese: Ah, I think I got this wrong actually, I mistook the date there for the other migration but actually it's one that's ready in main, sorry! You can revert this change back to how it was.

thomasleese: I think it should be `2023_09_12_144508` actually as that's the latest migration in `main` (the migration in this file comes earlier than that one).

app/views/support_interface/countries/index.html.erb
```diff
@@ -16,11 +16,14 @@
...
-            <% else %>
-              <%= govuk_tag(text: "Standard eligibility", colour: "grey", classes: ["govuk-!-margin-left-2"]) %>
-            <% end %>
+            <% if country.subject_limited %>
+  <%= govuk_tag(text: "Subject limited eligibility", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
+<% elsif country.eligibility_skip_questions %>
+  <%= govuk_tag(text: "Reduced eligibility", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
+<% else %>
+  <%= govuk_tag(text: "Standard eligibility", colour: "grey", classes: ["govuk-!-margin-left-2"]) %>
+<% end %>
```
thomasleese, line 25:
```suggestion
            <% if country.subject_limited %>
              <%= govuk_tag(text: "Expanded eligibility", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
            <% elsif country.eligibility_skip_questions %>
              <%= govuk_tag(text: "Reduced eligibility", colour: "yellow", classes: ["govuk-!-margin-left-2"]) %>
            <% else %>
              <%= govuk_tag(text: "Standard eligibility", colour: "grey", classes: ["govuk-!-margin-left-2"]) %>
            <% end %>
```

I think the indentation is wrong here, but also it occurs to me that we should probably match the naming of the question.

app/views/support_interface/countries/preview.html.erb
```diff
@@ -53,6 +53,19 @@
...
 
+<% if @country.eligibility_skip_questions_changed? || @country.subject_limited_changed? %>
+  <h3 class="govuk-heading-s">Which route will applicants take through the eligibility checker?</h3>
+  <p class="govuk-body">
+    <% if @country.subject_limited %>
+      <%= govuk_tag(text: "Subject limited", colour: "yellow") %>
+    <% elsif @country.eligibility_skip_questions %>
+      <%= govuk_tag(text: "Reduced eligibility", colour: "yellow") %>
+    <% else %>
+      <%= govuk_tag(text: "Standard eligibility", colour: "grey") %>
```
thomasleese, line 64:
```suggestion
      <%= govuk_tag(text: "Expanded", colour: "yellow") %>
    <% elsif @country.eligibility_skip_questions %>
      <%= govuk_tag(text: "Reduced", colour: "yellow") %>
    <% else %>
      <%= govuk_tag(text: "Standard", colour: "grey") %>
```

thomasleese (commented):
Just some very small comments related to the content, but otherwise looks good. I think it's worth squashing the commits together now too.

app/views/support_interface/countries/preview.html.erb
```diff
@@ -53,6 +53,19 @@
   <%= raw GovukMarkdown.render(@country.qualifications_information) %>
 <% end %>
 
+<% if @country.eligibility_skip_questions_changed? || @country.subject_limited_changed? %>
+  <h3 class="govuk-heading-s">Which route will applicants take through the eligibility checker?</h3>
+  <p class="govuk-body">
+    <% if @country.subject_limited %>
+      <%= govuk_tag(text: "Expanded", colour: "yellow") %>
+    <% elsif @country.eligibility_skip_questions %>
+      <%= govuk_tag(text: "Reduced eligibility", colour: "yellow") %>
```
thomasleese, line 62:
```suggestion
      <%= govuk_tag(text: "Reduced", colour: "yellow") %>
```

app/views/support_interface/countries/preview.html.erb
```diff
@@ -53,6 +53,19 @@
...
+    <% else %>
+      <%= govuk_tag(text: "Standard eligibility", colour: "grey") %>
```
thomasleese, line 64:
```suggestion
      <%= govuk_tag(text: "Standard", colour: "grey") %>
```

thomasleese (commented):
Just some final content comments to be consistent on the wording.

thomasleese (approved):

PR #1677, opened 2023-09-12
Reorder support console fields

thomasleese:
This moves around a number of fields so it matches the design. There shouldn't be any functional changes.


syed87 (approved):
looks good to me

PR #1680, opened 2023-09-13
Hide university degree hint text

thomasleese:
If it's a university degree we don't require validation, so we can hide the hint text.

[Trello Card](https://trello.com/c/nhgELHUG/2266-incorrect-hint-text-for-university-degree)


syed87 (approved):
looks good to me

PR #1681, opened 2023-09-13
Update award email and page content

thomasleese:
This updates the content of the award email and the page teachers see after they sign in, to clarify some of the next steps and to include a link to the access your teaching qualifications service.

Depends on #1679.


syed87 (approved):
looks good to me

PR #1686, opened 2023-09-13
Don't set started at on preliminary assessment sections

thomasleese:
This doesn't count as part of the assessment, so we shouldn't be recording the started at date of the assessment as part of the preliminary check.


syed87 (approved):
looks good to me

PR #1688, opened 2023-09-13
Only change to assessment in progress if not preliminary

thomasleese:
This changes the logic so applications will only move to assessment in progress if there are non-preliminary sections finished.


thomasleese:
Merging to fix a bug.

PR #1671, opened 2023-09-08
Add action required by workflow and filter

thomasleese:
This adds a new field on application forms (`actionable_by`) and implements the logic for when the value of this field changes. See individual commits for more details.

[Trello Card](https://trello.com/c/XN5Omc6e/2264-add-flagging-actionable-applications)

## Screenshots

Screenshot 2023-09-08 at 14 57 40
Screenshot 2023-09-08 at 15 27 34

PR #1689, opened 2023-09-14
Add application form stage

thomasleese:
This adds the new stage field to application forms which we'll be using in the future to allow users to filter on.

Depends on #1671.

[Trello Card](https://trello.com/c/9sZ5E9w8/2265-update-statuses-for-verification-and-review)


syed87 (approved):
looks good to me

PR #1703, opened 2023-09-19
Remove old application form status fields 

thomasleese:
This removes the field (and related fields) from the application form model as they've been replaced by others and are no longer being used anywhere.

Depends on #1702 

[Trello Card](https://trello.com/c/9sZ5E9w8/2265-update-statuses-for-verification-and-review)

PR #1707, opened 2023-09-20
Stop using old_state and new_state

thomasleese:
We're going to replace them with `old_value` and `new_value` to reduce the number of fields on the timeline event model.

Depends on #1703

[Trello Card](https://trello.com/c/vBBKmKoB/835-stop-using-oldstate-newstate-in-timeline-events)


syed87 (approved):
looks good to me

PR #1715, opened 2023-09-24
Rename state_changed to status_changed

thomasleese:
We're going to rename this to `status_changed` to better represent how we show it to users.

Depends on #1708 

[Trello Card](https://trello.com/c/vBBKmKoB/835-stop-using-oldstate-newstate-in-timeline-events)


syed87 (approved):
looks good to me

PR #1717, opened 2023-09-26
Track verification dates

thomasleese:
This adds new fields to the requestable models for tracking when they're requested and expired. This is because with the new verification flow we need this information to know when a requestable has been requested (as in the new flow, requestables can be created before they're requested).

[Trello Card](https://trello.com/c/kJ7RrYLn/2296-mark-lops-verification-to-be-done)


syed87 (approved):
looks good to me

PR #1720, opened 2023-09-26
Remove requestable state

thomasleese:
This removes the column from all the requestables as we're no longer using it for anything and it can be safely removed.

Depends on #1719 

[Trello Card](https://trello.com/c/kJ7RrYLn/2296-mark-lops-verification-to-be-done)


syed87 (approved):
looks good to me

PR #1727, opened 2023-09-29
2262 add verification permission

richardpattinson:
We need a new permission for staff users which captures whether they’ve got access to do new verification journey.

Added a migration to store the new permission on the staff model, and update the support console/example data to support this new permission.

### DIFF
--- a/app/views/personas/index.html.erb
+++ b/app/views/personas/index.html.erb
@@ -11,7 +11,7 @@
     <%= govuk_table do |table|
       table.with_colgroup do |colgroup|
         colgroup.with_col(span: 1)
-        colgroup.with_col(span: 6)
+        colgroup.with_col(span: 7)
         colgroup.with_col(span: 1)
       end
 
@@ -29,6 +29,7 @@
           row.with_cell(header: true, text: "Change work history")
           row.with_cell(header: true, text: "Reverse decisions")
           row.with_cell(header: true, text: "Support console")
+          row.with_cell(header: true, text: "Verify applicants")
           row.with_cell(header: true, text: "Withdraw")
           row.with_cell(header: true, text: "Actions", numeric: true)
         end
@@ -44,6 +45,7 @@
             row.with_cell { govuk_boolean_tag(staff.change_work_history_permission) }
             row.with_cell { govuk_boolean_tag(staff.reverse_decision_permission) }
             row.with_cell { govuk_boolean_tag(staff.support_console_permission) }
+            row.with_cell { govuk_boolean_tag(staff.verify_permission) }
             row.with_cell { govuk_boolean_tag(staff.withdraw_permission) }
 
             row.with_cell(numeric: true) do
--- a/app/views/personas/index.html.erb
+++ b/app/views/personas/index.html.erb
@@ -18,7 +18,7 @@
       table.with_head do |head|
         head.with_row do |row|
           row.with_cell(scope: false)
-          row.with_cell(header: true, text: "Permissions", colspan: 5)
+          row.with_cell(header: true, text: "Permissions", colspan: 7)
           row.with_cell(scope: false)
         end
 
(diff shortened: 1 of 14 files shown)


lib/tasks/example_data.rake
```diff
@@ -90,6 +93,7 @@ def staff_members
       change_work_history_permission: false,
       reverse_decision_permission: false,
       support_console_permission: false,
+      verification_permission: false,
```
thomasleese, line 96: Could we add a new "admin" persona who does have permission to perform verification, but nothing else?

app/views/personas/index.html.erb
```diff
@@ -29,6 +29,7 @@
           row.with_cell(header: true, text: "Change work history")
           row.with_cell(header: true, text: "Reverse decisions")
           row.with_cell(header: true, text: "Support console")
+          row.with_cell(header: true, text: "Verify Permissions")
```
thomasleese, line 32:
```suggestion
          row.with_cell(header: true, text: "Verify applicants")
```

Since that's a better description of what the permission gives you access to do, and we're already within a cell group titled "Permissions". Also, I think we'll need to increase the `colspan`s above to include this column.

app/views/personas/index.html.erb
```diff
@@ -45,6 +46,7 @@
             row.with_cell { govuk_boolean_tag(staff.reverse_decision_permission) }
             row.with_cell { govuk_boolean_tag(staff.support_console_permission) }
             row.with_cell { govuk_boolean_tag(staff.withdraw_permission) }
+            row.with_cell { govuk_boolean_tag(staff.withdraw_permission) }
```
thomasleese, line 49:
```suggestion
            row.with_cell { govuk_boolean_tag(staff.verification_permission) }
```

Could we also put this before the `withdraw_permission` to maintain the alphabetical ordering?

db/migrate/20230929084821_add_verification_permission_to_staff_users.rb
```diff
@@ -0,0 +1,9 @@
+class AddVerificationPermissionToStaffUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :staff,
+               :verification_permission,
```
thomasleese, line 4: I'm wondering about the name here, the others are named as an action that can be taken if the user has the write permission (e.g. they can _withdrawn_ or they can _reverse decision_).

Perhaps `verify_permission` would better fit with the existing conventions?

thomasleese (commented):
Functionality looks good, I've just got a few small comments on the naming.

db/migrate/20230929084821_add_verification_permission_to_staff_users.rb
richardpattinson: Should be addressed now =)

app/views/personas/index.html.erb
richardpattinson: Done =)

app/views/personas/index.html.erb
```diff
@@ -11,7 +11,7 @@
     <%= govuk_table do |table|
       table.with_colgroup do |colgroup|
         colgroup.with_col(span: 1)
-        colgroup.with_col(span: 6)
+        colgroup.with_col(span: 7)
```
thomasleese, line 14: I think we also need to change the `colspan` below on line 21.

app/views/personas/index.html.erb
```diff
@@ -11,14 +11,14 @@
...
         colgroup.with_col(span: 1)
       end
 
       table.with_head do |head|
         head.with_row do |row|
           row.with_cell(scope: false)
-          row.with_cell(header: true, text: "Permissions", colspan: 5)
+          row.with_cell(header: true, text: "Permissions", colspan: 6)
```
thomasleese, line 21:
```suggestion
          row.with_cell(header: true, text: "Permissions", colspan: 7)
```

I actually think this needs to be 7 because there are 7 permission columns (I think the value of 5 before was incorrect).

thomasleese (approved):

PR #1733, opened 2023-10-04
Make requestables verifiable

thomasleese:
This updates the requestables to allow them to be verified by an admin user by adding the necessary database fields and service class.

Depends on #1732 

[Trello Card](https://trello.com/c/8sKZ1aEm/2309-lops-response-review)


richardpattinson (approved):
Looks great; no comments

PR #1738, opened 2023-10-06
Fix reference to passed field

thomasleese:
The field no longer exists and has been replaced with `review_passed?`.

This should fix: https://dfe-teacher-services.sentry.io/issues/4529235517/?referrer=slack&notification_uuid=0d34115a-51a4-4db7-a5cc-f1b570de60a6&alert_rule_id=11189458&alert_type=issue


thomasleese:
I'm going to merge this as it's causing issues in production.

PR #1739, opened 2023-10-06
Fix reference to failed method

thomasleese:
This has been replaced by `review_failed?`.


thomasleese:
I'm going to merge this as it's causing issues with production.

PR #1741, opened 2023-10-06
Fix logic when to trigger a received reference

thomasleese:
This logic no longer works after we deployed https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1719 as there are no longer three distinct states for a requestable. Instead, I've fixed the logic and made it clearer exactly what it's doing.


richardpattinson (approved):
Makes sense to me

PR #1744, opened 2023-10-06
Allow admins to make recommendations

thomasleese:
This creates a policy which allows users with the verify permission (as well as the existing award/decline permission) to make recommendations on the application. This is required for the new verification journey that allows admins to go on and award applications following assessment and verification.


syed87 (approved):
looks good to me

PR #1751, opened 2023-10-10
Rename award/decline permission

thomasleese:
This renames the permission as it does more than just allow awards and declines, and this works better with the new `verify_permission` field.

[Trello Card](https://trello.com/c/MToXbrNQ/2316-rename-award-decline-permission)


syed87 (approved):
looks good to me

PR #1756, opened 2023-10-11
2320 refactor page objects to add a namespace prefix

richardpattinson:
We use page objects in our system tests (https://github.com/DFE-Digital/apply-for-qualified-teacher-status/tree/main/spec/support/autoload/page_objects ) and they’re all in modules according to which part of the service they refer to.

We access these helpers from a file which constructs them, but in that file we don’t use the module names in the naming of the methods: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/spec/support/page_helpers.rb

This has lead to a number of confusingly named methods and it means we’ve started adding prefixes to some, but not all of them, and in some cases it’s hard to know which page is being tested.

We should refactor this file to make sure every method is prefixed with an appropriate namespace.


thomasleese (commented):
Looks good, I think we just need to make sure the `page_helpers.rb` file is sorted alphabetically correctly.

thomasleese (approved):

PR #1757, opened 2023-10-11
Remove ready for review field

thomasleese:
This removes a field that we're no longer using anywhere.

Depends on #1722


richardpattinson (approved):

syed87 (approved):
looks good to me

PR #1769, opened 2023-10-18
Handle navigating past the country page after selecting an ineligible…

richardpattinson:
… country

https://dfe-teacher-services.sentry.io/issues/3910860876/?project=6426061&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=2

We’ve got reports of this issue in Sentry, only happened 29 times and only seems to happen every few weeks, but we should investigate and fix the issue.

What happens if eligibility_check.status is something that_ isn’t _in paths.keys ?[https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/8cd32f1b31c[…]/app/controllers/concerns/enforce_eligibility_question_order.rb](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/8cd32f1b31cb5c16a54c8788462cf6a024b9bb45/app/controllers/concerns/enforce_eligibility_question_order.rb#L16-L32)

Because there is one case where the region can be nil and the status will be something not in the list:[https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/8cd32f1b31cb5c16a54c8788462cf6a024b9bb45/app/models/eligibility_check.rb#L145](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/8cd32f1b31cb5c16a54c8788462cf6a024b9bb45/app/models/eligibility_check.rb#L145 "‌")

Reproducing the issue pre fix: 

The steps are:

- Start the eligibility checker

- Choose an ineligible country (e.g. North Korea)

- Ignore the ineligible result and go to the /eligibility/work-experience page

In that case we’ve got an eligibility check where region is nil , the status is :eligibility so it passes the order enforcement logic

The solution I've implemented is to check that the applicant is ineligible and redirect to the ineligible page if so. The applicant can still navigate to the country page.


syed87 (approved):
looks good to me

app/controllers/concerns/enforce_eligibility_question_order.rb
```diff
@@ -18,6 +22,11 @@ def current_page_is_allowed?
     current_position && order && current_position <= order
   end
 
+  def ineligible_application_without_skip_additional_questions
+    eligibility_check&.status == :eligibility &&
```
thomasleese, line 26:
```suggestion
    eligibility_check&.status == :eligibility
```

The status of `:eligibility` includes both eligible and ineligible checks - if the check is eligible I don't think we want to redirect to the ineligible page.

app/controllers/concerns/enforce_eligibility_question_order.rb
```diff
@@ -4,6 +4,10 @@ module EnforceEligibilityQuestionOrder
   included { before_action :redirect_by_status }
 
   def redirect_by_status
+    if request.path != eligibility_interface_countries_path &&
+         ineligible_application_without_skip_additional_questions
+      redirect_to eligibility_interface_ineligible_path
+    end
```
thomasleese, line 10: I'm not sure if this is the right place to put this logic as this class shouldn't be concerned about the specifics of skipping additional questions, that's already handled by the `status` method so we don't want to be duplicating ourselves here (https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/13d4dd3201ad687d64ed5e19ab41f6269acb952a/app/models/eligibility_check.rb#L150-L160).

We also don't want to be just checking for this case when we're not on the countries path. Although at the moment this is only place we can get this sort of out-of-order issue, we should build this class in a generic way that allows it to work for any potential issues in the future.

The logic as written here also introduces a bug where users will get redirected to the ineligible page if they're eligible, skips additional questions and not on the countries path.

thomasleese (changes requested):

thomasleese (approved):

PR #1770, opened 2023-10-23
Changes made to titles and Management integrated into task list

syed87:
https://trello.com/c/znCe7eOH/2342-rename-application-overview-assessment-title

Assessment header changed to Tasklist
Management tasks change to Management
Management links integrated into task list

Screenshot 2023-10-23 at 03 50 16

### DIFF
--- a/app/view_objects/assessor_interface/application_forms_show_view_object.rb
+++ b/app/view_objects/assessor_interface/application_forms_show_view_object.rb
@@ -59,8 +59,8 @@ class AssessorInterface::ApplicationFormsShowViewObject
            assessment,
          ).rollback?
         {
-          title: "Reverse decision",
-          path: [:rollback, :assessor_interface, application_form, assessment],
+          name: "Reverse decision",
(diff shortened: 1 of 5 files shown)


app/view_objects/assessor_interface/application_forms_show_view_object.rb
```diff
@@ -90,6 +91,18 @@ def management_tasks
            to: :assessment
   delegate :canonical_email, to: :teacher
 
+  def management_task_list_section
+    return nil if management_tasks.blank?
+
+    {
+      title: "Management",
+      items:
+        management_tasks.map do |task|
```
thomasleese, line 100: Since `management_tasks` is now only being used in one place (here), we can change that method to return the items directly (an array of hashes with a `name` and a `link`) which removes the need to transform it here with `map`.

thomasleese (commented):
Looks good, just one small comment to make the code neater.

thomasleese (approved):

thomasleese (approved):

PR #1778, opened 2023-10-31
2322-send-referee-reminders-to-applicants

richardpattinson:
When reference request reminder emails are sent to referees the associated teacher will be notified. They will be contacted when their referees have 28 and 14 days to respond.

### DIFF
--- a/app/models/concerns/expirable.rb
+++ b/app/models/concerns/expirable.rb
@@ -9,6 +9,11 @@ module Expirable
     requested_at + expires_after
   end
 
+  def days_until_expired
+    return nil if expires_at.nil?
+    (expires_at.to_date - Time.zone.today).to_i
+  end
+
   def expired!
     update!(expired_at: Time.zone.now)
   end
--- a/app/views/teacher_mailer/references_reminder.text.erb
+++ b/app/views/teacher_mailer/references_reminder.text.erb
@@ -1,0 +1,17 @@
+Dear <%= application_form_full_name(@application_form) %>
+
+## We’re still waiting for a response from one or more of the references you provided to verify your work history.
+
+<% if @number_of_reminders_sent == 0 %>
+We contacted your references on <%= @reference_requests.first.requested_at.to_date.to_fs(:long_ordinal_uk) %> but have not received a response from:
+<% elsif @number_of_reminders_sent == 1 %>
+The following references have just 2 weeks to respond to the reference request.
+<% end %>
+
+<% @reference_requests.each do |reference_request| %>
+- <%= reference_request.work_history.contact_name %> — <%= reference_request.work_history.school_name %>
+<% end %>
+
+They need to respond by <%= @reference_requests.first.expires_at.to_date.to_fs(:long_ordinal_uk) %>. If your references do not respond, and, as a result, we cannot verify your work history, we may not be able to award you QTS. You should contact your reference to remind them about the request.
+
+<%= render "shared/teacher_mailer/footer" %>
--- a/spec/mailers/teacher_mailer_spec.rb
+++ b/spec/mailers/teacher_mailer_spec.rb
@@ -442,15 +442,18 @@ RSpec.describe TeacherMailer, type: :mailer do
   end
 
   describe "#references_requested" do
-    subject(:mail) { described_class.with(teacher:).references_requested }
+    subject(:mail) do
+      described_class.with(
+        teacher:,
+        reference_requests: [create(:reference_request, :requested)],
+      ).references_requested
+    end
 
     describe "#subject" do
       subject(:subject) { mail.subject }
 
       it do
-        is_expected.to eq(
-          "Your qualified teacher status application – we’ve contacted your references",
-        )
+        is_expected.to eq("We’ve contacted your references – QTS application")
       end
     end
 
@@ -466,7 +469,7 @@ RSpec.describe TeacherMailer, type: :mailer do
       it { is_expected.to include("Dear First Last") }
       it do
         is_expected.to include(
-          "We’ve contacted the references you provided to verify your work history",
+          "We’ve contacted the references you provided to verify the work history",
         )
       end
     end
--- a/spec/mailers/teacher_mailer_spec.rb
+++ b/spec/mailers/teacher_mailer_spec.rb
@@ -366,6 +366,81 @@ RSpec.describe TeacherMailer, type: :mailer do
     it_behaves_like "an observable mailer", "professional_standing_received"
   end
 
+  describe "#references_reminder" do
+    let(:reference_request) do
+      create(
+        :reference_request,
+        requested_at: Date.new(2020, 1, 1),
+        work_history:
+          create(
+            :work_history,
+            school_name: "St Smith School",
+            contact_name: "John Smith",
+          ),
+        assessment:,
+      )
+    end
+    let(:number_of_reminders_sent) { nil }
+    subject(:mail) do
+      described_class.with(
+        teacher:,
+        number_of_reminders_sent:,
+        reference_requests: [reference_request],
+      ).references_reminder
+    end
+
+    describe "#subject" do
+      subject(:subject) { mail.subject }
+
+      context "with no reminder emails" do
+        let(:number_of_reminders_sent) { 0 }
+        it { is_expected.to eq("Waiting on references – QTS application") }
+      end
+
+      context "with one reminder email" do
+        let(:number_of_reminders_sent) { 1 }
+        it do
+          is_expected.to eq(
+            "Your references only have two weeks left to respond",
+          )
+        end
+      end
+    end
+
+    describe "#to" do
+      subject(:to) { mail.to }
+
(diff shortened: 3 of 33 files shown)


app/views/teacher_mailer/references_reminder.text.erb
```diff
@@ -0,0 +1,19 @@
...
+<% elsif @referees_to_be_reminded.first.reminder_emails.count == 2 %>
+  The following references have just 2 weeks to respond to the reference request.
+<% end %>
+
+<% @referees_to_be_reminded.each do |reference_request|  %>
+    <%= reference_request.contact_name + ". " + reference_request.work_history.school_name %>
+<% end %>
+
+
+They need to respond by <%= @application_form.work_histories.first.reference_request.expires_at.to_date.to_fs(:long_ordinal_uk) %>. If your references do not respond, and, as a result, we cannot verify your work history, we may not be able to award you QTS. You should contact your reference to remind them about the request.
```
thomasleese, line 17: `@application_form.work_histories.first.reference_request` means we'll always show the value for the first work history, even if we're not sending an email about the first work history.

app/services/send_reference_request_reminder_to_teacher.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+#
+class SendReferenceRequestReminderToTeacher
```
thomasleese, line 3: Rather than introducing a new service class and job for sending these emails, can we instead send the email at the same point that we send the reminder to the referees?

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/13d4dd3201ad687d64ed5e19ab41f6269acb952a/app/models/reference_request.rb#L101-L103

thomasleese: Talked in person and we need this to send one email together.

app/models/teacher.rb
```diff
@@ -45,4 +46,28 @@ def send_magic_link(*)
...
+        reference_request.should_send_reminder_email?(
+          (reference_request.expires_at.to_date - Time.zone.today).to_i,
+          reminder_emails.count,
+        )
+      end
+    Rails.logger.debug @remindable_reference_requests
+    @remindable_reference_requests.any?
+  end
+
+  def send_reminder_email(_number_of_reminders_sent)
```
thomasleese, line 67: We've got `number_of_reminders_sent` here available so we can pass it as a parameter and then use it in the view, this avoids issues around delays in email jobs and the database changing.

app/views/teacher_mailer/references_reminder.text.erb
```diff
@@ -0,0 +1,18 @@
...
+We’re still waiting for a response from one or more of the references you provided to verify your work history.
+
+<% if @teacher.reminder_emails.count == 0 %>
+  We contacted your references on <%= @application_form.work_histories.first.reference_request.requested_at.to_date.to_fs(:long_ordinal_uk) %> but have not received a response from:
+<% elsif @teacher.reminder_emails.count == 1 %>
+  The following references have just 2 weeks to respond to the reference request.
+<% end %>
+
+<% @reference_requests.each do |reference_request|  %>
+    <%= reference_request.contact_name + ". " + reference_request.work_history.school_name %>
```
thomasleese, line 12:
```suggestion
    <%= reference_request.contact_name %>. %<= reference_request.work_history.school_name %>
```

app/views/teacher_mailer/references_reminder.text.erb
```diff
@@ -0,0 +1,18 @@
+Dear <%= application_form_full_name(@application_form) %>
+
+We’re still waiting for a response from one or more of the references you provided to verify your work history.
+
+<% if @teacher.reminder_emails.count == 0 %>
```
thomasleese, line 5: It looks like we define a `@number_of_reminders_sent` variable in the mailer but then don't use it in the view.

spec/mailers/teacher_mailer_spec.rb
```diff
@@ -384,37 +384,4 @@
 
     it_behaves_like "an observable mailer", "professional_standing_received"
   end
-
-  describe "#references_requested" do
```
thomasleese, line 388: Why are we removing these tests?

app/models/concerns/expirable.rb
```diff
@@ -4,6 +4,7 @@ module Expirable
   extend ActiveSupport::Concern
 
   def expires_at
+    return true if is_a?(Teacher)
```
thomasleese, line 7: Is there a way we can return `nil` here, changing the return type from an optional date to an optional date or a boolean doesn't quite feel right.

app/models/teacher.rb
```diff
@@ -45,4 +46,28 @@ def send_magic_link(*)
...
+      ReferenceRequest
+        .joins(:work_history)
+        .where(work_histories: { application_form_id: application_form.id })
+        .where(received_at: nil, expired_at: nil)
+    @remindable_reference_requests =
+      all_reference_requests.select do |reference_request|
+        reference_request.should_send_reminder_email?(
+          (reference_request.expires_at.to_date - Time.zone.today).to_i,
+          reminder_emails.count,
+        )
```
thomasleese, line 61: I think we should move this logic to the `remindable` concern, since it has access to all these things, and it might allow us to have a single method `should_send_reminder_email?` for both teachers and requestables.

For example:

```rb
def days_until_expired
  (reference_request.expires_at.to_date - Time.zone.today).to_i
end
```

app/models/concerns/expirable.rb
```diff
@@ -4,6 +4,7 @@ module Expirable
   extend ActiveSupport::Concern
 
   def expires_at
+    return nil if is_a?(Teacher)
```
thomasleese, line 7: Nice that this works, could we then add to the `Teacher` class:

```rb
def expires_at
  nil
end
```

To overwrite this method and then we don't need a special case in the concern.

app/models/teacher.rb
```diff
@@ -45,4 +46,27 @@ def send_magic_link(*)
...
+        .joins(:work_history)
+        .where(work_histories: { application_form_id: application_form.id })
+        .where(received_at: nil, expired_at: nil)
+    @remindable_reference_requests =
+      all_reference_requests.select do |reference_request|
+        reference_request.should_send_reminder_email?(
+          reference_request.days_until_expired,
+          reminder_emails.count,
+        )
+      end
```
thomasleese, line 62: Since we're getting everything from the `reference_request` now, we could change the method in the concern to not take any arguments. An example of how to replace the method in `ReferenceRequest`:

```rb
def should_send_reminder_email?
  (days_until_expired <= 28 && reminder_emails.count.zero?) ||
      (days_until_expired <= 14 && reminder_emails.count == 1)
end
```

This would allow us to remove the special case in the `SendReminderEmail` service, as both teachers and requestables would have a version of this method without arguments.

app/views/teacher_mailer/references_reminder.text.erb
```diff
@@ -0,0 +1,18 @@
...
+We’re still waiting for a response from one or more of the references you provided to verify your work history.
+
+<% if @number_of_reminders_sent == 0 %>
+  We contacted your references on <%= @reference_requests.first.requested_at.to_date.to_fs(:long_ordinal_uk) %> but have not received a response from:
+<% elsif @number_of_reminders_sent == 1 %>
+  The following references have just 2 weeks to respond to the reference request.
+<% end %>
+
+<% @reference_requests.each do |reference_request|  %>
+    <%= reference_request.contact_name %> . %<= reference_request.work_history.school_name %>
```
thomasleese, line 12:
```suggestion
  <%= reference_request.contact_name %> . <%= reference_request.work_history.school_name %>
```

spec/mailers/teacher_mailer_spec.rb
```diff
@@ -417,4 +417,56 @@
...
+
+    describe "#body" do
+      subject(:body) { mail.body.encoded }
+
+      it { is_expected.to include("Dear First Last") }
+      it do
+        is_expected.to include(
+          "We’re still waiting for a response from one or more of the references you provided to verify your work history."
+        )
+      end
```
thomasleese, line 467: Could we also add some tests to check the content related to the individual reference requests also comes through correctly?

app/models/teacher.rb
thomasleese:
```rb
TeacherMailer
      .with(teacher: self, reference_requests: @remindable_reference_requests, number_of_reminders_sent:)
```

and then in the mailer:

```rb
@number_of_reminders_sent = params[:number_of_reminders_sent]
```

spec/mailers/teacher_mailer_spec.rb
thomasleese: I'm thinking something that would have caught https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/1778#discussion_r1389159635

thomasleese (approved):

PR #1785, opened 2023-11-03
2366 create link with dynamic name

syed87:
LoPS link on Review verifications page has been made dynamic to pull through the name of the competent authority

https://trello.com/c/FfuC3FAK/2366-create-link-with-dynamic-name

### DIFF
--- a/app/controllers/assessor_interface/review_verifications_controller.rb
+++ b/app/controllers/assessor_interface/review_verifications_controller.rb
@@ -13,11 +13,11 @@ module AssessorInterface
 
       @region = @assessment.application_form.region
 
-      if @region.teaching_authority_name.present?
-        @teaching_authority_name = @region.teaching_authority_name
-      else
-        @teaching_authority_name = "Contact relevant competent authority"
-      end      
+      @teaching_authority_name =
+        (
+          @region.teaching_authority_name.presence ||
+            "Contact relevant competent authority"
+        )
 
       render layout: "full_from_desktop"
     end
--- a/app/controllers/assessor_interface/review_verifications_controller.rb
+++ b/app/controllers/assessor_interface/review_verifications_controller.rb
@@ -11,6 +11,14 @@ module AssessorInterface
       @professional_standing_request =
         assessment.professional_standing_request if assessment.professional_standing_request.verify_failed?
 
+      @region = @assessment.application_form.region
+
+      if @region.teaching_authority_name.present?
+        @teaching_authority_name = @region.teaching_authority_name
+      else
+        @teaching_authority_name = "Contact relevant competent authority"
+      end      
+
       render layout: "full_from_desktop"
     end
 
--- a/app/views/assessor_interface/review_verifications/index.html.erb
+++ b/app/views/assessor_interface/review_verifications/index.html.erb
@@ -7,17 +7,20 @@
   The following verification tasks have been flagged for review:
 </p>
 
-<%= govuk_table do |table|
-  if @professional_standing_request.present?
-    table.with_caption(text: "LoPS")
-    table.with_body do |body|
-      body.with_row do |row|
-        row.with_cell { govuk_link_to region_certificate_name(@application_form.region), [:review, :assessor_interface, @application_form, @assessment, :professional_standing_request] }
-        row.with_cell { render(StatusTag::Component.new(@professional_standing_request.review_status)) }
-      end
-    end
-  end
-end %>
+<%= render(TaskList::Component.new(
+  [
+    {
+      title: "LoPS",
+      items: [
+        {
+          name: @teaching_authority_name,
+          link: [:review, :assessor_interface, @application_form, @assessment, :professional_standing_request],
+          status: @professional_standing_request.review_status,
+        }
+      ],
+    }
+  ]
+)) %>
 
 <div class="govuk-!-padding-top-3">
   <%= govuk_button_link_to "Back to overview", [:assessor_interface, @application_form] %>
--- a/spec/system/assessor_interface/reviewing_professional_standing_spec.rb
+++ b/spec/system/assessor_interface/reviewing_professional_standing_spec.rb
@@ -94,8 +94,9 @@ RSpec.describe "Assessor reviewing verifications", type: :system do
   end
 
   def when_i_click_on_lops
-    assessor_review_verifications_page.rows.first.link.click
+    find(".govuk-link").click
   end
+  
 
   def when_i_fill_in_the_review_lops_form
     form = assessor_review_professional_standing_request_page.form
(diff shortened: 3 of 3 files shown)


app/controllers/assessor_interface/review_verifications_controller.rb
```diff
@@ -11,6 +11,14 @@ def index
       @professional_standing_request =
         assessment.professional_standing_request if assessment.professional_standing_request.verify_failed?
 
+      @region = @assessment.application_form.region
+
+      @teaching_authority_name =
+        (
+          @region.teaching_authority_name.presence ||
+            "Contact relevant competent authority"
+        )
```
thomasleese, line 20: I think we should use this helper interface: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/29cbfd24eac87c84b5db1e4f9b363e6edfe9f254/app/helpers/region_helper.rb#L14-L16

If we need a custom default value, I think it makes sense to pass that as a new argument to the helper, but I think every region will have a teaching authority name set.

app/views/assessor_interface/review_verifications/index.html.erb
```diff
@@ -7,17 +7,20 @@
...
-    table.with_caption(text: "LoPS")
-    table.with_body do |body|
-      body.with_row do |row|
-        row.with_cell { govuk_link_to region_certificate_name(@application_form.region), [:review, :assessor_interface, @application_form, @assessment, :professional_standing_request] }
-        row.with_cell { render(StatusTag::Component.new(@professional_standing_request.review_status)) }
-      end
-    end
-  end
-end %>
+<%= render(TaskList::Component.new(
```
thomasleese, line 10: Nice refactor!

app/views/assessor_interface/review_verifications/index.html.erb
```diff
@@ -7,17 +7,20 @@
   The following verification tasks have been flagged for review:
 </p>
 
-<%= govuk_table do |table|
-  if @professional_standing_request.present?
```
thomasleese, line 11: Although we only support LoPS at the moment we'll need this condition in the future when we implement review of qualifications and references so I think it makes sense to keep it here.

spec/system/assessor_interface/reviewing_professional_standing_spec.rb
```diff
@@ -94,7 +94,7 @@ def when_i_click_on_assessment_decision
   end
 
   def when_i_click_on_lops
-    assessor_review_verifications_page.rows.first.link.click
+    find(".govuk-link").click
```
thomasleese, line 97: Rather than calling `find` we should update the page object (https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/spec/support/autoload/page_objects/assessor_interface/review_verifications.rb) with the new structure of the page. We can use the existing page object for tasks lists: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/spec/support/autoload/page_objects/task_list.rb

Here's an example of it being used:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/29cbfd24eac87c84b5db1e4f9b363e6edfe9f254/spec/support/autoload/page_objects/teacher_interface/application.rb#L13

spec/system/assessor_interface/reviewing_professional_standing_spec.rb
```diff
@@ -109,15 +109,14 @@ def when_i_click_on_back_to_overview
   end
 
   def and_i_see_the_lops_not_started
-    expect(assessor_review_verifications_page.rows.first.tag.text).to eq(
-      "NOT STARTED",
-    )
+    status_text = find(".govuk-tag.govuk-tag--grey.app-task-list__tag").text
```
thomasleese, line 112: Same applies here about using `find` directly as above.

spec/system/assessor_interface/reviewing_professional_standing_spec.rb
```diff
@@ -109,15 +109,14 @@ def when_i_click_on_back_to_overview
...
+    expect(status_text).to eq("NOT STARTED")
   end
 
   def and_i_see_the_lops_accepted
-    expect(assessor_review_verifications_page.rows.first.tag.text).to eq(
-      "ACCEPTED",
-    )
+    status_text =
+      find("strong.govuk-tag.govuk-tag--green.app-task-list__tag").text
```
thomasleese, line 118: Same applies here about using `find` directly as above.

app/controllers/assessor_interface/review_verifications_controller.rb
```diff
@@ -11,6 +12,8 @@ def index
       @professional_standing_request =
         assessment.professional_standing_request if assessment.professional_standing_request.verify_failed?
 
+      @teaching_authority_name = region_teaching_authority_name(@assessment.application_form.region, "Contact relevant competent authority")
```
thomasleese, line 15: One of the benefits of helpers is that we can use them directly in the views, it means all the presentation logic (i.e. how things look on the page) can kept in the view rather than in the view and the controller.

spec/support/autoload/page_objects/assessor_interface/review_verifications.rb
```diff
@@ -11,6 +11,8 @@ class ReviewVerifications < SitePrism::Page
         element :tag, ".govuk-tag"
       end
 
+      section :task_list, TaskList, ".app-task-list"
```
thomasleese, line 14: Nice, we should be able to remove lines 9-12 now as they shouldn't be being used anywhere.

spec/support/autoload/page_objects/task_list.rb
```diff
@@ -1,5 +1,8 @@
 module PageObjects
   class TaskList < SitePrism::Section
+    element :not_started_status, ".govuk-tag.govuk-tag--grey"
+    element :accepted_status, ".govuk-tag.govuk-tag--green"
```
thomasleese, line 4: Rather than adding elements specific to one page in the `TaskList` page object, it's better to use the `find_item` method below. For example, something like:

```rb
item = assessor_application_page.task_list.find_item("Contact relevant competent authority")
expect(item.status_tag.text).to eq("NOT STARTED")
```

This has the benefit of implicitly checking that the right item exists in the task list at the same time.

app/controllers/assessor_interface/review_verifications_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 module AssessorInterface
   class ReviewVerificationsController < BaseController
+    include RegionHelper
```
thomasleese, line 5: I don't think we need this line as the helper isn't used in this controller.

app/views/assessor_interface/review_verifications/index.html.erb
```diff
@@ -7,17 +7,22 @@
...
-        row.with_cell { govuk_link_to region_certificate_name(@application_form.region), [:review, :assessor_interface, @application_form, @assessment, :professional_standing_request] }
-        row.with_cell { render(StatusTag::Component.new(@professional_standing_request.review_status)) }
-      end
-    end
+    [
+      {
+        title: "LoPS",
+        items: [
+          {
+            name: (region_teaching_authority_name(@application_form.region) == "teaching authority") ? "Contact relevant competent authority" : region_teaching_authority_name(@application_form.region),
```
thomasleese, line 17: I think it's better to have the default in the helper itself as a second argument, and then pass that in, I think you had that at one point in this PR.

spec/support/autoload/page_objects/task_list.rb
thomasleese, line 4: I think we can remove these lines.

thomasleese (approved):

PR #1787, opened 2023-11-06
Split verification in to two pages

thomasleese:
If the admin decides that the LoPS is not acceptable, and they'd like to send this to review, we've decided to split this in to two pages so the internal notes appears on its own page.


richardpattinson (approved):
Looks good to me

PR #1791, opened 2023-11-08
Use references in URLs

thomasleese:
This changing the routing and URLs to use the application form reference rather than the ID. This change allows the assessors to navigate directly to an application if it's no longer visible in the case management list, since the IDs are an internal identifier and not presented anywhere.

[Trello Card](https://trello.com/c/pJXGiS2v/2345-switch-application-form-urls-to-use-references)


syed87 (approved):
looks good to me

PR #1795, opened 2023-11-09
Remove dates from PII information

thomasleese:
We'd like to do analysis on these fields and they're not PII so we can expose them in the analytics.


syed87 (approved):
looks good to me

PR #1798, opened 2023-11-09
Add teacher facing failure reasons

thomasleese:
This adds content for failures reasons show to teachers, as opposed to the current ones we show which are for the assessors.

[Trello Card](https://trello.com/c/ZgVtLWdj/2392-review-fi-decline-reasons-so-that-they-speak-directly-to-the-applicant)


syed87 (approved):
looks good to me

PR #1799, opened 2023-11-10
Show email previews without an API key

thomasleese:
We often don't have a GOV.UK Notify API key available when running the application locally, and if we're not testing the previews specifically it can be a significant overhead to get one. I've changed the mailer preview to handle this case gracefully.

In production we already require an API key in the ActionMailer initialiser so this doesn't introduce the ability to run in production without an API key.


syed87 (approved):
looks good to me

PR #1800, opened 2023-11-10
Stop recording status changed timeline events

thomasleese:
We're going to be shortly removing this event anyway and it clutters up the timeline events view with information that's not shown anywhere else, so we can safely remove this.


syed87 (approved):
looks good to me

PR #1801, opened 2023-11-10
Set application form status columns in one save

thomasleese:
This avoids the need for an SQL update command for each field that changes and keeps it as one single SQL update. This should have a small performance improvement, and makes the analytics clearer as we'll be sending a single update to BigQuery rather than numerous smaller ones.


syed87 (approved):
looks good to me

PR #1803, opened 2023-11-10
Reduce duplicate status changes

thomasleese:
This reduces the number of times we update the status from the `RequestRequestable` service class to remove duplicate timeline events. Everywhere that uses this class now calls the status updater anyway as part of normal operation, so we don't need it here.


syed87 (approved):
looks good to me

PR #1810, opened 2023-11-14
Reload after requesting a requestable

thomasleese:
This should ensure that when we then update the status of the application, it's fully up to date with what we would expect.

[Trello Card](https://trello.com/c/iD0Y1JEx/2395-investigate-applications-showing-at-incorrect-status)


thomasleese:
Merging to get the bug fixed while there's no other dev available.

PR #1822, opened 2023-11-23
Improve assessor work history duration calculation

thomasleese:
This improves the work history duration calculation for assessors by considering the certificate of the teaching qualification when compared to the start date of the work history.

### DIFF
--- a/lib/tasks/application_forms.rake
+++ b/lib/tasks/application_forms.rake
@@ -34,47 +34,4 @@ namespace :application_forms do
         puts "#{application_form.reference}: #{application_form.action_required_by} - #{application_form.status}"
       end
   end
-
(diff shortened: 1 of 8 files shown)


lib/tasks/application_forms.rake
```diff
@@ -34,6 +34,7 @@ namespace :application_forms do
         puts "#{application_form.reference}: #{application_form.action_required_by} - #{application_form.status}"
       end
   end
+<<<<<<< HEAD
```
syed87, line 37: does this need to be removed?

thomasleese: Yeah, a rebase that went wrong, I've fixed this now!

syed87 (approved):
looks good to me

PR #1833, opened 2023-11-29
Update content on decline email

thomasleese:
This updates the content shown when we decline due to the work history.


syed87 (approved):
looks good to me

PR #1838, opened 2023-11-30
Update work history duration decline content

thomasleese:
We want to change it to be two paragraphs rather than four sentences showing as individual paragraphs.


syed87 (approved):
looks good to me

PR #1841, opened 2023-12-05
Refactor WorkHistoryHelper

thomasleese:
This changes the content shown for each work history to match the new designs, and removes the need to pass in an index.


syed87 (approved):
looks good to me

PR #1842, opened 2023-12-05
Refactor recommendation controllers

thomasleese:
This changes the controllers so the show method is the entrypoint rather than the edit method, allowing us to implement a summary page for the verify flow.


syed87 (approved):
looks good to me

PR #1845, opened 2023-12-05
Make teaching qualification duration optional

thomasleese:
We want to do some final calculations on this so we need to bring back the ability to set this to false. This will be reverted once we've done the analysis.

[Jira Issue](https://dfedigital.atlassian.net/browse/AQTS-52)


syed87 (approved):
looks good to me

PR #1847, opened 2023-12-06
Set app_replicas to 2 for aks production

johnake:


thomasleese (commented):
The code looks good, I've just left a few comments on the style. Could we also add a description to the commit explaining why we're making this change?

PR #1848, opened 2023-12-06
Set app_replicas to 2 for production

johnake:


johnake:
[Increase Apply for QTS replicas](https://trello.com/c/MlkqcnGV/811-increase-apply-for-qts-replicas)

terraform/application/application.tf
```diff
@@ -49,6 +49,7 @@ module "web_application" {
 
   kubernetes_config_map_name = module.application_configuration.kubernetes_config_map_name
   kubernetes_secret_name     = module.application_configuration.kubernetes_secret_name
+  replicas                   = var.app_replicas
```
thomasleese, line 52: I'd suggest this belongs with the `docker_image` parameter rather than with the Kubernetes ones.

thomasleese (commented):
Looks good, could we also add a description to the commit explaining why we're making this change?

terraform/application/variables.tf
```diff
@@ -2,6 +2,12 @@ variable "app_environment" {
   type = string
 }
 
+variable "app_replicas" {
+  type        = number
+  description = "number of replicas of the web app"
```
saliceti, line 7: Number instead of number please

thomasleese (approved):

thomasleese (commented):
Code looks great, I would like to see a description in the commit message of why we're making this change.

I also wonder if there's anything we need to be careful of when deploying this as obviously we can't test it in any non-production environments? Should this change also be applied to pre-production?

saliceti:
> Code looks great, I would like to see a description in the commit message of why we're making this change.
> 
> I also wonder if there's anything we need to be careful of when deploying this as obviously we can't test it in any non-production environments? Should this change also be applied to pre-production?

On the infrastructure side, it's a very common configuration so there is no risk here. The only issue would be if the app is stateful and keeps something in the container as requests will now be spread across 2 instances. But I think we would have seen issue before, as containers are restarted from time to time which would remove the state.

thomasleese (approved):

PR #1851, opened 2023-12-06
Refactor form views

thomasleese:
This refactors the form views to make use of the [`title_with_error_prefix` helper from govuk-components](https://govuk-components.netlify.app/helpers/title-with-error-prefix/). I've also refactored the form variables and removes unnecessary code.

I expect this PR is easier to review commit by commit.


syed87 (approved):
looks good to me

PR #1855, opened 2023-12-07
Refactor assessor policies

thomasleese:
This refactors the policies used in the assessor interface to have bespoke policies for each kind of controller/model allowing better flexibility as we've got more permissions now and as we build the new verification journeys.


syed87 (approved):
looks good to me

PR #1857, opened 2023-12-07
Only show assessors who are assigned to applications

thomasleese:
This will reduce the number of assessors shown in the list, and it also means we can take away the assess permission from staff who have left without them disappearing from this list.


syed87 (approved):
looks good to me

PR #1867, opened 2023-12-12
Allow indexing references and qualifications

thomasleese:
There's no sensitive information on these pages and no action that can be taken.


thomasleese:
Merging this in to fix #1855

PR #1871, opened 2023-12-15
Swap qualifications

thomasleese:
This adds a service which allows us to swap the qualifications on an application, effectively changing which one is the teaching qualification. We also record a note to ensure that we have a record of this being done.

I've tested the task locally, but we will also run this in pre-production before applying to production.


richardpattinson (approved):
Looks great, cheers =)

PR #1881, opened 2023-12-19
Handle applications from countries which are now ineligible

thomasleese:
This adds the ability to delete or decline applications which are now associated with a country that has become ineligible since the applications were started or submitted. For draft applications we send a bespoke email before deleting the application, and for submitted applications they are declined and see an appropriate reason.


syed87 (approved):
looks good to me

PR #1883, opened 2023-12-21
Remove hidden_from_assessment column

thomasleese:
This removes a column which we used previously to hide applications from assessors, but we're no longer using this feature so we can remove the column to reduce maintenance.

This can only be merged after #1881 is deployed.


syed87 (approved):
looks good to me

PR #1886, opened 2023-12-21
Remove BackLinkHelper

thomasleese:
We're not using a `next` parameter on the eligibility checker where this helper is being used, so we can reduce the chances of it being used maliciously.


thomasleese:
Merging this in to fix a medium security vulnerability found in our ITHC.

PR #1889, opened 2023-12-22
Simplify migration to add rules

thomasleese:
This has already been run in production so we don't need it to be any more complicated, and [it's currently causing problems with deployment of review apps](https://dfe-teacher-services.sentry.io/issues/4747320438/).


thomasleese:
Merging this in as it's blocking review deployments.

PR #1892, opened 2023-12-22
Rename radio items for site_prism

thomasleese:
In the new version of site_prism we need to rename the `no_radio_item` DSL name as it's no longer compatible with the latest version.


thomasleese:
Merging this to fix some tests with the GOV.UK Frontend upgrade.

PR #1893, opened 2023-12-22
Upgrade GOV.UK Frontend

thomasleese:
This upgrades to GOV.UK Frontend version 5.0 which introduces a number of breaking changes and new features, notable a new style for tags and the task list is now a built in component.

Replacing #1866 as the build wasn't passing.


syed87 (approved):
looks good to me

PR #1898, opened 2023-12-27
Upgrade to Ruby 3.3.0

thomasleese:
This version was released a couple of days ago and doesn't have any breaking changes with our service, so we should upgrade as soon as possible to avoid ending up out of date.


syed87 (approved):
looks good to me

PR #1906, opened 2024-01-03
Enable analytics entity table check

thomasleese:
This enables a new feature of DfE Analytics that allows us to check the latest version of an entity table in BigQuery is in sync with the database.


syed87 (approved):
looks good to me

PR #1917, opened 2024-01-10
Aqts 36 add decline note to timeline for all cases

syed87:
This feature will add a Declined status tag under a new Application Declined timeline even for all cases where the application has been declined. All notes for decline reason will also show in the timeline event.

https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-36

### DIFF
--- a/app/components/timeline_entry/component.html.erb
+++ b/app/components/timeline_entry/component.html.erb
@@ -61,6 +61,24 @@
         has changed from <strong class="govuk-!-font-weight-bold"><%= description_vars[:old_value] %></strong>
         to <strong class="govuk-!-font-weight-bold"><%= description_vars[:new_value] %></strong>.
       </p>
+
+    <% elsif timeline_event.application_declined? %>
+      <p class="govuk-body">
+        Status has changed to: <%= render(StatusTag::Component.new(:declined)) %>
+      </p>
+      
+      <% TeacherInterface::ApplicationFormViewObject.new(application_form: timeline_event.application_form).declined_reasons.each do |title, reasons| %>
+        <% if title.present? %>
+          <h4 class="govuk-heading-s"><%= title %></h4>
+        <% end %>
+
+        <ul class="govuk-list govuk-list--bullet">
+          <% reasons.each do |reason| %>
+            <li><%= simple_format reason %></li>
+          <% end %>
+        </ul>
+      <% end %>
+
     <% else %>
       <p class="govuk-body"><%= simple_format description.html_safe %></p>
     <% end %>
--- a/app/components/timeline_entry/component.rb
+++ b/app/components/timeline_entry/component.rb
@@ -199,5 +199,9 @@ module TimelineEntry
           ).strip,
       }
     end
+
+    def application_declined_vars
+      {}
+    end
   end
 end
--- a/app/services/decline_qts.rb
+++ b/app/services/decline_qts.rb
@@ -14,6 +14,7 @@ class DeclineQTS
     ActiveRecord::Base.transaction do
       application_form.update!(declined_at: Time.zone.now)
       ApplicationFormStatusUpdater.call(application_form:, user:)
+      CreateTimelineEvent.call("application_declined", application_form:, user:)
     end
 
     TeacherMailer.with(application_form:).application_declined.deliver_later
--- a/spec/system/assessor_interface/completing_assessment_spec.rb
+++ b/spec/system/assessor_interface/completing_assessment_spec.rb
@@ -8,6 +8,9 @@ RSpec.describe "Assessor completing assessment", type: :system do
     double(generate_template_preview: notify_template_preview)
   end
   let(:notify_template_preview) { double(html: "I am an email") }
+  let(:assessor_user) do
+    create(:staff, :with_assess_permission, :confirmed, name: "Authorized User")
+  end
 
   around do |example|
     ClimateControl.modify GOVUK_NOTIFY_API_KEY: notify_key do
@@ -184,6 +187,9 @@ RSpec.describe "Assessor completing assessment", type: :system do
 
     when_i_click_on_overview_button
     then_the_application_form_is_declined
+
+    and_i_click_view_timeline
+    then_i_see_application_declined
   end
 
   private
@@ -397,6 +403,21 @@ RSpec.describe "Assessor completing assessment", type: :system do
     expect(assessor_application_page.status_summary.value).to have_text(
       "DECLINED",
     )
+  end
+
+  def and_i_click_view_timeline
+    assessor_application_page.click_link(
+      "View timeline of this applications actions",
+    )
+  end
+
+  def then_i_see_application_declined
+    declined_timeline_event =
+      assessor_timeline_page.timeline_items.find do |timeline_event|
+        timeline_event.title.text.include?("Application declined")
+      end
+
+    expect(declined_timeline_event).to_not be_nil
   end
 
   delegate :reference, to: :application_form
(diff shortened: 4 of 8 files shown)


app/components/timeline_entry/component.rb
```diff
@@ -199,5 +199,21 @@ def stage_changed_vars
...
+            StatusTag::Component.new(
+              "declined",
+              class_context: "timeline-event",
+            ),
+          ).strip
+        else
+          ""
+        end
+
+      { declined: }
```
thomasleese, line 216: It looks like this `declined` variable isn't used anywhere so we can probably delete it.

app/models/further_information_request.rb
```diff
@@ -67,8 +67,11 @@ def after_received(*)
   end
 
   def after_expired(user:)
-    if application_form.withdrawn_at.nil?
-      DeclineQTS.call(application_form:, user:)
+    return if application_form.withdrawn_at.present?
+
+    ActiveRecord::Base.transaction do
+      application_form.update!(declined_at: Time.zone.now)
+      ApplicationFormStatusUpdater.call(application_form:, user:)
```
thomasleese, line 74: Why don't we use the `DeclineQTS` service here anymore?

Without it we won't get this timeline event or decline email.

spec/system/assessor_interface/completing_assessment_spec.rb
```diff
@@ -410,6 +416,31 @@ def then_the_application_form_is_declined
...
+  def and_i_click_view_timeline
+    assessor_application_page.click_link(
+      "View timeline of this applications actions",
+    )
+  end
+
+  def then_i_see_application_declined
+    timeline_item = assessor_timeline_page.timeline_items.first
+
+    unless timeline_item.title.has_content?("Email sent")
```
thomasleese, line 428: Rather than checking the title of the first timeline event, we should make sure that a timeline event exists with that title regardless of whether it's the first one, by searching through the array of `timeline_events`.

spec/system/assessor_interface/completing_assessment_spec.rb
```diff
@@ -410,6 +416,31 @@ def then_the_application_form_is_declined
...
+      expect(timeline_item.title).to have_content("Application declined")
+    end
+  end
+
+  def create_timeline_event_for_decline
```
thomasleese, line 433: It looks like this method isn't used so maybe we don't need it.

app/services/decline_qts.rb
```diff
@@ -14,6 +14,16 @@ def call
...
+
+      creator = user.is_a?(String) ? nil : user
+      creator_name = user.is_a?(String) ? user : ""
+
+      TimelineEvent.create!(
+        event_type: "application_declined",
+        creator:,
+        creator_name:,
+        application_form:,
+      )
```
thomasleese, line 26: Could we put these lines in a private method on the class? I think this `def call` method is getting quite long now.

app/components/timeline_entry/component.rb
```diff
@@ -199,5 +199,21 @@ def stage_changed_vars
...
+          if timeline_event.application_form.declined_at?
+            render(
+              StatusTag::Component.new(
+                "declined",
+                class_context: "timeline-event",
+              ),
+            ).strip
+          else
+            ""
+          end,
```
thomasleese, line 215: It doesn't look like this variable is used anywhere so I think we can remove it.

app/components/timeline_entry/component.html.erb
```diff
@@ -61,6 +61,25 @@
...
+        <h4 class="govuk-heading-s"><%= title %></h4>
+      <% end %>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <% reasons.each do |reason| %>
+          <li><%= simple_format reason %></li>
+        <% end %>
+      </ul>
+    <% end %>
+  </p>
```
thomasleese, line 81: It looks like this lines need indenting in by 4 spaces.

app/components/timeline_entry/component.html.erb
```diff
@@ -61,6 +61,25 @@
...
+            <h4 class="govuk-heading-s"><%= title %></h4>
+          <% end %>
+
+          <ul class="govuk-list govuk-list--bullet">
+            <% reasons.each do |reason| %>
+              <li><%= simple_format reason %></li>
+            <% end %>
+          </ul>
+        <% end %>
+      </p>
```
thomasleese, line 81: I don't think this section needs to be wrapped in a paragraph (`<p>`), since we've got headings and lists inside which doesn't quite make sense.

app/components/timeline_entry/component.rb
```diff
@@ -199,5 +199,15 @@ def stage_changed_vars
...
+
+    def application_declined_vars
+      if timeline_event.application_form.declined_at?
+        render(
+          StatusTag::Component.new("declined", class_context: "timeline-event"),
+        ).strip
+      else
+        ""
+      end
+    end
```
thomasleese, line 211:
```suggestion
    def application_declined_vars
      {}
    end
```

I don't think we're using any variables from this method in the view so we can return an empty hash here.

app/services/decline_qts.rb
```diff
@@ -13,6 +13,7 @@ def call
 
     ActiveRecord::Base.transaction do
       application_form.update!(declined_at: Time.zone.now)
+      create_timeline_event
```
thomasleese, line 16: Could we create the timeline event after `ApplicationFormStatusUpdater.call`? This would mean that the application declined event appears as the most recent event and the stage/action required by changed timeline events appear below it.

spec/system/assessor_interface/completing_assessment_spec.rb
```diff
@@ -410,6 +416,20 @@ def then_the_application_form_is_declined
...
+      "View timeline of this applications actions",
+    )
+  end
+
+  def then_i_see_application_declined
+    expect(
+      assessor_timeline_page.timeline_items.any? do |event|
+        event.title.has_content?("Application declined")
+      end,
+    ).to be_truthy
```
thomasleese, line 430: This works, but I wonder if this might be a better way of structuring it (to make it clear what the code is doing):

```rb
declined_timeline_event = assessor_timeline_page.timeline_items.find do |timeline_event|
  timeline_event.title == "Application declined"
end

expect(declined_timeline_event).to_not be_nil
```

It's probably a good idea to add a test to the `DeclineQTS` service to check that the timeline event is being created. You can use the [`have_recorded_timeline_event` matcher](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/spec/support/matchers/have_recorded_timeline_event.rb), for example something like this:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/79708ea2fe9763b7de0955a46c80c1d502a6b57e/spec/services/create_note_spec.rb#L28-L33

thomasleese (commented):
Looks pretty close

thomasleese (approved):
Nice, all looks good to me.

app/services/decline_qts.rb
```diff
@@ -23,4 +24,16 @@ def call
...
+    creator = user.is_a?(String) ? nil : user
+    creator_name = user.is_a?(String) ? user : ""
+
+    TimelineEvent.create!(
+      event_type: "application_declined",
+      creator:,
+      creator_name:,
+      application_form:,
+    )
+  end
```
thomasleese, line 38: I think we can remove this method as it's not being used anywhere.

thomasleese (approved):
Looks good, can you make sure to squash all the commits together and add a description to the single commit before merging.

PR #1918, opened 2024-01-11
Improve example data generation

thomasleese:
This improves the example data that we generate for work history and qualifications to make it more realistic and make it easier for product and design reviews.


syed87 (approved):
looks good to me

PR #1919, opened 2024-01-11
Small improvements to factories

thomasleese:
This makes a number of small improvements to the factories to ensure they better represent real examples.


syed87 (approved):
looks good to me

PR #1921, opened 2024-01-15
Preparation for GOV.UK Frontend upgrade

thomasleese:
This makes a number of changes to the views which has been extracted from #1893 but can be merged in advance to reduce the number of changes in that PR.


syed87 (approved):
looks good to me

PR #1922, opened 2024-01-15
Add create timeline event service

thomasleese:
This adds a new service for creating timeline events which encapsulates the logic around whether the `user` is a `String` or an instance of a `Staff` model.


syed87 (approved):
looks good to me

PR #1923, opened 2024-01-15
Hide bullet points when there's one decline reason

thomasleese:
This changes the design of the decline pages so that if there's only one decline reason we show only a single bullet point.

### DIFF
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -9,11 +9,15 @@
     <h4 class="govuk-heading-s"><%= title %></h4>
   <% end %>
 
-  <ul class="govuk-list govuk-list--bullet">
-    <% reasons.each do |reason| %>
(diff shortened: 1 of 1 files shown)


app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -9,11 +9,15 @@
     <h4 class="govuk-heading-s"><%= title %></h4>
   <% end %>
 
-  <ul class="govuk-list govuk-list--bullet">
-    <% reasons.each do |reason| %>
-      <li><%= simple_format reason %></li>
-    <% end %>
-  </ul>
+  <% if reasons.count == 1 %>
+    <%= simple_format reasons.first %>
```
syed87, line 13: would this single reason be showing a bullet point? The commit suggests you'd want a single bullet point but I don't think the code would use a bullet point here.

syed87 (approved):
looks good to me

PR #1927, opened 2024-01-16
Remove some fields from PII list

thomasleese:
The job title and the institution country code are not personally identifable pieces of information so we can send these to analytics.


syed87 (approved):
looks good to me

PR #1928, opened 2024-01-16
Refactor TeacherMailer to take the application form

thomasleese:
Currently the `TeacherMailer` takes in the teacher as a parameter and then extracts the most recent application form from it. This works in most cases, but there are a few specific cases where this causes problems, notably if the teacher has multiple application forms. In this case we can end up choosing the incorrect application form and emails can end up associated with the wrong application form.


syed87 (approved):
looks good to me

PR #1932, opened 2024-01-18
Update qualification swap approval date

thomasleese:
The new batch was approved today and we need to change the message that's shown.


syed87 (approved):
looks good to me

PR #1933, opened 2024-01-18
Stop calculating work history after award/decline

thomasleese:
When an application is awarded or declined, and the work history is marked as ongoing, we stop counting the number of months when the application is awarded or declined to ensure the number doesn't keep increasing.


syed87 (approved):
looks good to me

PR #1937, opened 2024-01-22
Refactor work history order scopes

thomasleese:
We've got two ways of ordering the work history so we should have these two scopes available and use them in all places.


syed87 (approved):
looks good to me

PR #1938, opened 2024-01-22
Refactor PreliminaryAssessmentSectionsFactory spec

thomasleese:
This refactors the spec to reduce the changes of a country code conflict where the application form happens to be created with a country that has already been created.


syed87 (approved):
looks good to me

PR #1944, opened 2024-01-23
Update review error message

thomasleese:
When the assessor has chosen a response, we want a clearer error message.


syed87 (approved):
lgtm

PR #1946, opened 2024-01-24
Remove performance page

thomasleese:
This removes the performance page as the data is available in our other dashboards.


syed87 (approved):
lgtm

PR #1953, opened 2024-01-26
Enable azure rbac deployment

RMcVelia:
### Trello card

https://trello.com/c/3m8PNLG6/937-enable-azure-rbac-deployment-on-all-services

Not to be deployed until terraform modules has been updated on Monday Jan 29th.

### Context

Enable azure rbac on deployment (for when it is enabled in the cluster)

### Changes proposed in this pull request

Update Makefile, terraform and workflows to use rbac if configured for the cluster

### Guidance to review

make env terraform-plan
check review app deploys successfully

### DIFF
--- a/.github/actions/deploy-environment/action.yml
+++ b/.github/actions/deploy-environment/action.yml
@@ -33,7 +33,7 @@ runs:
         terraform_version: 1.5.0
         terraform_wrapper: false
 
-    - uses: DFE-Digital/github-actions/set-arm-environment-variables@master
+    - uses: DFE-Digital/github-actions/set-kubelogin-environment@master
       with:
         azure-credentials: ${{ inputs.azure-credentials }}
 
@@ -56,8 +56,10 @@ runs:
       if: ${{ inputs.pull-request-number != '' }}
       shell: bash
       run: |
-        az aks get-credentials -g s189t01-tsc-ts-rg -n s189t01-tsc-test-aks
-        kubectl exec -n tra-development deployment/apply-for-qts-review-${{ inputs.pull-request-number }}-web -- sh -c "cd /app && /usr/local/bin/bundle exec rails db:seed review_app:configure example_data:generate"
+        make ci review get-cluster-credentials
+        kubectl exec -n tra-development deployment/apply-for-qts-review-${PULL_REQUEST_NUMBER}-web -- sh -c "cd /app && /usr/local/bin/bundle exec rails db:seed review_app:configure example_data:generate"
+      env:
+        PULL_REQUEST_NUMBER: ${{ inputs.pull-request-number }}
 
     - id: key-vault-name
       if: ${{ inputs.smoke-test-credentials-required == 'true' }}
--- a/Makefile
+++ b/Makefile
@@ -11,13 +11,13 @@ help: ## Show this help
 	@grep -E '^[a-zA-Z\._\-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 .PHONY: development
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -194,3 +194,15 @@ domains-apply: domains-init ## terraform apply for dns resources
...
 domains-destroy: domains-init ## terraform destroy for dns resources
 	terraform -chdir=terraform/domains/environment_domains destroy -var-file config/$(CONFIG).tfvars.json
+
+test-cluster:
+	$(eval CLUSTER_RESOURCE_GROUP_NAME=s189t01-tsc-ts-rg)
+	$(eval CLUSTER_NAME=s189t01-tsc-test-aks)
+
+production-cluster:
+	$(eval CLUSTER_RESOURCE_GROUP_NAME=s189p01-tsc-pd-rg)
+	$(eval CLUSTER_NAME=s189p01-tsc-production-aks)
```
thomasleese, line 204: Having the values hard coded in the Makefile means we need to keep these in sync with the cluster data in the Terraform modules. Can we instead use `terraform` to get these values like we do here:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/ec537b9518625b0d3039715936d0e6ba65988e3c/.github/actions/set-kubernetes-credentials/action.yaml#L37-L38

saliceti: The output is only present if you run terraform init, which requires a PIM request and the Contributor role. This is a security risk as the role has read/write access to all the azure resources of all the services.

We want to simplify and increase security and let the devs access AKS without elevating to Contributor. This will prevent running terraform by default. Unfortunately this causes a little bit of duplication.

.github/actions/deploy-environment/action.yml
```diff
@@ -56,8 +56,10 @@ runs:
       if: ${{ inputs.pull-request-number != '' }}
       shell: bash
       run: |
-        az aks get-credentials -g s189t01-tsc-ts-rg -n s189t01-tsc-test-aks
+        make ci review get-cluster-credentials
         kubectl exec -n tra-development deployment/apply-for-qts-review-${{ inputs.pull-request-number }}-web -- sh -c "cd /app && /usr/local/bin/bundle exec rails db:seed review_app:configure example_data:generate"
```
saliceti, line 60: `${{ inputs.pull-request-number }}` can be replaced by `${PULL_REQUEST_NUMBER}`

Makefile
thomasleese: Okay fair enough. I think we can remove those outputs as they were only used for this.

.github/actions/deploy-environment/action.yml
RMcVelia: updated

Makefile
thomasleese: We can also remove the outputs themselves as they were only used in the `set-kubernetes-credentials` action:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/cb1b9de3b42699c489313c6cac7d21620df54d73/terraform/application/outputs.tf#L13-L19

docs/infrastructure.md
```diff
@@ -73,11 +73,10 @@ Select account for az:
 $ az account set -s s189-teacher-services-cloud-test
 ```
 
-Get access credentials for a managed Kubernetes cluster (passing the
-resource group and the name):
+Get access credentials for a managed Kubernetes cluster (passing the afqts env):
```
thomasleese, line 76:
```suggestion
Get access credentials for a managed Kubernetes cluster (in this case for the `development` environment):
```

We don't really use the acronym `afqts` so this could be confusing for developers on the team.

Makefile
RMcVelia: removed

docs/infrastructure.md
RMcVelia: updated as suggested

saliceti (approved):

PR #1955, opened 2024-01-29
Decline email content update

syed87:
Content updates to decline email and sign in page

https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-82

### DIFF
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -30,8 +30,6 @@
   <% end %>
 <% end %>
 
-
-
 <%= render "teacher_interface/application_forms/show/what_can_you_do_next", view_object: %>
 
 <% unless view_object.from_ineligible_country? %>
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -1,10 +1,8 @@
-<h1>Apply for qualified teacher status (QTS)</h2>
-
 <h2 class="govuk-heading-l">Your QTS application has been declined</h2>
 
 <p class="govuk-body">Thank you for applying for qualified teacher status (QTS) and for your patience while we reviewed your application. Unfortunately, your application has been declined.</p>
 
-<h2 class="govuk-heading-m">Reason for decline</h2>
+<h3 class="govuk-heading-m">Reason for decline</h3>
 
 <% view_object.declined_reasons.each do |title, reasons| %>
   <% if title.present? %>
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -6,13 +6,13 @@
 
 <% view_object.declined_reasons.each do |title, reasons| %>
   <% if title.present? %>
-    <h3 class="govuk-heading-s"><%= title %></h3>
+    <h4 class="govuk-heading-s"><%= title %></h4>
   <% end %>
 
   <% if reasons.size == 1 %>
     <%= simple_format reasons.first[:name] %>
     <% if reasons.first[:assessor_note].present? %>
-      <h3 class="govuk-heading-s"><%= "Assessor notes" %></h3>
+      <h5 class="govuk-heading-s"><%= "Assessor notes" %></h5>
       <%= simple_format reasons.first[:assessor_note] %>
     <% end %>
   <% else %>
@@ -21,7 +21,7 @@
         <li>
           <%= simple_format reason[:name] %>
           <% if reason[:assessor_note].present? %>
-          <h3 class="govuk-heading-s"><%= "Assessor notes" %></h3>
+          <h5 class="govuk-heading-s"><%= "Assessor notes" %></h5>
             <%= simple_format reason[:assessor_note] %>
           <% end %>
         </li>
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -1,8 +1,10 @@
+<h1>Apply for qualified teacher status (QTS)</h2>
+
 <h2 class="govuk-heading-l">Your QTS application has been declined</h2>
 
-<p class="govuk-body">Thank you for applying for qualified teacher status (QTS) and for your patience while we reviewed your application.</p>
+<p class="govuk-body">Thank you for applying for qualified teacher status (QTS) and for your patience while we reviewed your application. Unfortunately, your application has been declined.</p>
 
-<h3 class="govuk-heading-m">Reason for decline</h3>
+<h2 class="govuk-heading-m">Reason for decline</h2>
 
 <% view_object.declined_reasons.each do |title, reasons| %>
   <% if title.present? %>
@@ -36,6 +38,7 @@
   <h3 class="govuk-heading-m">Decision review</h3>
 
   <p class="govuk-body">Applicants who have been declined for QTS are entitled to a review of the decline decision by a Casework Manager.</p>
+
   <p class="govuk-body">If you would like to request a decision review, you’ll need to provide:</p>
 
   <ul class="govuk-list govuk-list--bullet">
@@ -44,6 +47,8 @@
   </ul>
 
   <p class="govuk-body">Your request for review must be received within 28 days of receipt of the decision to decline QTS.</p>
+
   <p class="govuk-body">Email your request for review, including the information required as above, to <%= govuk_link_to t("service.email.enquiries"), "mailto:#{t("service.email.enquiries")}" %>.</p>
+
   <p class="govuk-body">If you are not satisfied with the outcome of the decision review, you can request a final formal appeal to a senior Teacher Qualification Manager.</p>
 <% end %>
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -6,12 +6,13 @@
 
 <% view_object.declined_reasons.each do |title, reasons| %>
   <% if title.present? %>
-    <h4 class="govuk-heading-s"><%= title %></h4>
+    <h3 class="govuk-heading-s"><%= title %></h3>
   <% end %>
 
   <% if reasons.size == 1 %>
     <%= simple_format reasons.first[:name] %>
     <% if reasons.first[:assessor_note].present? %>
+      <h3 class="govuk-heading-s"><%= "Assessor notes" %></h3>
       <%= simple_format reasons.first[:assessor_note] %>
     <% end %>
   <% else %>
@@ -20,6 +21,7 @@
         <li>
           <%= simple_format reason[:name] %>
           <% if reason[:assessor_note].present? %>
+          <h3 class="govuk-heading-s"><%= "Assessor notes" %></h3>
             <%= simple_format reason[:assessor_note] %>
           <% end %>
         </li>
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -12,7 +12,7 @@
   <% if reasons.size == 1 %>
     <%= simple_format reasons.first[:name] %>
     <% if reasons.first[:assessor_note].present? %>
-      <h5 class="govuk-heading-s"><%= "Assessor notes" %></h5>
+      <h5 class="govuk-heading-s">"Assessor notes"</h5>
       <%= simple_format reasons.first[:assessor_note] %>
     <% end %>
   <% else %>
@@ -21,7 +21,7 @@
         <li>
           <%= simple_format reason[:name] %>
           <% if reason[:assessor_note].present? %>
-          <h5 class="govuk-heading-s"><%= "Assessor notes" %></h5>
+          <h5 class="govuk-heading-s">"Assessor notes"</h5>
             <%= simple_format reason[:assessor_note] %>
           <% end %>
         </li>
--- a/app/views/teacher_interface/application_forms/show/_declined.html.erb
+++ b/app/views/teacher_interface/application_forms/show/_declined.html.erb
@@ -12,7 +12,7 @@
   <% if reasons.size == 1 %>
     <%= simple_format reasons.first[:name] %>
     <% if reasons.first[:assessor_note].present? %>
-      <h5 class="govuk-heading-s">"Assessor notes"</h5>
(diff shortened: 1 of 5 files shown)


app/views/teacher_mailer/application_declined.text.erb
```diff
@@ -19,29 +19,24 @@ Thank you for applying for QTS in England. Unfortunately, your application has b
 <% unless @view_object.declined_cannot_reapply? || @view_object.from_ineligible_country? %>
 # What you can do next
 
-You’ll be able to make a new application for QTS once you've addressed the decline reasons we've given. If you reapply for QTS, you’ll not be able to review the details of your previous application.
+You'll be able to make a new application once you've addressed all decline reasons given. If you reapply for QTS, you’ll not be able to review the details of your previous application.
```
thomasleese, line 22:
```suggestion
You’ll be able to make a new application once you've addressed all decline reasons given. If you reapply for QTS, you’ll not be able to review the details of your previous application.
```

app/views/teacher_mailer/application_declined.text.erb
```diff
@@ -19,29 +19,24 @@ Thank you for applying for QTS in England. Unfortunately, your application has b
...
+You may want to explore [other routes to getting QTS](https://www.gov.uk/government/publications/apply-for-qualified-teacher-status-qts-if-you-teach-outside-the-uk/routes-to-qualified-teacher-status-qts-for-teachers-and-those-with-teaching-experience-outside-the-uk).
 
-You may want to [explore other routes to getting QTS](https://www.gov.uk/government/publications/apply-for-qualified-teacher-status-qts-if-you-teach-outside-the-uk/routes-to-qualified-teacher-status-qts-for-teachers-and-those-with-teaching-experience-outside-the-uk).
+[Apply again](<%= new_teacher_interface_application_form_url %>)
 
 <% end %>
-<% if @view_object.from_ineligible_country? %>
-[Sign in](<%= new_teacher_session_url %>) for more information about QTS in England.
-<% else %>
 # Decision review
```
thomasleese, line 29: The decision review section should only be visible for countries that are eligible, so you'll need to keep in the conditional on `@view_object.from_ineligible_country?`.

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -1,8 +1,10 @@
+<h1>Apply for qualified teacher status (QTS)</h2>
```
thomasleese, line 1: I think this will already be visible from: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/ec537b9518625b0d3039715936d0e6ba65988e3c/app/views/teacher_interface/application_forms/show.html.erb#L3

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -1,8 +1,10 @@
...
+
 <h2 class="govuk-heading-l">Your QTS application has been declined</h2>
 
-<p class="govuk-body">Thank you for applying for qualified teacher status (QTS) and for your patience while we reviewed your application.</p>
+<p class="govuk-body">Thank you for applying for qualified teacher status (QTS) and for your patience while we reviewed your application. Unfortunately, your application has been declined.</p>
 
-<h3 class="govuk-heading-m">Reason for decline</h3>
+<h2 class="govuk-heading-m">Reason for decline</h2>
```
thomasleese, line 7: I think this should stay `h3` as it's not the same level as the `h2` above.

app/views/teacher_mailer/application_declined.text.erb
```diff
@@ -19,29 +19,27 @@ Thank you for applying for QTS in England. Unfortunately, your application has b
...
+You’ll be able to make a new application once you've addressed all decline reasons given. If you reapply for QTS, you’ll not be able to review the details of your previous application.
 
-[Apply again](<%= new_teacher_interface_application_form_url %>)
+You may want to explore [other routes to getting QTS](https://www.gov.uk/government/publications/apply-for-qualified-teacher-status-qts-if-you-teach-outside-the-uk/routes-to-qualified-teacher-status-qts-for-teachers-and-those-with-teaching-experience-outside-the-uk).
 
-You may want to [explore other routes to getting QTS](https://www.gov.uk/government/publications/apply-for-qualified-teacher-status-qts-if-you-teach-outside-the-uk/routes-to-qualified-teacher-status-qts-for-teachers-and-those-with-teaching-experience-outside-the-uk).
+[Apply again](<%= new_teacher_interface_application_form_url %>)
 
 <% end %>
 <% if @view_object.from_ineligible_country? %>
```
thomasleese, line 29:
```suggestion
<% unless @view_object.from_ineligible_country? %>
```

Otherwise the logic has switched, we'd only be showing the "Decision review" section to applications where the country is ineligible.

app/views/teacher_mailer/application_declined.text.erb
```diff
@@ -19,29 +19,27 @@ Thank you for applying for QTS in England. Unfortunately, your application has b
...
 
-Your request for review must be received within 28 days of receipt of the decision to decline QTS. Email your request for review, including the information required as above, to [<%= t("service.email.enquiries") %>](mailto:<%= t("service.email.enquiries") %>).
+Your request for review must be received within 28 days of receipt of the decision to decline QTS.
 
-If you are not satisfied with the outcome of the decision review, you can request a final formal appeal to a senior Teacher Qualification Manager.
-
-[Sign in](<%= new_teacher_session_url %>) to review the outcome of your application.
 <% end %>
 
+Email your request for review, including the information required as above to [<%= t("service.email.enquiries") %>](mailto:<%= t("service.email.enquiries") %>).
```
thomasleese, line 43: I think this line belongs within the `if` block as it relates to the "Decision review" section.

thomasleese (dismissed):
Looks good to me!

app/views/teacher_mailer/application_declined.text.erb
```diff
@@ -12,36 +12,34 @@ Thank you for applying for QTS in England. Unfortunately, your application has b
 
 <% end %>
 <% reasons.each do |reason| %>
+## Assessor notes
```
syed87, line 15: @thomasleese this shows as h3 on the front end, how can I make this h4?

thomasleese: I think Notify only supports two levels of headings:

https://www.notifications.service.gov.uk/using-notify/formatting#headings-and-subheadings

It doesn't support bold text so I don't think there are any other options here for subheadings.

config/locales/assessor_interface.en.yml
```diff
@@ -153,55 +153,55 @@ en:
 
       failure_reasons:
         as_statement:
-          additional_degree_certificate_illegible: An additional degree certificate (or translation) is illegible or in a format that we cannot accept.
```
thomasleese, line 156: These are the failure reasons as shown to the assessor, but the ones we show to the teacher are here instead:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/22dfe3ee3efaf7fdea747c0803a8ef86270409b0/config/locales/teacher_interface.en.yml#L6

If we want them to be the same (those shown to the assessor and those shown to the teacher) then we should delete one of these sets of failure reasons to avoid duplication.

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -14,7 +14,14 @@
   <% else %>
     <ul class="govuk-list govuk-list--bullet">
       <% reasons.each do |reason| %>
-        <li><%= simple_format reason %></li>
+        <% parts = reason.split("\n\n") %>
+        <% decline_reason = parts[0] %>
+        <% assessor_note = parts[1] %>
```
thomasleese, line 19: This approach doesn't work if the assessor adds multiple new lines to their note, in that case we'll only see the first paragraph of the assessor note. It also feels quite brittle, if we were to add a decline reason with newlines in it would break too. [The two strings are combined with the newlines in the view object](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/0e1178b4b394eba490a0b9d342d10aff58eb84e0/app/view_objects/teacher_interface/application_form_view_object.rb#L256) and then we're splitting them out again here, we could instead keep the two fields separate until they're rendered in the view.

I think a better alternative would be to change the structure of the return value of the [`declined_reasons`](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/0e1178b4b394eba490a0b9d342d10aff58eb84e0/app/view_objects/teacher_interface/application_form_view_object.rb#L62) to separate the assessor feedback from the decline reason. Currently the return value is something like:

```rb
{
  "section title" => [
    "decline reason \n\n assessor note"
  ]
}
```

But the return value could be changed to:

```rb
{
  "section title" => [
    {
      decline_reason: "...",
      assessor_note: "...",
  ]
}
```

This refactor will require some changes wherever `declined_reasons` is used so I would suggest you do the refactor first in a separate PR without changing any of the content (this can be deployed without needing a product review) and then make the content changes in this PR, this'll mean the two PRs are smaller and easier to review and understand.

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -1,17 +1,18 @@
 <h2 class="govuk-heading-l">Your QTS application has been declined</h2>
 
-<p class="govuk-body">Thank you for applying for qualified teacher status (QTS) and for your patience while we reviewed your application.</p>
+<p class="govuk-body">Thank you for applying for qualified teacher status (QTS) and for your patience while we reviewed your application. Unfortunately, your application has been declined.</p>
 
 <h3 class="govuk-heading-m">Reason for decline</h3>
 
 <% view_object.declined_reasons.each do |title, reasons| %>
   <% if title.present? %>
-    <h4 class="govuk-heading-s"><%= title %></h4>
```
thomasleese, line 9: This needs to stay `h4` as it belongs within the `h3` of "Reason for decline".

app/views/teacher_mailer/application_declined.text.erb
```diff
@@ -23,29 +24,26 @@ Thank you for applying for QTS in England. Unfortunately, your application has b
...
 # What you can do next
 
-You’ll be able to make a new application for QTS once you've addressed the decline reasons we've given. If you reapply for QTS, you’ll not be able to review the details of your previous application.
+You’ll be able to make a new application once you've addressed all decline reasons given. If you reapply for QTS, you’ll not be able to review the details of your previous application.
 
-[Apply again](<%= new_teacher_interface_application_form_url %>)
+You may want to explore [other routes to getting QTS](https://www.gov.uk/government/publications/apply-for-qualified-teacher-status-qts-if-you-teach-outside-the-uk/routes-to-qualified-teacher-status-qts-for-teachers-and-those-with-teaching-experience-outside-the-uk).
 
-You may want to [explore other routes to getting QTS](https://www.gov.uk/government/publications/apply-for-qualified-teacher-status-qts-if-you-teach-outside-the-uk/routes-to-qualified-teacher-status-qts-for-teachers-and-those-with-teaching-experience-outside-the-uk).
+<%= govuk_button_to 'Apply again', 'new_teacher_interface_application_form_url' %>
```
thomasleese, line 31:
```suggestion
<%= govuk_button_to 'Apply again', new_teacher_interface_application_form_url %>
```

The link should be going to the URL rather than the string `"new_teacher_interface_application_form_url"`.

Also, does this render correctly in the emails? `govuk_button_to` will return HTML which I didn't think was compatible with GOV.UK Notify.

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -1,17 +1,18 @@
...
+    <h3 class="govuk-heading-s"><%= title %></h3>
   <% end %>
 
   <% if reasons.size == 1 %>
     <%= simple_format reasons.first[:name] %>
     <% if reasons.first[:assessor_note].present? %>
+      <h3 class="govuk-heading-s"><%= "Assessor notes" %></h3>
```
thomasleese, line 15:
```suggestion
      <% if title.present? %>
        <h5 class="govuk-heading-s">Assessor notes</h5>
      <% else %>
        <h4 class="govuk-heading-s">Assessor notes</h4>
      <% end %>
```

This should be `h5` as it belongs within the `h4` of the section title, or a `h4` if there's no title as it belongs within the `h3` of "Reason for decline".

app/views/teacher_mailer/application_declined.text.erb
syed87: `govuk_button_to` doesn't render. I can't find another option for a button. Does that option exist?

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -12,6 +12,7 @@
   <% if reasons.size == 1 %>
     <%= simple_format reasons.first[:name] %>
     <% if reasons.first[:assessor_note].present? %>
+      <h5 class="govuk-heading-s"><%= "Assessor notes" %></h5>
```
thomasleese, line 15:
```suggestion
      <h5 class="govuk-heading-s">Assessor notes</h5>
```

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -20,6 +21,7 @@
         <li>
           <%= simple_format reason[:name] %>
           <% if reason[:assessor_note].present? %>
+          <h5 class="govuk-heading-s"><%= "Assessor notes" %></h5>
```
thomasleese, line 24:
```suggestion
          <h5 class="govuk-heading-s">Assessor notes</h5>
```

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -12,6 +12,7 @@
   <% if reasons.size == 1 %>
     <%= simple_format reasons.first[:name] %>
     <% if reasons.first[:assessor_note].present? %>
+      <h5 class="govuk-heading-s">"Assessor notes"</h5>
```
thomasleese, line 15:
```suggestion
      <h5 class="govuk-heading-s">Assessor notes</h5>
```

Without this, the quotes will be shown to the user which I don't think we want.

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -20,6 +21,7 @@
         <li>
           <%= simple_format reason[:name] %>
           <% if reason[:assessor_note].present? %>
+          <h5 class="govuk-heading-s">"Assessor notes"</h5>
```
thomasleese, line 24:
```suggestion
          <h5 class="govuk-heading-s">Assessor notes</h5>
```

thomasleese (approved):

PR #1963, opened 2024-02-01
Ensure all applications are verified

thomasleese:
Currently reduced evidence countries or countries where work history isn't required can be awarded without going through verification. We still want to verify these applications, just without any references.


syed87 (approved):
looks good to me

PR #1965, opened 2024-02-05
Add consent document types

thomasleese:
This adds two new kinds of document types, unsigned and signed consent documents, which will be used in the qualification verification flow. I've also refactored the models with documents by introducing a `Documentable` concern which handles the common logic of creating empty documents.


syed87 (approved):
looks good to me

PR #1966, opened 2024-02-05
Add unsigned_consent_document_downloaded column

thomasleese:
This adds a new column to qualification requests for capturing whether an applicant has downloaded the unsigned consent document.


syed87 (approved):
looks good to me

PR #1974, opened 2024-02-07
Change to assessment_declined_reasons to seperate return of assessor …

syed87:
This pull request modifies the format of decline reasons and assessor feedback within the assessment_declined_reasons method.

### DIFF
--- a/app/view_objects/teacher_interface/application_form_view_object.rb
+++ b/app/view_objects/teacher_interface/application_form_view_object.rb
@@ -63,32 +63,41 @@ class TeacherInterface::ApplicationFormViewObject
     if from_ineligible_country?
       country_name = CountryName.from_country(country)
       teaching_authority_name = region_teaching_authority_name(region)
-
       {
         "" => [
-          "As we are unable to verify professional standing documents with the #{teaching_authority_name} in " \
-            "#{country_name}, we have removed #{country_name} from the list of eligible countries.\n\nWe need to be " \
-            "able to verify all documents submitted by applicants with the relevant authorities. This is to ensure " \
-            "QTS requirements are applied fairly and consistently to every teacher, regardless of the country they " \
-            "trained to teach in.",
+          {
+            name:
+              "As we are unable to verify professional standing documents with the #{teaching_authority_name} in " \
+                "#{country_name}, we have removed #{country_name} from the list of eligible countries.\n\n" \
+                "We need to be able to verify all documents submitted by " \
+                "applicants with the relevant authorities. This is to ensure QTS requirements are applied " \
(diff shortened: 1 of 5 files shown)


app/view_objects/teacher_interface/application_form_view_object.rb
```diff
@@ -253,9 +253,9 @@ def assessment_declined_reasons
                  assessor_feedback = failure_reason.assessor_feedback
                ).present? &&
                  FailureReasons.decline?(failure_reason: failure_reason.key)
-              "#{title}\n\n#{assessor_feedback}"
+              { decline_reason: title, assessor_note: assessor_feedback }
```
thomasleese, line 256: Thinking about the structure of this actually, since it's coming from a method that's returning the "decline reasons" (which are made up of a name and optionally a note from the assessor) I reckon we should return the title under a `name` field:

```suggestion
              { name: title, assessor_note: assessor_feedback }
```

app/views/teacher_interface/application_forms/show/_declined.html.erb
```diff
@@ -9,17 +9,27 @@
     <h4 class="govuk-heading-s"><%= title %></h4>
   <% end %>
 
-  <% if reasons.count == 1 %>
-    <%= simple_format reasons.first %>
+  <% if reasons.size == 1 %>
+    <%= simple_format reasons.first[:decline_reason] %>
+    <% if reasons.first[:assessor_note].present? %>
+      <p><%= simple_format reasons.first[:assessor_note] %></p>
```
thomasleese, line 15:
```suggestion
      <%= simple_format reasons.first[:assessor_note] %>
```

`simple_format` already returns a paragraph tag so you don't need to wrap this in a paragraph.

thomasleese (commented):
The refactor looks good, [doing a search for `declined_reasons`](https://github.com/search?q=repo%3ADFE-Digital%2Fapply-for-qualified-teacher-status%20declined_reasons&type=code) shows a few other places where this method is used that I think we'll need to consider (and perhaps add some test so they're caught if this changes in the future):

### Application declined email

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/0901bc80c6555fa25f1bf95f4e4d7553c7a3d5d1/app/views/teacher_mailer/application_declined.text.erb#L9-L18

### Application declined timeline event

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/0901bc80c6555fa25f1bf95f4e4d7553c7a3d5d1/app/components/timeline_entry/component.html.erb#L70-L80

thomasleese (approved):

thomasleese (approved):

PR #1978, opened 2024-02-08
Handle documents with a single upload only

thomasleese:
This adds the ability for some documents to only support a single upload, which is what we need for the consent documents.


syed87 (approved):
looks good to me

PR #1984, opened 2024-02-13
Allow sending reminders for consent requests

thomasleese:
This ensures that we send reminder emails for consent if we haven't received a response from the user in 3 weeks. It uses the same reminder mechanism for the reference requests, although as we don't send individual reminders we don't mark the qualification request class as remindable.


syed87 (approved):
lgtm

PR #1986, opened 2024-02-13
Add consent statuses

thomasleese:
This adds the new consent statuses used in the new consent workflow.


syed87 (approved):
looks good to me

PR #1987, opened 2024-02-13
Set namespace correctly for errors

thomasleese:
This is to prevent the requests being recorded in the analytics database as being part of the eligibility namespace when that might not be the case.


syed87 (approved):
lgtm

PR #1990, opened 2024-02-14
Handle sign out without a resource

thomasleese:
Sometimes the sign out might not be related to a resource, in which case we need to handle that case.

https://dfe-teacher-services.sentry.io/issues/4976448619/


syed87 (approved):
lgtm

PR #1994, opened 2024-02-15
Fix eligible region content flaky test

thomasleese:
The test relies on the application form not skipping work history so we should make that explicit in the test.


syed87 (approved):
lgtm

PR #1995, opened 2024-02-15
Capture consent method more accurately

thomasleese:
This replaces the `signed_consent_document_required` column with a `consent_method` enum which captures all possible states that the qualification request consent can be in.


syed87 (approved):
lgtm

PR #1996, opened 2024-02-15
Add unsigned_consent_document_generated column

thomasleese:
This adds a column which records when the assessor has generated the unsigned consent document.


syed87 (approved):
lgtm

PR #1997, opened 2024-02-15
Fix reference to helpers

thomasleese:
This was missed in 1e78047239c86d8ec6d4695d6d022023baa29783, the reference to the helper used to exist before the refactor.


syed87 (approved):
looks good to me

PR #2001, opened 2024-02-16
Move unsigned_consent_document_generated

thomasleese:
This moves the column to the assessment model as the unsigned consent document only has be generated once, not per qualification.


syed87 (approved):
looks good to me

PR #2006, opened 2024-02-18
Add qualifications_assessor_note column

thomasleese:
This will be used to capture any notes from the assessor when they send an application to verification by admins.


syed87 (approved):
looks good to me

PR #2007, opened 2024-02-18
Use history stack in assessor interface

thomasleese:
This changes the assessor interface to use the history stack in a similar way to the teacher interface: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/787

See https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/docs/history-and-back-links.md for more information on the history stack.


syed87 (approved):
looks good to me

PR #2008, opened 2024-02-18
Move consent letter to be on collection

thomasleese:
This changes the consent letter controller method to be on the collection since we only need to generate it once per application form, rather than per qualification request.


syed87 (approved):
looks good to me

PR #2009, opened 2024-02-18
Use UnreceiveRequestable service

thomasleese:
This changes two forms to use the UnreceiveRequestable service rather than duplicating logic.


syed87 (approved):
looks good to me

PR #2015, opened 2024-02-19
Switch to the new crown

thomasleese:
This updates the crown to use the new design. This is a manual change until we can upgrade to GOV.UK Component 5.0 which does this automatically.


syed87 (approved):
looks good to me

PR #2019, opened 2024-02-19
Remove email previews

thomasleese:
We've decided to remove these emails previews as they're not useful and sometimes show inaccurate information. Rather than fixing the previews we will remove them.


syed87 (approved):
looks good to me

PR #2020, opened 2024-02-20
Track history on each controller

thomasleese:
By including the HistoryTrackable module on the BaseController we end up sharing the origin, check and reset actions across the various controllers that inherit from the BaseController which can lead to bugs if the actions are named the same (for example any `:edit` action was being recorded as a check action because it had been defined once).

[Jira Issue](https://dfedigital.atlassian.net/browse/AQTS-173)


syed87 (approved):
looks good to me

PR #2022, opened 2024-02-20
Don't track history for documents

thomasleese:
When visiting the documents we don't want to track these in the history stack as they're opened in new tabs, and therefore once closed shouldn't have affected the history.


syed87 (approved):
lgtm

PR #1998, opened 2024-02-15
Add QualificationRequestsViewObject

thomasleese:
This adds a view object for the qualification requests.

PR #2024, opened 2024-02-21
Refactor qualifications view object

thomasleese:
This makes a couple of changes to the qualifications view object that were missed in #1998.


syed87 (approved):
looks good to me

PR #2026, opened 2024-02-21
Refactor document upload forms

thomasleese:
This refactors the existing document upload form for teachers to use a concern which can then be applied to the new document upload form that we'll be presenting to assessors.


syed87 (approved):
looks good to me

PR #2032, opened 2024-02-21
Show existing recommendation note

thomasleese:
This ensures that when navigating back to this page we show any existing recommendation note.


syed87 (approved):
looks good to me

PR #2037, opened 2024-02-23
Refactor consent requests

thomasleese:
This refactors the qualification requests to use a new model, consent requests, and removes the consent fields from the qualification requests. This should make the code easier to understand, and allows us to use the shared requestable services and timeline events for consent requests rather than needing to duplicate the logic each time.


syed87 (approved):
looks good to me

PR #2040, opened 2024-02-27
Store recommendation note on assessment

thomasleese:
This allows it to be seen in the email preview which will give confidence back to the assessors, and it ensures that we don't accidentally show a different decline reason from the session.

[Jira Issue](https://dfedigital.atlassian.net/browse/AQTS-178)


syed87 (approved):
lgtm

PR #2041, opened 2024-02-27
Prevent duplicate FIs being created

thomasleese:
With the new back links introduce in #2007 sometimes the assessors end up back on the request FI page and try to create the further information request again. This adds some safeguards to prevent duplicate FIs from being created, while also fixing the back links so assessors don't end up on the page.

[Jira Issue](https://dfedigital.atlassian.net/browse/AQTS-176)


syed87 (approved):
looks good to me

PR #2043, opened 2024-02-27
Remove convert to PDF link

thomasleese:
We're not going to have a convert to PDF link available on all uploads, only those related to consent so we can remove these from the document links.


syed87 (approved):
looks good to me

PR #2045, opened 2024-02-27
Add teaching_qualification_part_of_degree

thomasleese:
This column will record whether the teaching qualification is part of a degree or not. This used to be stored on the first qualification, but since we only ask the question once it makes sense to store it on the application form.


syed87 (approved):
looks good to me

PR #2046, opened 2024-02-27
Use teaching_qualification_part_of_degree (part 2 of 3)

thomasleese:
This changes the code to use the `teaching_qualification_part_of_degree` column which has been added to the application form for capturing whether the teaching qualification is part of the degree. We're moving away from using the `part_of_university_degree` column on the qualification as there is only one entry for this.


syed87 (approved):
looks good to me

PR #2047, opened 2024-02-27
Remove part_of_university_degree (part 3 of 3)

thomasleese:
This removes the column that is no longer used anywhere and has been replaced with the `teaching_qualification_part_of_degree` column on the application form.


syed87 (approved):
looks good to me

PR #2049, opened 2024-02-27
Validate contact email address

thomasleese:
We need to validate that the email address being requested from further information is validate to prevent incorrect values being passed in.


syed87 (approved):
looks good to me

PR #2051, opened 2024-02-28
Add review columns to consent requests

thomasleese:
We are going to be sending consent requests to review so we need to add these columns to capture any logic related to sending to review.


syed87 (approved):
looks good to me

PR #2053, opened 2024-03-01
Consent reminder email

syed87:
https://dfedigital.atlassian.net/browse/AQTS-112

When a consent document is sent to the applicant.
If it is not returned then a reminder email will be sent to the applicant after 3 weeks.


thomasleese:
Replaced by #2062 so this can be closed now.

PR #2064, opened 2024-03-12
Hide LoPS request page when not received

thomasleese:
If the LoPS has been received we should hide the page asking the applicants to request their LoPS as it's confusing.


syed87 (approved):
looks good to me

PR #2068, opened 2024-03-14
modified example data to script to randomise awarded_at/declined_at dates

nickdip:
I also made the submitted dates and created dates at an earlier date to avoid any confusion (submitted 2 months before awarded/declined, created 2 months before that)


lib/tasks/example_data.rake
```diff
@@ -247,7 +253,13 @@ def create_application_forms
           assessment,
           %i[received expired].sample,
         )
+      elsif application_form.awarded_at.present?
+        application_form.update(awarded_at: Faker::Time.between(from: 6.months.ago, to: Time.zone.now))
+        application_form.update(submitted_at: application_form.awarded_at - 2.months)
+        application_form.update(created_at: application_form.submitted_at - 2.months)
```
thomasleese, line 259: I'd suggest moving this `elsif` immediately after the decline block so they're next to each other, as they do similar things it makes sense to keep them together and keeps the code more readable.

PR #2078, opened 2024-03-20
All applicants filter

nickdip:
Added functionality to show all applications.

### DIFF
--- a/app/lib/filters/show_all_applications.rb
+++ b/app/lib/filters/show_all_applications.rb
@@ -5,7 +5,9 @@ module Filters
         ninety_days_ago = 90.days.ago
         new_scope =
           scope.where(
-            "awarded_at >= :ninety_days OR declined_at >= :ninety_days OR withdrawn_at >= :ninety_days",
+            "(awarded_at >= :ninety_days OR awarded_at IS NULL)
+          AND (declined_at >= :ninety_days OR declined_at IS NULL)
+          AND (withdrawn_at >= :ninety_days OR withdrawn_at IS NULL)",
             ninety_days: ninety_days_ago,
           )
         return new_scope
--- a/app/lib/filters/show_all_applications.rb
+++ b/app/lib/filters/show_all_applications.rb
@@ -1,7 +1,7 @@
 module Filters
   class ShowAllApplications < Base
     def apply
-      if show_all_applications.empty?
+      unless show_all_applications.include?("show_all")
         ninety_days_ago = 90.days.ago
         new_scope =
           scope.where(
--- a/app/lib/filters/show_all_applications.rb
+++ b/app/lib/filters/show_all_applications.rb
@@ -16,8 +16,8 @@ module Filters
 
     private
 
-    def show_all_applications
-      Array(params[:display]).compact_blank
+    def show_all?
+      ActiveModel::Type::Boolean.new.cast(params[:show_all])
     end
   end
 end
--- a/app/lib/filters/show_all_applications.rb
+++ b/app/lib/filters/show_all_applications.rb
@@ -1,7 +1,7 @@
 module Filters
   class ShowAllApplications < Base
     def apply
-      unless show_all_applications.include?("show_all")
+      unless show_all?
         ninety_days_ago = 90.days.ago
         return scope.where(
             "(awarded_at >= :ninety_days OR awarded_at IS NULL)
--- a/app/lib/filters/show_all_applications.rb
+++ b/app/lib/filters/show_all_applications.rb
@@ -3,12 +3,14 @@ module Filters
     def apply
       unless show_all?
         ninety_days_ago = 90.days.ago
-        return scope.where(
+        return(
+          scope.where(
             "(awarded_at >= :ninety_days OR awarded_at IS NULL)
           AND (declined_at >= :ninety_days OR declined_at IS NULL)
           AND (withdrawn_at >= :ninety_days OR withdrawn_at IS NULL)",
             ninety_days: ninety_days_ago,
           )
+        )
       end
 
       scope
--- a/app/lib/filters/show_all_applications.rb
+++ b/app/lib/filters/show_all_applications.rb
@@ -1,0 +1,21 @@
+module Filters
+  class ShowAllApplications < Base
+    def apply
+        Rails.logger.debug "DEBUG showallapplications #{params.inspect}"
+
+        if show_all_applications.empty?
+            ninety_days_ago = 90.days.ago
+            new_scope = scope.where("awarded_at >= :ninety_days OR declined_at >= :ninety_days OR withdrawn_at >= :ninety_days", ninety_days: ninety_days_ago)
+            return new_scope
+        end
+
+        return scope
+    end
+
+    private
+
+    def show_all_applications
+       Array(params[:display]).compact_blank
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


app/lib/filters/show_all_applications.rb
```diff
@@ -0,0 +1,23 @@
+module Filters
+  class ShowAllApplications < Base
+    def apply
+      if show_all_applications.empty?
```
thomasleese, line 4:
```suggestion
      unless show_all_applications.include?("show_all")
```

This value `show_all` comes from the value of the checkbox: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2078/files#diff-14c048dcf48c873d6deb7ce9a22b554330cfadcac4a8e1c805b480f593b15983R69

I think your solution works too but I prefer this as it's more explicit and allows us to add more checkboxes in the future.

nickdip:
I had to do some changes with the filtering, hopefuly it makes sense and works. I also ended up using a scope that removed draft applications

app/models/application_form.rb
```diff
@@ -162,15 +162,7 @@ class ApplicationForm < ApplicationRecord
 
   STATUS_COLUMNS.each { |column| enum column, STATUS_VALUES, prefix: column }
 
-  scope :assessable, -> { where.not(stage: %i[draft completed]) }
```
thomasleese, line 165: I think this scope is used in a number of other places so we'll need to keep it:

https://github.com/search?q=repo%3ADFE-Digital%2Fapply-for-qualified-teacher-status%20assessable&type=code

app/models/application_form.rb
```diff
@@ -162,15 +162,7 @@ class ApplicationForm < ApplicationRecord
...
-
-  scope :active,
-        -> do
-          assessable
-            .or(where("awarded_at >= ?", 90.days.ago))
-            .or(where("declined_at >= ?", 90.days.ago))
-            .or(where("withdrawn_at >= ?", 90.days.ago))
-        end
+  scope :submitted, -> { where.not(stage: %i[draft]) }
```
thomasleese, line 165:
```suggestion
  scope :submitted, -> { where.not(stage: "draft") }
```

Just a minor thing.

app/models/application_form.rb
nickdip: Oh yes...that's fixed now

app/views/assessor_interface/application_forms/index.html.erb
```diff
@@ -63,6 +63,12 @@
         <% end %>
       <% end %>
     </section>
+
+    <section id="app-applications-show-all-applicants">
+      <%= f.govuk_check_boxes_fieldset :display, legend: { size: "s" }, small: true do %>
+        <%= f.govuk_check_box :display, :show_all, label: { text: "All applications" }, checked: @view_object.filter_form.display&.include?("show_all") %>
```
thomasleese, line 69:
```suggestion
        <%= f.govuk_check_box :display, :show_all, label: { text: "Show applications completed older than 90 days ago" }, checked: @view_object.filter_form.display&.include?("show_all") %>
```

spec/view_objects/assessor_interface/application_forms_index_view_object_spec.rb
```diff
@@ -149,8 +159,6 @@
   end
 
   describe "#country_filter_options" do
-    subject(:country_filter_options) { view_object.country_filter_options }
-
```
thomasleese, line 153: Did you mean to remove this line?

app/views/assessor_interface/application_forms/index.html.erb
thomasleese, line 69:
```suggestion
      <%= f.govuk_check_boxes_fieldset :display, legend: nil, small: true do %>
        <%= f.govuk_check_box :display, :show_all, label: { text: "All applications" }, checked: @view_object.filter_form.display&.include?("show_all") %>
```

I think we can hide the label here.

thomasleese, line 69:
```suggestion
      <%= f.govuk_check_boxes_fieldset :show_all, legend: { size: "s" }, multiple: false, small: true do %>
        <%= f.govuk_check_box :show_all, :true, multiple: false, label: { text: "All applications" }, checked: @view_object.filter_form.show_all == "true" %>
```

Thinking about the naming of these fields, I think it makes sense to stick with the `show_all` name as `display` is a little too generic in my opinion, unclear exactly what that might be.

app/lib/filters/show_all_applications.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      scope
+    end
+
+    private
+
+    def show_all_applications
+      Array(params[:display]).compact_blank
+    end
```
thomasleese, line 23: Following on from my suggestion about renaming the check box name.

```suggestion
    def show_all?
      ActiveModel::Type::Boolean.new.cast(params[:show_all])
    end
```

`ActiveModel::Type::Boolean.new.cast` parses the string for us so the string `"true"` is treated as a `true` boolean.

app/lib/filters/show_all_applications.rb
```diff
@@ -0,0 +1,25 @@
...
+      unless show_all_applications.include?("show_all")
+        ninety_days_ago = 90.days.ago
+        new_scope =
+          scope.where(
+            "(awarded_at >= :ninety_days OR awarded_at IS NULL)
+          AND (declined_at >= :ninety_days OR declined_at IS NULL)
+          AND (withdrawn_at >= :ninety_days OR withdrawn_at IS NULL)",
+            ninety_days: ninety_days_ago,
+          )
+        return new_scope
```
thomasleese, line 13:
```suggestion
        return scope.where(
            "(awarded_at >= :ninety_days OR awarded_at IS NULL)
          AND (declined_at >= :ninety_days OR declined_at IS NULL)
          AND (withdrawn_at >= :ninety_days OR withdrawn_at IS NULL)",
            ninety_days: ninety_days_ago,
          )
```

I don't think we need to define a variable here.

thomasleese (commented):
Overall looks good, just get a couple of small comments left.

app/views/assessor_interface/application_forms/index.html.erb
nickdip: Makes sense - I think I'll need to adapt this in the spec too

spec/view_objects/assessor_interface/application_forms_index_view_object_spec.rb
nickdip: no!

app/models/application_form.rb
```diff
@@ -164,13 +164,9 @@ class ApplicationForm < ApplicationRecord
...
   scope :assessable, -> { where.not(stage: %i[draft completed]) }
 
-  scope :active,
-        -> do
-          assessable
-            .or(where("awarded_at >= ?", 90.days.ago))
-            .or(where("declined_at >= ?", 90.days.ago))
-            .or(where("withdrawn_at >= ?", 90.days.ago))
-        end
+  scope :assessable, -> { where.not(stage: %i[draft completed]) }
```
thomasleese, line 167: It looks like we've got two `assessable` scopes defined now.

thomasleese (commented):
One comment left otherwise it looks good!

config/feature_flags.yml
```diff
@@ -34,4 +34,9 @@ feature_flags:
       Allow starting an application on this service directly after completing
       an eligibility check.
 
+  show_all_applicants:
```
thomasleese, line 37:
```suggestion
  show_all_applications_filter:
```

Just a suggestion to make the name a little more explicit.

thomasleese (approved):

PR #2082, opened 2024-03-20
Add route and controller for tracking support email link and update o…

syed87:
[…ne link](https://dfedigital.atlassian.net/browse/AQTS-228)

This feature will allow us to track which pages users are requesting support from via the support email

### DIFF
--- a/app/controllers/static_controller.rb
+++ b/app/controllers/static_controller.rb
@@ -2,4 +2,9 @@
 
 class StaticController < ApplicationController
   include EligibilityCurrentNamespace
+
+  def email
+    address = I18n.t(params[:key], scope: %i[service email])
+    redirect_to "mailto:#{address}", allow_other_host: true
+  end
 end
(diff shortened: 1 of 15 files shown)


app/controllers/tracking_controller.rb
```diff
@@ -0,0 +1,5 @@
+class TrackingController < ApplicationController
+  def mailto_clicks
+    redirect_to "mailto:#{params[:email]}"
```
thomasleese, line 3: Rather than being open to any email address I wonder if we should restrict it to the email addresses that we know about:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/84c2ac88fa59ed7738cc8b11e733b4bd4ca58df8/config/locales/en.yml#L14-L16

```suggestion
    redirect_to "mailto:#{t(params[:email], scope: %i[service email])}"
```

And then we'd use it in the various controllers like this:

```rb
<%= govuk_link_to t("service.email.enquiries"), "/track_mailto/enquiries" %>
```

I don't like the idea of the controller method being able to be used to send emails to any address rather than just ours for security reasons.

thomasleese (approved):

PR #2086, opened 2024-03-22
Add spacing on Ecctis response page

thomasleese:
This adds some spacing that was missing from the implementation but is there in the design.


syed87 (approved):
lgtm

PR #2132, opened 2024-04-09
Disable assessment forms after assessment

thomasleese:
When a recommendation has been made by an assessor we should disable the forms to prevent them being modified after an assessment has been made. This is being made as we found that sometimes assessors were making changes after making a recommendation and expecting it to update the assessment, for example request more further information.

This is related to https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2114 which made this clearer in the timeline.


syed87 (approved):
looks good to me

PR #2149, opened 2024-04-15
Sync country, region and english language provider configuration

thomasleese:
This adds new Rake tasks and a GitHub Actions workflow to support syncing specific information between production and the testing environments. In this case we want to sync the countries, the regions and the english language providers to ensure that the test environments appear to show what would show in production, making it a better testing environment.


syed87 (approved):
looks good to me

PR #2151, opened 2024-04-16
Ensure draft applications are filled out

thomasleese:
This changes how the fake data is generated to ensure that it's pre-populated as this is more useful than having an empty form. If we want to generate an empty form we can go via the eligibility checkers.


syed87 (approved):
looks good to me

PR #2152, opened 2024-04-17
Fix issue with configuration sync workflow

thomasleese:
This ensures that we only run the import jobs after the export job has completed and fixes the incorrect names.


syed87 (approved):
looks good to me

saliceti:
[AFQTS: bug with google service account](https://trello.com/c/OZuxmTFB/2152-afqts-bug-with-google-service-account)

PR #2164, opened 2024-04-24
Track subject limited on application forms

thomasleese:
This tracks whether an application form is subject limited on each individual application form rather than going out to the country, where the value may have changed since the application was submitted.


syed87 (approved):
looks good to me

PR #2166, opened 2024-04-24
Don't install Microsoft packages

thomasleese:
This fixes an issue with the Microsoft repository that's not working currently, and since we don't use any Microsoft packages we can skip them.

https://github.com/orgs/community/discussions/120966


syed87 (approved):
looks good to me

PR #2167, opened 2024-04-24
Ensure we use the correct requires_preliminary_check

thomasleese:
Previously we were taking the value from the application form rather than taking the latest version from the region.


thomasleese:
Merging this in to fix a bug.

PR #2168, opened 2024-04-24
Don't mark contact_job as PII

thomasleese:
We don't mark it as PII in the work history itself so we don't need it to be PII in the further information.


syed87 (approved):
looks good to me

PR #2172, opened 2024-04-25
Add task for changing qualification certificate dates

thomasleese:
This adds a Rake task that makes it possible to change the certificate date of a qualification and record that in the timeline as requested by assessors.


syed87 (approved):
looks good to me

PR #2173, opened 2024-04-25
Improve how regional content is shown

thomasleese:
This standardises how we show region content and ensures that any content related to the written statement (from the status or the sanction checks) are shown in the request professional standing certificate page.


syed87 (approved):
looks good to me

PR #2175, opened 2024-04-25
Refactor date/time formatting

thomasleese:
This refactors how we format dates and times to have a default formatter (for dates and times) and any additional custom formatting where necessary. This reduces the need to pick a formatter, instead we can use `to_fs` without any arguments.


syed87 (approved):
looks good to me

PR #2186, opened 2024-05-01
Fix flaky date formatting test

thomasleese:
This test works until the day of the month is only a single character where we're checking for an empty space before the day which doesn't exist.


syed87 (approved):
looks good to me

PR #2197, opened 2024-05-09
Increase frequency of DQT duplicate checks

thomasleese:
At the moment we only check for a match in DQT when an application is submitted and then never again. We find that some applicants end up in DQT via other means while they're waiting for their application to be assessed, so we'd like to start checking DQT more frequently. This approach will check DQT every 24 hours per applications until the application is either awarded, declined or withdrawn.

I've also refactored the job which performs to checks.


syed87 (approved):
looks good to me

PR #2199, opened 2024-05-10
Update bank holidays

thomasleese:
This updates the bank holidays to ensure they're correct and adds a test to make sure we stay on top of them in the future.


syed87:
Probably unnecessary but there is a bank holiday api which could be used if we don't want to keep updating the yml file

syed87 (approved):
looks good to me

PR #2200, opened 2024-05-10
Add analytics publication extract task

thomasleese:
This adds a task for extracting the publication data for comparision with the data from BigQuery. I haven't added in tests as we'll only need this temporarily.


syed87 (approved):
lgtm

PR #2205, opened 2024-05-15
Fix date range in analytics publication extract

thomasleese:
We were missing applications from the last day as we had set the cut off at midnight rather than 23:59:59, so now we convert the value to a date first.


syed87 (approved):
lgtm

PR #2207, opened 2024-05-16
Limit data publication to submitted applications

thomasleese:
This is to match the SQL from the data publication query.


syed87 (approved):
looks good to me

PR #2211, opened 2024-05-16
Check assessment subjects rather than applicant subjects

thomasleese:
We need to check the subjects that the assessors are selecting rather than ones provided by the applicant.


syed87 (approved):
lgtm

PR #2213, opened 2024-05-16
Only check subjects for awards

thomasleese:
The previous change still doesn't quite match the SQL query we run against BigQuery, so instead going with a different approach.


syed87 (approved):
lgtm

PR #2226, opened 2024-05-22
Always record assessment section timeline events

thomasleese:
When assessment sections are assessed we want to always record a timeline event to ensure this is captured even if the status changes, but the selected failure reasons might be different.

I've also refactored the service class to make it simpler.


syed87 (approved):
lgtm

PR #2227, opened 2024-05-22
Update content related to applicant already being in DQT

thomasleese:
This updates the content we show to teachers when they've already got a match in DQT as requested by the TRA.


syed87 (approved):
lgtm

thomasleese:
This will be picked up again later.

PR #2230, opened 2024-05-27
Hide qualification information except on teaching qualification

thomasleese:
We only want it to be shown on the form for the teaching qualification, not the bachelors degree or additional qualification as the content added to that field is irrelevant to those contexts.


syed87 (approved):
lgtm

PR #2235, opened 2024-05-31
Set Rails Semantic Logger format to JSON

thomasleese:
The configuration `log_format` doesn't exist and therefore it wasn't being used anywhere, instead we need to specify the logging format when we add the log appender. In production we want this to be a JSON log so it can be parsed by Logit.


syed87 (approved):
lgtm

PR #2239, opened 2024-06-04
Update publication extract calculation

thomasleese:
This updates the publication extract according to the new calculations where we only consider the submitted applications on a certain date range, regardless of when they transitioned to a different state.


syed87 (commented):
lgtm

PR #2243, opened 2024-06-05
Reset requested date when changing reference request email addresses

thomasleese:
This updates the `UpdateWorkHistoryContact` service to re-request the reference request (if there is one) so the expiration date is shifted by 6 weeks, allowing the new contact enough time to respond to the reference.


syed87 (approved):
looks good to me

PR #2248, opened 2024-06-07
Simplify back links in assessor interface

thomasleese:
This removes the use of the HistoryTrackable module, where possible, as it causes issues with accessing the page multiple times in parallel. The history is stored in the session which ends up being shared between tabs, causing problems with the back links. Instead, for the assessor interface, we can hard code the majority of back links ensuring they're consistent and go the correct place, and leaving the use of HistoryTrackable only where strictly necessary.


syed87 (approved):
lgtm

PR #2251, opened 2024-06-11
Add age and subject restriction content

thomasleese:
This adds some content for countries which have an age and subject restriction, shown on both the qualifications form and the eligibility checker "proof of qualifications" section. Currently this content is included in the teaching qualifications content in the support console which will need to be removed as this gets deployed.


syed87 (approved):
lgtm

PR #2254, opened 2024-06-12
Expire requestable synchronously

thomasleese:
This fixes an issue where if more than one requestable happened to expire at the same time, there is a race condition between when the status and action required at fields are updated, leading to incorrect application statuses.


syed87 (approved):
looks good to me

PR #2255, opened 2024-06-12
Refactor support region and country forms

thomasleese:
This ensures that we use a form to manage the region information and the country information, which removes a number of unnecessary methods from the country and region models. It also fixes a number of issues where the forms aren't setting values correctly.


syed87 (approved):
looks good to me

PR #2277, opened 2024-07-02
Add validation on maximum age

thomasleese:
Currently we allow the assessors to choose any age range. This causes problems when the application is sent to DQT as the API requires the maximum age to be 19 or less.


thomasleese:
We've decided not to implement this for now.

PR #2281, opened 2024-07-03
Update feedback form

thomasleese:
This updates the feedback form links and content to consolidate them all in to a single feedback form for all users.


syed87 (approved):
lgtm

PR #2282, opened 2024-07-03
Set Sidekiq log level to warning

thomasleese:
This sets the log level for Sidekiq to warning to avoid sending lots of unnecessary generic logging, when we can track these in other ways via the Sidekiq dashboard. Instead, we only need to send logs about errors or problems that might need attention.


saliceti (approved):

saliceti:
[Remove dfe analytics logs from Logit.io](https://trello.com/c/fxSQSbA6/1907-remove-dfe-analytics-logs-from-logitio)

PR #2288, opened 2024-07-05
Database table created to store info for failed emails

syed87:
This is the initial part of AQTS-230

https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-230

### DIFF
--- a/app/models/mail_delivery_failure.rb
+++ b/app/models/mail_delivery_failure.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # == Schema Information
 #
 # Table name: mail_delivery_failures
--- a/app/models/mail_delivery_failure.rb
+++ b/app/models/mail_delivery_failure.rb
@@ -1,0 +1,13 @@
+# == Schema Information
+#
+# Table name: mail_delivery_failures
+#
+#  id                   :bigint           not null, primary key
+#  email_address        :string
+#  mailer_action_method :string
+#  mailer_class         :string
+#  created_at           :datetime         not null
+#  updated_at           :datetime         not null
+#
+class MailDeliveryFailure < ApplicationRecord
+end
--- a/app/models/mail_delivery_failure.rb
+++ b/app/models/mail_delivery_failure.rb
@@ -5,11 +5,15 @@
 # Table name: mail_delivery_failures
 #
 #  id                   :bigint           not null, primary key
-#  email_address        :string
-#  mailer_action_method :string
-#  mailer_class         :string
+#  email_address        :string           not null
+#  mailer_action_method :string           not null
+#  mailer_class         :string           not null
 #  created_at           :datetime         not null
 #  updated_at           :datetime         not null
+#
+# Indexes
+#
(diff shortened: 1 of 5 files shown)


db/migrate/20240705082709_create_mail_delivery_failures.rb
```diff
@@ -0,0 +1,11 @@
+class CreateMailDeliveryFailures < ActiveRecord::Migration[7.1]
+  def change
+    create_table :mail_delivery_failures do |t|
+      t.string :email_address
+      t.string :mailer_class
+      t.string :mailer_action_method
```
thomasleese, line 6: I think we'll want these to all be `null: false` to ensure they can't be set to `NULL`, as we'd expect them always to be set.

db/migrate/20240705082709_create_mail_delivery_failures.rb
```diff
@@ -0,0 +1,11 @@
+class CreateMailDeliveryFailures < ActiveRecord::Migration[7.1]
+  def change
+    create_table :mail_delivery_failures do |t|
+      t.string :email_address
```
thomasleese, line 4: I think we should add an index on the `email_address` column to so that when we're searching for these records we get results quickly.

https://apidock.com/rails/ActiveRecord/ConnectionAdapters/Table/index

https://stackoverflow.com/questions/1108/how-does-database-indexing-work

db/schema.rb
```diff
@@ -182,6 +182,14 @@
     t.index ["code"], name: "index_countries_on_code", unique: true
   end
 
+  create_table "create_mail_delivery_failures", force: :cascade do |t|
```
thomasleese, line 185: Looks like there are some older versions of the table here in this file which will need to be removed.

thomasleese (commented):
Just got a few comments, but overall looks good to me, will be nice to get this merged in early.

app/models/mail_delivery_failure.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: mail_delivery_failures
+#
+#  id                   :bigint           not null, primary key
+#  email_address        :string
+#  mailer_action_method :string
+#  mailer_class         :string
```
thomasleese, line 10: It looks like these don't match the table, the comments are generated automatically when `db:migrate` is run but they should refer to these fields being `not null`.

db/migrate/20240705082709_create_mail_delivery_failures.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    create_table :mail_delivery_failures do |t|
+      t.string :email_address, null: false
+      t.string :mailer_class, null: false
+      t.string :mailer_action_method, null: false
+
+      t.timestamps
+    end
+
+    add_index :mail_delivery_failures, :email_address
```
thomasleese, line 13: This works, but I think you can do `t.index :email_address` within the `create_table` block.

thomasleese (approved):

PR #2289, opened 2024-07-05
Require a note after further information requests

thomasleese:
If declining after a further information request, we want to require that the assessors write in a recommendation note as otherwise the applicant will see no reason for their decline.


syed87 (approved):
lgtm

PR #2291, opened 2024-07-05
Remove line about the application form

thomasleese:
This removes a line on the page when starting a new application which isn't necessary, as requested by our content designer.


syed87 (approved):
lgtm

PR #2293, opened 2024-07-09
Entity-relationship diagram

thomasleese:
This adds an entity-relationship diagram for the models, and sets it up so that new diagrams will be generated each time the migrations are run.


syed87 (approved):
lgtm

PR #2294, opened 2024-07-09
Added maintenance window for prod postgress

VenkataChalla554:
Terraform plan 
 

      + maintenance_window {
          + day_of_week  = 0
          + start_hour   = 3
          + start_minute = 0
        }

### DIFF
--- a/terraform/application/config/production/variables.tfvars.json
+++ b/terraform/application/config/production/variables.tfvars.json
@@ -1,6 +1,11 @@
 {
   "app_environment": "production",
   "app_replicas": 2,
+  "azure_maintenance_window": {
+    "day_of_week": 0,
+    "start_hour": 2,
+    "start_minute": 0
+  },
   "cluster": "production",
   "dqt_api_url": "https://teacher-qualifications-api.education.gov.uk",
   "enable_logit": true,
--- a/terraform/application/databases.tf
+++ b/terraform/application/databases.tf
@@ -26,7 +26,8 @@ module "postgres" {
 
   cluster_configuration_map = module.cluster_data.configuration_map
 
-  use_azure               = var.deploy_azure_backing_services
-  azure_enable_monitoring = var.enable_monitoring
-  azure_extensions        = ["pg_stat_statements", "pgcrypto"]
+  use_azure                = var.deploy_azure_backing_services
+  azure_enable_monitoring  = var.enable_monitoring
+  azure_extensions         = ["pg_stat_statements", "pgcrypto"]
+  azure_maintenance_window = var.azure_maintenance_window
 }
--- a/terraform/application/variables.tf
+++ b/terraform/application/variables.tf
@@ -17,6 +17,15 @@ variable "azure_resource_prefix" {
   type = string
 }
 
+variable "azure_maintenance_window" {
+  type = object({
+    day_of_week  = number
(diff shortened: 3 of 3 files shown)


terraform/application/config/production/variables.tfvars.json
```diff
@@ -7,5 +7,10 @@
   "enable_monitoring": true,
   "external_hostname": "apply-for-qts-in-england.education.gov.uk",
   "namespace": "tra-production",
-  "uploads_storage_environment_tag": "Production"
+  "uploads_storage_environment_tag": "Production",
+  "azure_maintenance_window": {
+    "day_of_week": 0,
+    "start_hour": 3,
+    "start_minute": 0
+  }
```
thomasleese, line 15: Can we keep these variables in alphabetical order to make them easier to find?

terraform/application/.terraform.lock.hcl
thomasleese, line 1: Could we upgrade these dependencies in a separate commit (or preferably a separate PR)? Then we can deploy those on their own first, and it keeps this PR related only to the maintenance window changes.

thomasleese (commented):
Thanks for this, could you add a description to the commit message explaining why this time has been chosen as the maintenance window and what that means?

It looks like there's also a linting issue that needs fixing.

terraform/application/config/production/variables.tfvars.json
VenkataChalla554: fixed

terraform/application/variables.tf
```diff
@@ -17,6 +17,8 @@ variable "azure_resource_prefix" {
   type = string
 }
 
+variable "azure_maintenance_window" { default = null }
```
thomasleese, line 20:
```suggestion
variable "azure_maintenance_window" {
  type = object({
      day_of_week  = number
      start_hour   = number
      start_minute = number
  })
  default = null
}
```

Just to make sure we always pass this in correctly.

VenkataChalla554: ya it make sense , updated the codes.

terraform/application/databases.tf
```diff
@@ -29,4 +29,6 @@ module "postgres" {
   use_azure               = var.deploy_azure_backing_services
   azure_enable_monitoring = var.enable_monitoring
   azure_extensions        = ["pg_stat_statements", "pgcrypto"]
+
+  azure_maintenance_window = var.azure_maintenance_window
```
thomasleese, line 33:
```suggestion
  use_azure                = var.deploy_azure_backing_services
  azure_enable_monitoring  = var.enable_monitoring
  azure_extensions         = ["pg_stat_statements", "pgcrypto"]
  azure_maintenance_window = var.azure_maintenance_window
```

Just one minor thing, could we keep all the azure variables together so they're easier to see as a group.

thomasleese (commented):
Sorry just some comment left, otherwise it's ready to go!

terraform/application/databases.tf
VenkataChalla554: This is fixed .

VenkataChalla554:
[add prod postgres maintenance window](https://trello.com/c/4NPqv8MF/1903-add-prod-postgres-maintenance-window)

thomasleese (approved):

PR #2300, opened 2024-07-11
Add suitability record models

thomasleese:
This adds a migration and the models related to the new suitability records.


syed87 (approved):
lgtm

PR #2301, opened 2024-07-11
Prevent users from creating multiple applications

thomasleese:
This ensures that if there's an existing application form, it's in the completed stage before an applicant can submit another one.


syed87 (approved):
lgtm

PR #2304, opened 2024-07-12
Add suitability records section to header

thomasleese:
This starts the work on the suitability records by adding a link in the main navigation bar to where the suitability records section of the site will live.


syed87 (approved):
lgtm

PR #2308, opened 2024-07-16
Record who created a suitability record

thomasleese:
This adds a field on the suitability records that we can use to track who created the record.


syed87 (approved):
lgtm

PR #2309, opened 2024-07-16
Add ability to show suitability records

thomasleese:
This adds the ability to show the suitability records including pagination and summary list for each record.


syed87 (approved):
lgtm

PR #2311, opened 2024-07-17
AQTS230 Save failed emails to MailDeliveryFailure table

syed87:
[ticket](https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-230)

MailDeliveryFailure model has been merged into the codebase, this pr will provide the functionality to create entries for the table by recording errors when sending emails.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -15,16 +15,16 @@ class ApplicationMailer < Mail::Notify::Mailer
     #
     # @see https://github.com/rails/rails/issues/39018
     if respond_to?(:headers)
-      recipient_email = headers["To"].value
-      MailDeliveryFailure.create!(recipient_email:)
+      email_address = headers["To"].value
+      mailer_action_method = action_name
+      mailer_class = mailer_name
+      MailDeliveryFailure.create!(email_address:, mailer_action_method:, mailer_class:)
     end
 
     raise
   end
 
   def notify_email(headers)
-    headers =
-      headers.merge(rails_mailer: mailer_name, rails_mail_template: action_name)
     view_mail(GOVUK_NOTIFY_TEMPLATE_ID, headers)
   end
 end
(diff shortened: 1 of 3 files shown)


app/views/anonymous/test_notify_error.text.erb
thomasleese, line 1: Do we need this file? I can't see where it's used.

spec/mailers/application_mailer_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ApplicationMailer, :sidekiq do
```
thomasleese, line 5:
```suggestion
RSpec.describe ApplicationMailer, type: :mailer do
```

app/mailers/application_mailer.rb
```diff
@@ -8,4 +8,27 @@ class ApplicationMailer < Mail::Notify::Mailer
...
+        email_address:,
+        mailer_action_method:,
+        mailer_class:,
+      )
+    end
+
+    raise
+  end
+
+  def notify_email(headers)
```
thomasleese, line 31: Do we need this method here? It looks like a convenience method for calling `view_mail` with the template ID already provided, but it's not used anywhere.

thomasleese (commented):
A few small comments, but otherwise looks good.

app/mailers/application_mailer.rb
syed87: It was originally going to add the mailer_action_method and mailer_class to the header but that was done automatically. However It's still used in the application mailer spec on line 15 to set some dummy info.

app/views/anonymous/test_notify_error.text.erb
syed87: Screenshot 2024-08-19 at 10 24 32


I get this error if it's removed, the spec file needs to have this view for it to work.

thomasleese: Hmm okay interesting, it's a pity because it's a view that's only used in the tests, but maybe we need it.

Hassanmir92 (approved):

PR #2326, opened 2024-07-22
Add migration to add suitability decline reasons

thomasleese:
When we're ready to go live with the new suitability reasons we can deploy this migration to apply it to existing application forms.


syed87 (approved):
lgtm

thomasleese:
I've applied this change manually so we no longer need the migration.

PR #2334, opened 2024-07-28
Add validation on minimum file size

thomasleese:
We're seeing some files being uploaded empty and we think this could be due to applicants uploading these files, so we expect this validation will fix this problem.


syed87 (approved):
lgtm

PR #2335, opened 2024-07-28
Add reference request error page

thomasleese:
This adds a page which is shown to referees if they try and access a link that is no longer valid.


syed87 (approved):
lgtm

PR #2354, opened 2024-08-13
Send recommendation assessor note to BigQuery

thomasleese:
We want to do analysis on this value so we should move it out of the blocklist and in to the hidden PII configuration.


syed87 (approved):
lgtm

PR #2370, opened 2024-08-20
Prevent changing work history for received reference

thomasleese:
If the reference has already been received it shouldn't be possible to change the name of the contact as it would then be incorrect, suggesting a different person responded to the reference.


syed87 (approved):
lgtm

PR #2374, opened 2024-08-22
Don't hide decline reasons if expired

thomasleese:
If the further information request or LoPS has expired, in most cases this leads to an automatic decline. However, it's possible to rollback this decision and allow the assessor to continue with the assessment. Currently if the assessor then goes on to decline the application manually, these reasons will not be shown.


syed87 (approved):
lgtm

PR #2376, opened 2024-08-23
Update links to privacy policy

thomasleese:
We'll be switching to a DfE-wide privacy policy that covers our service so we don't need to have one specific to the service anymore.


syed87 (approved):
lgtm

PR #2379, opened 2024-08-23
Add teacher license form for Ghana

thomasleese:
### Description

The Registration number, also known as the license number for Ghanaian specific applicants is a 3 part alphanumeric which has a specific format. To make it easier for applicants to not make mistakes when entering this number, we have decided to change the UI with additional validations in place to guide users.

Additionally, the application form status updater has also been updated with regards to the status of the registration form. If a Ghanaian applicant, submits their teacher license/registration number and it is not valid, the status should be "in_progress", meanwhile if it is empty, it should be "not_started". If it is valid then only then it will move to "completed"

Important Design Note: 
As part of this work a new grouped input had to be introduced with inspiration taken from the GOV design systems date field. Since, this is a custom group of inputs, there are no helpers available and as a result of this, custom input fields with GOV design system classes have been used to replicate the desired design.

Ghana Specific Form:
Screenshot 2024-09-03 at 08 43 09

Ghana Specific Form with Errors:
Screenshot 2024-09-03 at 08 43 30

Other:
Screenshot 2024-09-03 at 08 42 28

### DIFF
--- a/app/views/teacher_interface/registration_number/edit_ghana.html.erb
+++ b/app/views/teacher_interface/registration_number/edit_ghana.html.erb
@@ -1,0 +1,37 @@
+<% content_for :page_title, title_with_error_prefix(t("application_form.tasks.items.registration_number.ghana"), error: @form.errors.any?) %>
+<% content_for :back_link_url, back_history_path(default: teacher_interface_application_form_path) %>
+
+<%= form_with model: @form, url: %i[registration_number teacher_interface application_form] do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= f.govuk_text_field :registration_number, label: { size: "l", tag: "h1" } %>
+
+  <% if @application_form.region.status_check_online? %>
+    <% if @application_form.region.status_information.present? %>
+      <%= raw GovukMarkdown.render(@application_form.region.status_information) %>
+    <% end %>
+
+    <% if @application_form.region.country.status_information.present? %>
+      <%= raw GovukMarkdown.render(@application_form.region.country.status_information) %>
+    <% end %>
+  <% end %>
+
+  <% if @application_form.region.sanction_check_online? %>
+    <% if @application_form.region.sanction_information.present? %>
+      <%= raw GovukMarkdown.render(@application_form.region.sanction_information) %>
+    <% end %>
+
+    <% if @application_form.region.country.sanction_information.present? %>
+      <%= raw GovukMarkdown.render(@application_form.region.country.sanction_information) %>
+    <% end %>
+  <% end %>
+
+  <%= govuk_details(summary_text: "When we might need more information") do %>
+    <p>
+      If we’re unable to find you using your registration number, we may need to
+      contact you for more information as part of your application.
+    </p>
+  <% end %>
+
+  <%= render "shared/save_submit_buttons", f: %>
+<% end %>
--- a/app/views/teacher_interface/registration_number/edit_ghana.html.erb
+++ b/app/views/teacher_interface/registration_number/edit_ghana.html.erb
@@ -4,34 +4,74 @@
 <%= form_with model: @form, url: %i[registration_number teacher_interface application_form] do |f| %>
   <%= f.govuk_error_summary %>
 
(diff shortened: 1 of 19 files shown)


app/views/teacher_interface/registration_number/edit_ghana.html.erb
```diff
@@ -0,0 +1,77 @@
+<% content_for :page_title, title_with_error_prefix(t("application_form.tasks.items.registration_number.ghana"), error: @form.errors.any?) %>
+<% content_for :back_link_url, back_history_path(default: teacher_interface_application_form_path) %>
+
+<%= form_with model: @form, url: %i[registration_number teacher_interface application_form] do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <div class="govuk-form-group <%= 'govuk-form-group--error' if @form.errors.present? %>">
```
Hassanmir92, line 7: Note that this form group had to be done in a custom way since this form component does not exist. However, we have reused the date field component and ensured it matches the design system.

app/views/teacher_interface/registration_number/edit_ghana.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    <% if @application_form.region.sanction_information.present? %>
+      <%= raw GovukMarkdown.render(@application_form.region.sanction_information) %>
+    <% end %>
+
+    <% if @application_form.region.country.sanction_information.present? %>
+      <%= raw GovukMarkdown.render(@application_form.region.country.sanction_information) %>
+    <% end %>
+  <% end %>
+
+  <%= govuk_details(summary_text: "When we might need more information") do %>
```
syed87: I'm wondering if this section is still necessary for the Ghana view. I didn't see it in the screenshot. Do we need to keep it here?

app/views/teacher_interface/registration_number/edit_ghana.html.erb
```diff
@@ -0,0 +1,37 @@
+<% content_for :page_title, title_with_error_prefix(t("application_form.tasks.items.registration_number.ghana"), error: @form.errors.any?) %>
+<% content_for :back_link_url, back_history_path(default: teacher_interface_application_form_path) %>
+
+<%= form_with model: @form, url: %i[registration_number teacher_interface application_form] do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= f.govuk_text_field :registration_number, label: { size: "l", tag: "h1" } %>
+
+  <% if @application_form.region.status_check_online? %>
```
syed87: I can't see any issues with this but could we go through status_check_online and status_information just for my own understanding. I'm not sure what these are and how they relate to the workflow of the website.

config/locales/helpers.en.yml
```diff
@@ -223,6 +225,8 @@ en:
           provider: Upload proof of a Secure English Language Test (SELT) at B2 level, from one of the approved providers.
       teacher_interface_further_information_request_item_text_form:
         response: Enter your response
+      teacher_interface_ghana_registration_number_form:
+        registration_number: "Enter your teacher license number"
```
syed87, line 229: I noticed that the brackets are used in this section. It might be more consistent to remove them in this section.

app/views/teacher_interface/registration_number/edit_ghana.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    <% if @application_form.region.status_information.present? %>
+      <%= raw GovukMarkdown.render(@application_form.region.status_information) %>
```
syed87: I know this code get's replaced later on but could we go through how status_information and sanction_information work and how they're used in the edit view and website in general.

Hassanmir92: Of course. Lets maybe jump on a call and go through it?

config/locales/helpers.en.yml
Hassanmir92: Actioned: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2379/commits/38003732d1376f7f95196f7ad406a0ebd6b3cf7d

syed87 (approved):
looks good to me

PR #2386, opened 2024-09-03
Temporarily skip tests for inviting staff with azure active directory

Hassanmir92:
This is hopefully to be reverted soon. However, this specific test has started failing without impact from any new work and blocks our test suit. 

I have tested this feature on production and it has been working fine and we have had no reports of it not working since the issue started impacting our tests on 29th August 2024.


syed87 (approved):
looks good to me

PR #2387, opened 2024-09-03
Ensure emails sent in development appear in tmp/mails folder

Hassanmir92:
In development, this ensures that mails sent go into `/tmp/mail` folder for each specific email address.


syed87 (approved):
looks good to me

PR #2389, opened 2024-09-03
SPike: Enable GCP WIF on worker

saliceti:
Keep until it's implemented in the terraform module
https://trello.com/c/cQwpKIWx


saliceti:
Implemented in https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2441
Closing

PR #2401, opened 2024-09-13
[AQTS-551] GovOne Login Authentication

Hassanmir92:
https://dfedigital.atlassian.net/browse/AQTS-551

1. Configure GovOne using Omniauth OpenId Connect.
2. New controller action to handle callbacks from gov one in the case of a successful authentication.
3. Rules to ensure existing applications are merged correctly:
  a) In the case that a teacher record with the same gov one id (sub) exists, we will ensure to sign that teacher in.
  b) In the case that a teacher record with the same email exists but doesn't have a gov one id, then we will sign that teacher in and set the gov one id on the record. This is likely to happen for teachers that have already signed up previously to our application and will now go through GovOne. 
  c) In the case that no teacher record exists to either email or gov one id, we generate a new teacher record and carry onto their new application form.

To ensure we can test this, I have introduced some new buttons for signing up, signing in and signing out via GovOne. All buttons are wrapped around a feature flag. This ensures that once we're happy with this initial version of the integration, we can push this into main without impacting production until we fully finish the GovOne feature lists.


syed87 (approved):
looks good to me

PR #2409, opened 2024-09-20
[AQTS-578] Ability to archive staff in support console

Hassanmir92:
https://dfedigital.atlassian.net/browse/AQTS-578

1. Introducing the concept of "archived" on staff records.
2. Amending all policies to not authorize any access to any part of the support or assessor interfaces and actions to any staff  that is archived.
3. Removing archived staff from the list of staff that is displayed within the support console.


syed87 (approved):
looks good to me, the session we had going over pundit really meant this actually made sense

PR #2422, opened 2024-09-30
[AQTS-585][AQTS-591] Submission and LoPS request journeys (Nigeria, Hong Kong & Alberta, Canada)

Hassanmir92:
1. Update the UI and copy for applicants requiring Letter of Professional Standing (Alberta Canada, Hong Kong and Nigeria) in addition to any one of the requiring preliminary checks (Nigeria).
2. Remove a bug where "initial checks passed" email was sent to countries who didn't require any checks.
3. Update the email journey and copy for LoPS and Preliminary check journey.

https://dfedigital.atlassian.net/browse/AQTS-585
https://dfedigital.atlassian.net/browse/AQTS-591

PR #2440, opened 2024-10-14
Remove initial checks passed mailer template as it has been removed

Hassanmir92:
I identified that this email was removed as part of https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2422/files and the template is no longer required.


syed87 (approved):
lgtm

PR #2444, opened 2024-10-16
[AQTS-625] Validation for referee email address in CMS, FI & draft application when you save and continue

Hassanmir92:
https://dfedigital.atlassian.net/browse/AQTS-625

1. Ensuring work history referee email is validated on all assessor interfaces. This includes work history referee updates by admins and also at the final review stage.
2. Ensure work history contact email is validation while the applicant is filling out their draft application and if they decide to click "Save and come back later", the Work histories section does not move to complete if the email is invalid and instead it moves to "In progress".
3. Ensure work history contact email is validation while the applicant is filling out their response to further information request for referee details and if they decide to click "Save and come back later", the FI does not move to complete if the email is invalid and instead it moved to "Not Started".

### DIFF
--- a/app/forms/assessor_interface/work_history_contact_form.rb
+++ b/app/forms/assessor_interface/work_history_contact_form.rb
@@ -11,6 +11,8 @@ class AssessorInterface::WorkHistoryContactForm
 
   validates :work_history, :user, presence: true
 
+  validates :email, valid_for_notify: true
+
   def save
     return false if invalid?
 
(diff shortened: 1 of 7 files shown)


app/forms/assessor_interface/work_history_contact_form.rb
```diff
@@ -11,6 +11,8 @@ class AssessorInterface::WorkHistoryContactForm
 
   validates :work_history, :user, presence: true
 
+  validates :email, valid_for_notify: true
```
syed87, line 14: I've had a good search for valid_for_notify but couldn't find any documentation, what does this do?

syed87: ok, I think figured it out, it's a custom validation ValidForNotifyValidator. But when it's called you leave of Validator as convention?

syed87 (approved):
looks good to me

PR #2456, opened 2024-10-23
[AQTS-633] If user has no account and attempts to sign in they should go through eligibility check

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-633

Based on the new GOV.UK One Login integration, users have the ability to sign in before having gone through the eligibility checker. This ensures that users that do not have an active application (no teacher record), are forced into the eligibility flow and can only sign into our system if they are eligible.


syed87 (approved):
looks good to me

PR #2457, opened 2024-10-24
2041 configure bigquery in the terraform module new dataset

saliceti:


RMcVelia:
applied in different PR's

PR #2463, opened 2024-10-25
AQTS-632-FI-free-text-validation

syed87:
I've updated the error message and hit text on the page where there is a free text response to and FI request

https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-632

### DIFF
--- a/app/forms/teacher_interface/further_information_request_item_text_form.rb
+++ b/app/forms/teacher_interface/further_information_request_item_text_form.rb
@@ -6,7 +6,10 @@ module TeacherInterface
     attribute :response, :string
 
     validates :further_information_request_item, presence: true
-    validates :response, presence: true
+    validates :response,
+              presence: {
+                message: "Please respond to the assessor's request",
+              }
 
     def update_model
       further_information_request_item.update!(response:)
--- a/app/forms/teacher_interface/further_information_request_item_text_form.rb
+++ b/app/forms/teacher_interface/further_information_request_item_text_form.rb
@@ -6,13 +6,7 @@ module TeacherInterface
     attribute :response, :string
 
     validates :further_information_request_item, presence: true
-    validates :response,
-              presence: {
-                message:
-                  I18n.t(
-                    "teacher_interface.further_information_request_item_text_form.response.blank",
-                  ),
-              }
+    validates :response, presence: true
 
     def update_model
       further_information_request_item.update!(response:)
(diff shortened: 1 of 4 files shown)


app/forms/teacher_interface/further_information_request_item_text_form.rb
```diff
@@ -6,7 +6,10 @@ class FurtherInformationRequestItemTextForm < BaseForm
     attribute :response, :string
 
     validates :further_information_request_item, presence: true
-    validates :response, presence: true
+    validates :response,
+              presence: {
+                message: "Please respond to the assessor's request",
+              }
```
Hassanmir92, line 12: We should use the `I18n` locals instead.

Take a look at `teacher_interface.en.yml` file for other examples.

spec/forms/teacher_interface/further_information_request_item_text_form_spec.rb
```diff
@@ -14,7 +14,6 @@
   let(:response) { nil }
 
   it { is_expected.to validate_presence_of(:further_information_request_item) }
-  it { is_expected.to validate_presence_of(:response) }
```
Hassanmir92, line 17: Why is this removed?

app/forms/teacher_interface/further_information_request_item_text_form.rb
Hassanmir92: You shouldn't have to make any changes here.

spec/forms/teacher_interface/further_information_request_item_text_form_spec.rb
syed87: If I don't remove it I get this error

image

and just searching around one ai chat bot said this

Screenshot 2024-10-25 at 10 15 55

This might not be true because it's ai but I understood this to mean that this line I've removed is automatically checking and expecting the error message to be 'can't be blank' by default so I couldn't see a way of changing it. I didn't check any documentation for the 'shoulda-matchers' gem.

config/locales/teacher_interface.en.yml
```diff
@@ -95,6 +95,9 @@ en:
           school_details_cannot_be_verified: Tell us more about this school
           unrecognised_references: Tell us more about your work history
           work_history_break: Tell us more about your work history
+    further_information_request_item_text_form:
+      response:
+        blank: Please respond to the assessor's request
```
Hassanmir92, line 100: You can remove this now

Hassanmir92 (approved):
LGTM 💪 

Thanks for addressing comments

PR #2478, opened 2024-10-30
[AQTS-631] GOV.UK One Login monitor when an applicant updates the gov uk one login email address

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-631

This work ensures that over time we keep track of the latest email the users use to log in with the GOV.UK One Login account. 

Currently, when an applicant who already has an application with Apply for QTS decides to update their email on GOV.UK One Login, we do not change the email on the application. However, they can still use that account to access their application. With this change, we want to still keep track of this change and if in the future we decide to change our strategy, this can be done much easier.


syed87 (approved):
looks good to me

PR #2514, opened 2024-11-18
[2152] Use testing terraform modules in preproduction

saliceti:
## Description
Fix for https://github.com/DFE-Digital/apply-for-qualified-teacher-status/actions/runs/11895091096/job/33150995864

The service account name fix is required as the preproduction environment name is too long


Hassanmir92 (approved):
Lgtm

PR #2523, opened 2024-11-21
AQTS-643-region-form-testing-AC2

syed87:
https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-643

This pr adds tests for the save method

### DIFF
--- a/spec/forms/support_interface/region_form_spec.rb
+++ b/spec/forms/support_interface/region_form_spec.rb
@@ -226,6 +226,7 @@ RSpec.describe SupportInterface::RegionForm, type: :model do
         subject
         expect(region).to have_attributes(
           application_form_skip_work_history: true,
+          reduced_evidence_accepted: false,
         )
       end
     end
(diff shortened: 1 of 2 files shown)


spec/forms/support_interface/region_form_spec.rb
```diff
@@ -153,4 +153,94 @@
...
+          written_statement_optional: true,
+        )
+      end
+
+      it "processes emails correctly" do
+        subject
+        expect(region.teaching_authority_emails).to eq(
+          %w[email1@example.com email2@example.com],
+        )
+      end
```
Hassanmir92, line 222: You already have these in the `updates region attributes` test, don't think you need to repeat them.

syed87: ah yes, I wrote these first and then found I had to change the previous test to include the email rather than "" to get the test to pass.

Hassanmir92 (approved):

PR #2525, opened 2024-11-21
AQTS 593 visa sponsorship link

syed87:
This pr changes the link in the award communications, and service page so that candidates are directed to jobs which provide sponsorship. 

https://dfedigital.atlassian.net/browse/AQTS-593


Hassanmir92 (approved):
💪

PR #2553, opened 2024-12-09
[AQTS-670] Build client to make post and get requests for TRN allocation

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-670

As per documentation for the new [TRS TRN allocation API](https://github.com/DFE-Digital/teaching-record-system/blob/main/docs/api-designs/trn.md), this PR implements the 2 clients which will be used in the future alongside [setting QTS against](https://github.com/DFE-Digital/teaching-record-system/blob/main/docs/api-designs/qtls.md) a record.

This new client is not inserted into our flow yet as it can only work alongside QTS assignment. 

Future work:
1. Expect `potentialDuplicate` to be returned as part of the response for the V3 create TRN record request.
2. Implement setting QTS for TRN record once the API endpoint is ready on [TRS](https://github.com/DFE-Digital/teaching-record-system).
3. Replace the existing process with these 2 endpoints.
4. Update the API version header value for when all endpoints go into production on [TRS](https://github.com/DFE-Digital/teaching-record-system)

### DIFF
--- a/app/lib/trs/client/v3/create_trn_request.rb
+++ b/app/lib/trs/client/v3/create_trn_request.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class TRS::Client::V3::CreateTRNRequest
+  include ServicePattern
+  include TRS::Client::Connection
+
+  def initialize(request_id:, application_form:)
+    @request_id = request_id
+    @application_form = application_form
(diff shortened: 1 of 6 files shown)


syed87 (approved):
lgtm

PR #2555, opened 2024-12-10
AQTS 729 ineligibility content changes

syed87:
https://dfedigital.atlassian.net/browse/AQTS-729

This PR implements content modifications on the ineligible page of the eligibility checker.

### DIFF
--- a/app/views/eligibility_interface/result/ineligible.html.erb
+++ b/app/views/eligibility_interface/result/ineligible.html.erb
@@ -45,11 +45,13 @@
   </ul>
 
   <% if @eligibility_check.ineligible_reasons.include?(:work_experience) %>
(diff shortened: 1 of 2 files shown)


app/views/eligibility_interface/result/ineligible.html.erb
```diff
@@ -2,6 +2,7 @@
 
 <% content_for :page_title, "You’re not eligible" %>
 <% content_for :back_link_url, ineligible_country ? eligibility_interface_countries_path : eligibility_interface_misconduct_path %>
+<% current_date = Date.today %>
```
Hassanmir92, line 5: Could you use `Date.current`. Using that class method on date always considers the systems time zone.

PR #2547, opened 2024-12-05
[AQTS-688] New trs_trn_requests table to replace dqt_trn_requests

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-688

As part our move away from DQT, we hold a table called dqt_trn_requests. We want to rename this table to be trs_trn_requests.

In this PR, we:
1. Introduce the new trs_trn_requests table.
2. Ensure all new requests to TRS populate this new table.
3. We have a rake task to move all existing dqt_trn_requests to trs_trn_requests

Release order:
1. Merge this PR.
2. Run `rails dqt_to_trs_migration:replace_dqt_trn_requests_table_with_trs_trn_requests` rake task.
3. PR to add new table to dataform
4. Remove dqt_trn_requests table and any use of it.
5. PR to remove dqt_trn_requests to  to dataform.

PR #2556, opened 2024-12-10
[AQTS-688] Drop DQTTRNRequest model and table as it is no longer in use

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-688

As per merging in https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2547 and waiting for all existing `UpdateTRSTRNRequestJob` jobs in the scheduled queue to run, we can now remove the `dqt_trn_requests` table as it is no longer in use and fully replaced by `trs_trn_requests`. 

A rake task has already been run to migrate all existing records of `dqt_trn_requests` to `trs_trn_requests`.


syed87 (approved):
lgtm

PR #2562, opened 2024-12-16
[DO NOT MERGE] Removing verbose: false on sidekiq config

Hassanmir92:
Currently we have no logs when it comes to jobs triggered on Sidekiq. This has made it harder to confirm whether certain analytics jobs had been triggered or not.


Hassanmir92:
Closing PR as removing this will have an impact on logit allowance. This is not urgent and we can ignore.

PR #2566, opened 2024-12-18
[AQTS-724] Bug fix for country dropdowns not removing option when deleted

Hassanmir92:


Hassanmir92:
Closing as this is no longer a priority. 

Keeping the PR as reference incase it becomes a priority to update.

PR #2571, opened 2024-12-23
[AQTS-671] TRS QTS PUT endpoint

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-671

This work implements the client for the QTS award endpoint for [TRS](https://github.com/DFE-Digital/teaching-record-system)

This specific endpoint will be dependent on a TRN number already allocated to a teacher record before a request can be sent. 

The flow that we will implement will be initially making a request using the [TRN allocation endpoint](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/app/lib/trs/client/v3/create_trn_request.rb) and then using the new QTS professional statuses award endpoint implemented in the PR for the TRN record.

### DIFF
--- a/app/lib/trs/client/v3/update_qts_request.rb
+++ b/app/lib/trs/client/v3/update_qts_request.rb
@@ -18,7 +18,7 @@ class TRS::Client::V3::UpdateQTSRequest
         req.body = body
       end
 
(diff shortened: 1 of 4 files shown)


syed87 (approved):
lgtm

PR #2574, opened 2024-12-25
AQTS 643 region form testing AC3

syed87:
https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-643

This pr completes the final AC of the ticket for region form testing

### DIFF
--- a/spec/forms/support_interface/region_form_spec.rb
+++ b/spec/forms/support_interface/region_form_spec.rb
@@ -258,4 +258,73 @@ RSpec.describe SupportInterface::RegionForm, type: :model do
       end
     end
   end
+
+  describe ".for_existing_region" do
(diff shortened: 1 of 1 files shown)


spec/forms/support_interface/region_form_spec.rb
```diff
@@ -258,4 +258,73 @@
...
     end
   end
+
+  describe ".for_existing_region" do
+    subject(:form) { described_class.for_existing_region(region) }
+
+    let(:region) do
+      create(
+        :region,
+        application_form_skip_work_history: false,
```
Hassanmir92, line 268: How about we have this be defined separately inside a `let`?

This way, you can just then redefine it inside your contexts

Hassanmir92 (approved):

PR #2613, opened 2025-01-23
AQTS 782 remove archived staff

syed87:
https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-782

This pr will enable us to remove archived staff from the 'Assessor name' filter


app/view_objects/assessor_interface/application_forms_index_view_object.rb
```diff
@@ -25,6 +25,7 @@ def assessor_filter_options
     ApplicationForm
       .submitted
       .joins(:assessor)
+      .where(staff: { archived: false })
```
Hassanmir92, line 28: I don't think this is the right place.

With this solution, you would still be showing the archived users on the filter list but not return anything. Their applications can never be found.

What you need is to ensure that the names of staff that are archived do not appear on the assessor list you can filter for but their applications can still be searched for.

Hassanmir92 (approved):

PR #2638, opened 2025-02-10
[AQTS-821] Introduce a view only page for applications that have been submitted

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-821

This PR introduces a read only view of the applicants submission after they have submitted their application. This allows them to review the information they have submitted as it is not currently available. This page is currently only accessible via direct URL link. In the future, we may decide to allow easies access to this page.

### DIFF
--- a/app/controllers/teacher_interface/application_forms_controller.rb
+++ b/app/controllers/teacher_interface/application_forms_controller.rb
@@ -51,6 +51,9 @@ module TeacherInterface
       @view_object = ApplicationFormViewObject.new(application_form:)
     end
 
+    def answers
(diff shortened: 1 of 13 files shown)


app/controllers/teacher_interface/application_forms_controller.rb
```diff
@@ -51,6 +51,9 @@ def show
       @view_object = ApplicationFormViewObject.new(application_form:)
     end
 
+    def answers
```
syed87, line 54: How come this method is empty?

PR #2652, opened 2025-02-17
[AQTS-817] Remove the need for taking the applicant to a specific step on english language proficiency 

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-817

This fixes the issue where the application task for "English language proficiency" has logic to take the applicant back to where they last left their step, however this introduces some issues where the applicant is stuck without being able to go back. 

Our "Back" button is not the appropriate fix for this and therefore we are simplifying this step to ensure this applicant is always taken to the first step of the "English language proficiency" if the step is left incomplete and they come back later. This allows the applicant to review the already submitted answers and complete this stage. This is also aligned with how the other steps work.

### DIFF
--- a/app/controllers/teacher_interface/english_language_controller.rb
+++ b/app/controllers/teacher_interface/english_language_controller.rb
@@ -24,47 +24,10 @@ module TeacherInterface
                       application_form
                       english_language
                     ]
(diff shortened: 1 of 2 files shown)


syed87 (approved):
lgtm

PR #2657, opened 2025-02-19
[AQTS-799] Ensure logs exclude all parameters from forms on top of PI data

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-799

Adding new parameters to filter logs by all contents sent via form parameters to ensure no personal data goes through our logs if in the future new columns are introduced that may contain PI data.


syed87 (approved):
lgtm

PR #2660, opened 2025-02-20
[AQTS-728] Verification spoke showing on applications which were declined during assessment

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-728

This PR fixes the issue for when declined applications that were declined during initial assessment still show the verification decision task in the assessor UI.


syed87 (approved):
lgtm

PR #2662, opened 2025-02-25
Ensure test data for review app to have passport route by default

Hassanmir92:
This PR makes all the fake data generated have passport as a way to verify identity for review apps.


syed87 (approved):
lgtm

PR #2667, opened 2025-02-26
[AQTS-796] Remove teaching confirmation being set to true on edit by default

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-796

This PR ensures that the applicant must check the confirmation box whenever they go to edit the teaching qualification. Previously, if the user saved their progress and came back later, the checkbox would be checked by default which is the incorrect behaviour.


syed87 (approved):
Lgtm

PR #2669, opened 2025-02-28
[AQTS-853] Work history FI updates

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-853

This PR:
1. Updates the content for unrecognised references failure reason on the work history task.
2. "Unrecognised references" FI to now capture new referee details instead of "school details cannot be verified"
As part of this, we ensure that existing assessment that have already sent these reasons, still display the work histories selected under "school details cannot be verified" for future reference + the existing FI's sent out won't be impacted by this change and the applicant will still return the originally expected response type.

Screenshot 2025-03-04 at 10 52 37


syed87 (approved):
lgtm

PR #2677, opened 2025-03-05
AQTS 837 manage staff users permissions

syed87:
https://dfedigital.atlassian.net/browse/AQTS-837

This pr will add a column to the Staff table which will allow us to determine which user can give permissions to staff

There are also UI changes to the personas page to show the new permission


Hassanmir92 (approved):
Great work 💪 ✅

PR #2687, opened 2025-03-11
[AQTS-867] Ensure LoPS is not generated due to session value being passed

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-867

The verification recommendation form submission relies on session values stored during the process of filling out the recommendation by the assessor for each application. However, at times there may be LoPS requested for a different assessment being passed over for application that do not require them and as a result, it ends up with the application with incorrect verification steps. 

This PR for now ensures that LoPS and references are only generated if the application does not skip them. 

In the future we may want to move away from relying on storing the progress within the session as it could bring up more issues.


syed87 (approved):
lgtm

PR #2692, opened 2025-03-14
[AQTS-862] Remove review previous referee details on FI after it has been accepted

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-862

Minor change to remove showing previous referee details once the FI has already been reviewed. Since once it's accepted, the values will be the same because of the update and this may cause confusion.


syed87 (approved):
lgtm

PR #2693, opened 2025-03-17
Update ruby back to 3.3.0

Hassanmir92:
Updating ruby back to to 3.3.0 as there seem to be some issues with the upgrade.  https://github.com/ruby/ruby/releases/tag/v3_3_0

Our attempt to 3.4.2 seems to have issues with postgres14-dev not being available on the docker image, so we can look into that separately. To unblock us for now, we will just go back to the version that we had.


syed87 (approved):
lgtm

PR #2702, opened 2025-03-19
[AQTS-896] Do not show an error message when confirming an award assessor

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-896

Previously, when moving to the final section to award QTS to application, an error would should by default for the confirmation question. We are now removing this as part of this PR. 

The error was happening due to the fact that the previous page, the continue button was part of the next pages form and as a result of the value of the confirmation not being passed through, the error message was being rendered.


syed87 (approved):
lgtm

PR #2705, opened 2025-03-21
[DO NOT MERGE][AQTS-495] Move from azure-storage-blob to azure-blob

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-495

The [Azure Storage library](https://github.com/Azure/azure-storage-ruby) that we were using for file uploads has been deprecated and will no longer be supported after the 13th September.

That doesn’t mean file uploads will stop working after that date, but there is the potential for bugs or security issues to be discovered in the library that will then be up to us to fix.

[azure-blob](https://github.com/testdouble/azure-blob) is a library that replaces the abandoned azure-storage-blob. This library is being used by a few other services in DfE as well and is maintained.

### DIFF
--- a/app/services/fetch_malware_scan_result.rb
+++ b/app/services/fetch_malware_scan_result.rb
@@ -17,7 +17,7 @@ class FetchMalwareScanResult
 
     begin
       response = fetch_scan_result
-    rescue Azure::Core::Http::HTTPError
+    rescue AzureBlob::Http::Error
       upload.malware_scan_error!
       return
     end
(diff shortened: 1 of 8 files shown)


app/services/resend_stored_blob_data.rb
```diff
@@ -1,13 +1,10 @@
 # frozen_string_literal: true
 
+require "azure-blob"
+
 class ResendStoredBlobData
   include ServicePattern
 
-  # Only later versions of the Azure Storage REST API support tags operations.
```
Hassanmir92, line 6: The default API version on the new library is after "2022-11-02", so this may no longer be needed

app/services/fetch_malware_scan_result.rb
```diff
@@ -34,36 +31,25 @@ def call
...
+        access_key: ENV["AZURE_STORAGE_ACCESS_KEY"],
+        container: ENV["AZURE_STORAGE_CONTAINER"],
       )
   end
 
   def fetch_scan_result
-    blob_service.call(:get, get_tags_for_blob_url)
-  end
-
-  def get_tags_for_blob_url
```
Hassanmir92, line 47: With the new library the `get_blob` already generates the uri itself with the container name prepended so we no longer need these. 

Reference: https://github.com/testdouble/azure-blob/blob/main/lib/azure_blob/client.rb#L70

app/services/fetch_malware_scan_result.rb
```diff
@@ -34,36 +31,25 @@ def call
...
-    @get_tags_for_blob_url ||=
-      blob_service.generate_uri(
-        File.join("uploads", upload.attachment.key),
-        { comp: "tags" },
-      )
+    blob_service.get_blob("#{upload.attachment.key}?comp=tags")
   end
 
   def update_scan_result(response)
-    if response.success?
```
Hassanmir92, line 56: We no longer need this check, since the new library automatically raises an error instead. 

Reference: https://github.com/testdouble/azure-blob/blob/main/lib/azure_blob/http.rb#L49

PR #2709, opened 2025-03-26
[AQTS-876] FI response received UI updates + Accept/Reject per FI item

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-876

1. Updating the visuals on reviewing further information requests that have been received
2. New functionality to accept or decline further information requests based on per item rather than the FI as a whole. This enables us to move forward with multiple FI's in the future. 
a) If all items are accepted, we accept the FI and move into verification.
b) If one or more items are declined, the application moves to another step to leave an internal decline note before proceeding to decline the application.

Screenshot 2025-03-28 at 15 35 07
Screenshot 2025-03-28 at 15 34 46
Screenshot 2025-03-28 at 15 35 40


syed87 (approved):
lgtm

PR #2723, opened 2025-04-10
[AQTS-672] Plug in new TRS V3 API endpoint requests into application award flow

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-672

This PR updates the `UpdateTRSTRNRequestJob` to start using the V3 TRS endpoints instead of single V2 endpoint.

The main change except switching the client is that we now have 2 endpoints via V3 whereas old V2 endpoint was a single endpoint which handled everything. 

In the new endpoints one is for the TRN allocation (TRS::Client::V3::CreateTRNRequest) which will return a `trn` number, whether the record is a potential duplicate and the URL to access qualifications. Once the TRN allocation is successful and no potential duplicate has been flagged, the job will carry on to make a 2nd call to TRS to allocate QTS award (TRS::Client::V3::UpdateQTSRequest) and mark the record within our system as awarded (AwardQTS). 

As part of this change, any timeline changes on an application record will now reference the change made to be done by "TRS" instead of "DQT". This can be reviewed when looking at how we make a call `AwardQTS` and pass in `user: "TRS"`.


syed87 (approved):
lgtm

PR #2726, opened 2025-04-16
Multiple further information requests feature

Hassanmir92:
Tickets:
https://dfedigital.atlassian.net/browse/AQTS-875
https://dfedigital.atlassian.net/browse/AQTS-877
https://dfedigital.atlassian.net/browse/AQTS-879
https://dfedigital.atlassian.net/browse/AQTS-878
https://dfedigital.atlassian.net/browse/AQTS-912

This PR implements the new multiple further information (FI) requests  feature. Currently, FI requests can only be sent once and the response from applicants can be either accepted or declined (which leads to applications being declined). However, assessors sometimes need clarification based on the applicants response, and currently that happens outside the system, but we want assessors to be able to use the system to request any follow up questions to previous FI requests.

In this PR, I have introduced a new option within reviewing a further information for "No, request further information" alongside the accept and decline.
Screenshot 2025-04-24 at 15 23 25 

If any of the FI items have "No, request further information" and no declines have been selected on the review, then the flow moves into entering the `review_decision_note` for each of the FI items. In this specific page, all FI items selected for a follow up have been grouped by their original assessment section which is implemented by `grouped_follow_up_items_by_assessment_section` within the `AssessorInterface::FurtherInformationRequestViewObject` view object. 

Additionally, the `AssessorInterface::FurtherInformationRequestFollowUpForm` object uses a similar pattern to `AssessorInterface::AssessmentSectionForm` to dynamically build attributes and validations for the form based on the selection made from the previous section. 

For this pattern, we have 3 class methods:
1. `for_further_information_request(further_information_request)` - This is used to define the class from a specific further_information_request so that the relevant attributes and validations depending on which items have been selected for further information can be dynamically part of the characteristics of the class.
2. `initial_attributes(further_information_request)` - The purpose of this is to dynamically identify the attribute values and set them on the form object when it has already been filled out before. This helps display existing values on the page.
3. `permittable_parameters(further_information_request)` - The purpose of this is to ensure when wanted to define the permitted parameters used within the controller dynamically based on the FI item selections.

Screenshot 2025-04-25 at 00 00 48

The flow then moves into a page utilised by `grouped_follow_up_items_by_assessment_section` method to display a review section of the new FI request being sent.

Screenshot 2025-04-25 at 00 03 48

For the benefit of once an assessor has submitted I have renamed original `RequestFurtherInformation` class which is a service used to generate an FI request from initial assessment to `FurtherInformationRequests::CreateFromAssessmentSections` this then has allowed me to also create a new service class `FurtherInformationRequests::CreateFromFurtherInformationReview` which generates an FI request from an existing FI requests review. 

**Important:** An edge-case is also that for FI request items that change referee details, if any are accepted, the work history reference/contact details will be automatically updated, even if the request goes for any follow ups. This change can be seen within `after_reviewed` method inside the `FurtherInformationRequest` model which is triggered automatically after every `Requestable` object is reviewed. 

This process can then be repeated up to 2 times (3 FI requests total) before the assessor no longer is able to see the "No, request further information" option on the final/third request.

Screenshot 2025-04-25 at 00 16 53

Additionally, this PR ensures that 2nd and 3rd requests expire within 3 weeks (instead of 6 for the initial request) and that we only send 2 reminders at 1 week before expiry and 2 days (instead of 2 weeks, 1 week and 2 days before). This change can be seen within `expires_after` (part of the `Requestable` concern) and `should_send_reminder_email?` (part of the `Remindable` concern) methods within the `FurtherInformationRequest`.

### DIFF
--- a/app/controllers/assessor_interface/further_information_requests_controller.rb
+++ b/app/controllers/assessor_interface/further_information_requests_controller.rb
@@ -21,10 +21,13 @@ module AssessorInterface
     end
 
     def create
-      RequestFurtherInformation.call(assessment:, user: current_staff)
+      FurtherInformationRequests::CreateFromAssessmentSections.call(
+        assessment:,
+        user: current_staff,
+      )
 
       redirect_to [:status, :assessor_interface, application_form]
-    rescue RequestFurtherInformation::AlreadyExists
+    rescue FurtherInformationRequests::CreateFromAssessmentSections::AlreadyExists
       flash[:warning] = "Further information has already been requested."
       render :new, status: :unprocessable_entity
     end
--- a/app/controllers/assessor_interface/further_information_requests_controller.rb
+++ b/app/controllers/assessor_interface/further_information_requests_controller.rb
@@ -11,9 +11,10 @@ module AssessorInterface
     end
 
     before_action :ensure_can_decline, only: %i[edit_decline update_decline]
+    before_action :ensure_can_follow_up,
+                  only: %i[edit_follow_up update_follow_up]
     before_action :load_application_form_and_assessment, only: %i[new edit]
-    before_action :load_view_object,
-                  only: %i[edit update edit_decline update_decline]
+    before_action :load_view_object, except: %i[new create]
 
     def new
       @view_object =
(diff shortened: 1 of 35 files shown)


syed87 (approved):
lgtm

PR #2736, opened 2025-04-28
[AQTS-950] Adding new GOV.UK Notify API key and template IDs for our new account

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-950

This work introduces references to our new GOV.UK Notify account. I have decided to introduce a new API key name so that we have no downtime between updating the environment variable and the code deploying. Once deployed, we can remove the old key.


syed87 (approved):
lgtm

PR #2746, opened 2025-05-07
[AQTS-955] Adding new failure reason for ESOL expired

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-955

1. Adding new ESOL expired decline reason to applications that have ESOL as an option to prove English language proficiency (ELP).
2. Remove the existing ESOL expired during assessment FI reason - This will be applied to new and existing applications.
3. Remove the existing SELT expired during assessment FI reason - This will be applies to new applications while existing applications will have a "[DO NOT USE]" at the beginning of this FI reasons for the assessor.

### Post release action

Given that there are a few applications from the relevant country and with ESOL selected, we need to add this new reason to those existing submissions after we release the feature.

This can be done by running: `SyncAssessmentChecksAndFailureReasonsJob.perform_later(assessment)` for each of those application assessments.

It's important to note that we're only doing this to applications that have not yet started their assessment (which happens to be the case for all the applications with ESOL since it's a newly released feature)

### DIFF
--- a/app/lib/assessment_factories/english_language_proficiency_section.rb
+++ b/app/lib/assessment_factories/english_language_proficiency_section.rb
@@ -31,6 +31,7 @@ module AssessmentFactories
           [
             FailureReasons::EL_REQUIRE_ALTERNATIVE,
             FailureReasons::EL_ESOL_DOCUMENT_ILLEGIBLE,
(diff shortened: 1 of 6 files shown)


syed87 (approved):
lgtm

PR #2757, opened 2025-05-21
AQTS 978 unlinking onelogin accounts

syed87:
This pr makes a change to gov onelogin so that similar to `gov_one_email` the `gov_one_id` is updated every time someone logs in.

https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-978

### DIFF
--- a/spec/services/find_or_create_teacher_from_gov_one_spec.rb
+++ b/spec/services/find_or_create_teacher_from_gov_one_spec.rb
@@ -133,4 +133,19 @@ RSpec.describe FindOrCreateTeacherFromGovOne do
       expect(existing_teacher_matching_id.reload.gov_one_email).to eq(email)
     end
   end
+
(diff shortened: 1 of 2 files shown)


spec/services/find_or_create_teacher_from_gov_one_spec.rb
```diff
@@ -133,4 +133,25 @@
...
+
+    it "updates the gov_one_id on the existing teacher record" do
+      call
+      expect(existing_teacher.reload.gov_one_id).to eq(gov_one_id)
+    end
+
+    it "updates the gov_one_email on the existing teacher record" do
+      call
+      expect(existing_teacher.reload.gov_one_email).to eq(email)
+    end
```
Hassanmir92, line 155: Not sure if you need this test since the email is the same anyway right? That's how the record was found.

Hassanmir92 (approved):
Great work! 💪

PR #2759, opened 2025-05-22
[AQTS-965] Re-order the eligibility criteria questions

Hassanmir92:
Tickets:
https://dfedigital.atlassian.net/browse/AQTS-965
https://dfedigital.atlassian.net/browse/AQTS-992

1. Re-order the eligibility criteria so that the ages the applicant can teach and question related to subject restrictions come towards the end
2. Change to ensure that when users that choose "Scotland" as their country they trained in but answer "No" when they need to confirm they have trained in Scotland then they are directly taken to the result page without having to answer all the other eligibility questions


syed87 (approved):
lgtm

PR #2771, opened 2025-06-02
AQTS 996 filter by dob

syed87:
This pr will add a new filter to the CMS so that agents can filter applicants by their dob

https://dfedigital.atlassian.net/browse/AQTS-996


Hassanmir92 (approved):
Really nice work mate 👍 LGTM

PR #2774, opened 2025-06-05
Revert "AQTS 787 error message location"

Hassanmir92:
Reverts DFE-Digital/apply-for-qualified-teacher-status#2761

This causes issues where on "Check professional standing" where "show_form?" in the view object returns false and as a result there is no content for "assessment_section_top_of_form" being displayed.


Hassanmir92:
Closing in favour of actual fix: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2775

PR #2775, opened 2025-06-05
Fix bug for assessment section that does not yield the top of form section

Hassanmir92:
The issue came up as as we got a report from one of the assessors that the "Check professional standing" request is showing an empty page.

THis was due to the fact that both `@view_object.show_form?` and `@view_object.show_english_language_exemption_content?` were returning false and the content for `assessment_section_top_of_form` was not being yielded. 

This change ensures that we always show the top of form info even if no form or ELP content can be displayed.


syed87 (approved):
lgtm

PR #2783, opened 2025-06-12
AQTS-1037-error-message-bug-fix

syed87:
https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-1037

This pr is to fix the validation for the subject 1 field on the Verify age range and subjects page in the assessor view. 

image

### DIFF
--- a/spec/support/shared_examples/age_range_subjects_form.rb
+++ b/spec/support/shared_examples/age_range_subjects_form.rb
@@ -79,11 +79,7 @@ RSpec.shared_examples_for "an age range subjects form" do
 
       it "adds an error to subject_1" do
         form.valid?
-        expect(form.errors[:subject_1]).to include(
-          I18n.t(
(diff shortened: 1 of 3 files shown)


spec/support/shared_examples/age_range_subjects_form.rb
```diff
@@ -70,7 +70,7 @@
     it { is_expected.not_to validate_presence_of(:age_range_note) }
 
     it { is_expected.to validate_presence_of(:subject_1) }
-    it { is_expected.to validate_presence_of(:subject_1_raw) }
+    it { is_expected.not_to validate_presence_of(:subject_1_raw) }
```
Hassanmir92, line 73: You may need to move away from these shoulda matchers for subject_1 and subject_1_raw.

This line is not valid since `subject_1_raw` does have a presence validation. 

I recommend doing some custom tests for ensuring that subject_1 and subject_1_raw are both validated.

PR #2794, opened 2025-06-27
Linking our review app with TRS pre-prod

Hassanmir92:
We're updating the TRS API on our review apps to connect to the TRS preproduction environment as the dev environment doesn't seem stable for testing of feature changes that are coming up between the two systems.


syed87 (approved):
lgtm

PR #2795, opened 2025-06-27
TRS API naming updates

syed87:
https://dfedigital.atlassian.net/browse/AQTS-1056

This pr updates naming in preparation for TRS API changes

### DIFF
--- a/terraform/application/config/review/variables.yml
+++ b/terraform/application/config/review/variables.yml
@@ -1,3 +1,3 @@
 ---
-TRS_API_URL: https://dev.teacher-qualifications-api.education.gov.uk
+TRS_API_URL: https://preprod.teacher-qualifications-api.education.gov.uk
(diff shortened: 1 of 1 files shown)


app/lib/trs/v3/qts_request_params.rb
```diff
@@ -22,9 +22,9 @@ def initialize(application_form:, awarded_at:)
 
       def call
         {
-          routeTypeId: route_type_id,
+          routeToProfessionalStatusTypeId: route_type_id,
           status: AWARDED_QTS_STATUS,
```
syed87, line 26: In the ticket you mentioned 'Approved' and 'Awarded', I can't see 'Awarded' in the code so if the value for AWARDED_QTS_STATUS is set anywhere else then I could change this code to `status: "Holds"` to avoid any other status being set

PR #2817, opened 2025-07-17
Data and performance improvements to prioritisation features

Hassanmir92:
1. Introducing a new index for the `prioritised` column within the `assessments` table. This is because this specific column will be used in the CMS as a filter.
2. Introducing a new column for `working_days_between_submitted_and_prioritisation_decision` which is to support analysis of SLA's with regards to prioritised applications.


syed87 (approved):
lgtm

PR #2830, opened 2025-08-05
[AQTS-1125] Ensure uploaded translated documents are displayed for FI responses

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1125

This fixes the bug where FI response where an upload has translation documents were not being surfaced to the assessor.

### DIFF
--- a/app/views/assessor_interface/further_information_requests/edit.html.erb
+++ b/app/views/assessor_interface/further_information_requests/edit.html.erb
@@ -32,7 +32,8 @@
         <% end %>
 
         <% if item[:applicant_upload_response] %>
-          <%= document_link_to(item.fetch(:applicant_upload_response)) %>
+          <% applicant_upload_response = item.fetch(:applicant_upload_response) %>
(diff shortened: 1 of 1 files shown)


app/views/assessor_interface/further_information_requests/edit.html.erb
```diff
@@ -32,7 +32,7 @@
         <% end %>
 
         <% if item[:applicant_upload_response] %>
-          <%= document_link_to(item.fetch(:applicant_upload_response)) %>
+          <%= document_link_to(item.fetch(:applicant_upload_response), translated: item.fetch(:applicant_upload_response).translated_uploads.any?) %>
```
syed87, line 35: Would this be a tad easier to read?

```
<% if item[:applicant_upload_response] %>
    <% applicant_upload_response = item.fetch(:applicant_upload_response) %>
    <%= document_link_to(applicant_upload_response, translated: applicant_upload_response.translated_uploads.any?) %>
<% end %>
```

PR #2831, opened 2025-08-06
Smoke test update for prioritisation

Hassanmir92:
Once we go live with the prioritisation feature using the feature flag, we need to have the smoke test updated to reflect the new updates to the eligibility checker flow.

Additionally, we will ensure all future review apps will have work experience in England feature enabled by default.


syed87 (approved):
lgtm

PR #2904, opened 2025-09-29
[AQTS-1206] Eligiblity email domains CMS Add, view and update record

Hassanmir92:
Tickets:
https://dfedigital.atlassian.net/browse/AQTS-1206
https://dfedigital.atlassian.net/browse/AQTS-1208

This work implements the CMS journeys to add eligibility domain records with CMS used for assessment of work histories using our existing form objects pattern for all changes. 

Part of this PR was to ensure that these records hold an audit trail against the creation, archiving, reactivating and note taking using the `timeline_events` table we already have. Historically `timeline_events` were only applicable for `application_forms` and had a `null: false` constraint which meant that every single timeline_event record had an application_form_id value. This PR removed this constraint so that timeline_events can be used to keep an audit trail of more than just `application_forms` and now can be used for `eligibility_domains` as well.

As well as `timeline_events`, part of the feature list was for assessors to be able to leave notes against `eligibility_domains`. Similar to `timeline_events`, `notes` table was historically only used for `application_forms` and had a similar `null: false` constraint for `application_form_id`. This has now also been removed to ensure alongside `timeline_events`, `notes` can also be used for more than just `application_forms` and can now be created for `eligibility_domains` as well. This also meant that for creating a note feature, we could re-use the same `CreateNoteForm` object with all necessary updates required rather than having to create a brand new form object.


syed87 (approved):
lgtm

PR #2905, opened 2025-09-29
Upgrade rack version to 2.2.18

Hassanmir92:
Due to the [this security risk](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/security/dependabot/69) this upgrade is necessary.


syed87 (approved):
lgtm

PR #2912, opened 2025-10-02
AQTS 1218 reference questions amendments

syed87:
https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-1218

This pr updates the wording for two questions and also removes two questions for future reference requests.

### DIFF
--- a/app/models/reference_request.rb
+++ b/app/models/reference_request.rb
@@ -98,7 +98,7 @@ class ReferenceRequest < ApplicationRecord
     unless excludes_suitability_and_concerns_question?
       responses << satisfied_response
     end
-    
+
     responses.none?(&:nil?)
   end
 


app/controllers/teacher_interface/reference_requests_controller.rb
```diff
@@ -213,12 +219,21 @@ def update_misconduct
...
+          if_failure_then_render: :edit_misconduct,
+        )
+      else
+        handle_application_form_section(
+          form: @form,
+          if_success_then_redirect:
+            satisfied_teacher_interface_reference_request_path,
+          if_failure_then_render: :edit_misconduct,
+        )
+      end
```
Hassanmir92: I think you can refactor this a bit. There's a lot of duplication here and the only thing that really needs a conditional statement is the value of `if_failure_then_render`. Could you think of a different approach here?

app/services/verify_assessment.rb
```diff
@@ -81,7 +81,11 @@ def create_reference_requests
 
     work_histories.map do |work_history|
       ReferenceRequest
-        .create!(assessment:, work_history:)
+        .create!(
+          assessment:,
+          work_history:,
+          excludes_suitability_and_concerns_question: true,
+        )
```
Hassanmir92: Would this need a spec update?

spec/system/teacher_interface/reference_spec.rb
```diff
@@ -3,54 +3,99 @@
 require "rails_helper"
 
 RSpec.describe "Teacher reference", type: :system do
-  before { given_there_is_a_reference_request }
+  context "when filling in a legacy reference request" do
```
Hassanmir92: I recommend not using various ways to refer to the references that do and don't skip concern questions.

I would do:

when reference request excludes suitability and concerns questions
when reference request includes suitability and concern questions

spec/system/teacher_interface/reference_spec.rb
```diff
@@ -180,5 +268,23 @@ def reference_request
       )
   end
 
-  delegate :slug, to: :reference_request
+  def new_reference_request
```
Hassanmir92: I recommend keeping a single reference_request and in your context, just set the `excludes_suitability_and_concerns_question` in a `before do`

spec/models/reference_request_spec.rb
```diff
@@ -119,22 +120,80 @@
...
+        let(:reference_request) do
+          build(
+            :reference_request,
+            :with_responses,
+            excludes_suitability_and_concerns_question: false,
+          )
+        end
+
+        it { is_expected.to be true }
+      end
```
Hassanmir92: I don't feel that your test is actually testing the specific response differences. 

Could you give this spec another try?

What we want to ensure is that the suitability concern question can be nil and the result is still true

PR #2913, opened 2025-10-07
[AQTS-1133] All uploads to be visible to assessors during FI review and not just translations

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1133

This PR ensures that all translated+ original files are visible to assessors when reviewing FI responses.


syed87 (approved):
lgtm

PR #3162, opened 2026-04-10
[AQTS-1435] Draft user stuck with other work experience in england

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1435

We had a report come in that one of our users was stuck filling out their "Other work experience in England". This specific spoke is a multi-step form where the first questions asks them whether they have "Other work experience in England". If they select "Yes", they're progressed to creating the details of this work experience, starting with the school details. 
Screenshot 2026-04-10 at 14 53 57

However, this specific user decided they did not meet the requirements to fill this out, so they clicked "Save and come back later" which persisted the "work history" record and took them back to their main application page. At this stage, the "Add other work experience in England" was then "In progress". Whenever they went back into it to update their answer from "Yes" to "No", then they would end up back on updating the school details without any way for them to go and change their answer.
Screenshot 2026-04-10 at 14 53 39
Screenshot 2026-04-10 at 14 54 11

The workaround would have been for them to just complete the form with dummy data and delete the record to get the question pop-up again, however that is not an acceptable user experience.

In this PR, we ensure that whenever the "Add other work experience in England" is "In Progress", we take the user back to the original question of "Do you have other educational work experience in England that meets our criteria?" before continuing to update the incomplete record. Additionally, we ensure that if they select "No", any existing "Other work experience in England" roles are destroyed so we don't end up with weird abandoned work history records.

### DIFF
--- a/app/controllers/teacher_interface/other_england_work_histories_controller.rb
+++ b/app/controllers/teacher_interface/other_england_work_histories_controller.rb
@@ -39,17 +39,17 @@ module TeacherInterface
                       application_form
                       other_england_work_histories
                     ]
-      elsif (
-            work_history =
-              application_form
-                .work_histories
-                .other_england_educational_role
-                .order_by_user
-                .find(&:incomplete?)
-          )
-        redirect_to school_teacher_interface_application_form_other_england_work_history_path(
-                      work_history,
-                    )
+      elsif application_form
+            .work_histories
+            .other_england_educational_role
+            .order_by_user
+            .find(&:incomplete?)
+        redirect_to %i[
+                      meets_criteria
+                      teacher_interface
+                      application_form
+                      other_england_work_histories
+                    ]
       elsif (
             work_history =
               application_form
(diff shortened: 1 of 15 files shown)


app/controllers/teacher_interface/other_england_work_histories_controller.rb
```diff
@@ -39,17 +39,17 @@ def index
...
-              application_form
-                .work_histories
-                .other_england_educational_role
-                .order_by_user
-                .find(&:incomplete?)
-          )
-        redirect_to school_teacher_interface_application_form_other_england_work_history_path(
-                      work_history,
-                    )
+      elsif application_form
```
syed87: `application_form.work_histories.other_england_educational_role.order_by_user` is repeated 5 times in this file, do you think it should be put into a method like? 

```
def other_england_work_histories
  application_form
    .work_histories
    .other_england_educational_role
    .order_by_user
end
```

Hassanmir92: Actioned

syed87 (approved):
lgtm

PR #3177, opened 2026-04-23
AQTS 1426 add new country

syed87:
**Jira:** https://dfedigital.atlassian.net/browse/AQTS-1426

### What is changing
This change introduces a new “Create a new country” flow in the Support Console, allowing a new country can be created via a dedicated form, capturing the country of recognition and key eligibility configuration, and will be created as ineligible (inactive) by default.

### Why are we making this change?
Adding a new country currently requires developer involvement. This slows delivery when policy or service needs change and creates unnecessary dependency on engineering for routine configuration.
This change enables staff users with support console access to add a new country themselves.

### DIFF
--- a/app/forms/support_interface/new_country_form.rb
+++ b/app/forms/support_interface/new_country_form.rb
@@ -16,8 +16,8 @@ class SupportInterface::NewCountryForm
   validates :has_regions, inclusion: { in: [true, false] }
   validates :region_names, presence: true, if: :has_regions
 
-  def save!
-    raise ActiveRecord::RecordInvalid, self if invalid?
+  def save
+    return false if invalid?
 
     country =
       Country.new(
--- a/app/forms/support_interface/new_country_form.rb
+++ b/app/forms/support_interface/new_country_form.rb
@@ -9,10 +9,7 @@ class SupportInterface::NewCountryForm
   attribute :has_regions, :boolean
   attribute :region_names, :string
 
-  validates :location,
-            presence: {
-              message: "Select the country of recognition",
-            }
+  validates :location, presence: true
   validate :location_must_be_valid_country, if: -> { location.present? }
   validate :country_must_not_already_exist, if: -> { location.present? }
   validates :eligibility_route, inclusion: { in: %w[expanded reduced standard] }
@@ -54,17 +51,12 @@ class SupportInterface::NewCountryForm
   private
 
   def location_must_be_valid_country
-    unless Country::CODES.include?(country_code)
-      errors.add(:location, "Select the country of recognition")
-    end
+    errors.add(:location, :invalid) unless Country::CODES.include?(country_code)
   end
 
   def country_must_not_already_exist
     if Country.exists?(code: country_code)
-      errors.add(
-        :location,
-        "You must select a country that does not already exist",
-      )
+      errors.add(:location, :already_exists)
     end
   end
 end
--- a/app/forms/support_interface/new_country_form.rb
+++ b/app/forms/support_interface/new_country_form.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+class SupportInterface::NewCountryForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :location, :string
+  attribute :eligibility_route, :string
+  attribute :has_regions, :boolean
+  attribute :region_names, :string
+
+  validates :location,
+            presence: {
+              message: "Select the country of recognition",
+            }
+  validate :location_must_be_valid_country, if: -> { location.present? }
+  validate :country_must_not_already_exist, if: -> { location.present? }
+  validates :eligibility_route, inclusion: { in: %w[expanded reduced standard] }
+  validates :has_regions, inclusion: { in: [true, false] }
+  validates :region_names, presence: true, if: :has_regions
+
+  def save!
+    raise ActiveRecord::RecordInvalid, self if invalid?
+
+    country =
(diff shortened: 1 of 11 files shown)


app/forms/support_interface/new_country_form.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def country_code
+    CountryCode.from_location(location)
+  end
+
+  private
+
+  def location_must_be_valid_country
+    unless Country::CODES.include?(country_code)
+      errors.add(:location, "Select the country of recognition")
```
Hassanmir92: Could the error messages be keys such as :blank or :already_exists and then define the error messages in the yml file?

app/forms/support_interface/new_country_form.rb
```diff
@@ -0,0 +1,70 @@
...
+  attribute :has_regions, :boolean
+  attribute :region_names, :string
+
+  validates :location,
+            presence: {
+              message: "Select the country of recognition",
+            }
+  validate :location_must_be_valid_country, if: -> { location.present? }
+  validate :country_must_not_already_exist, if: -> { location.present? }
+  validates :eligibility_route, inclusion: { in: %w[expanded reduced standard] }
```
Hassanmir92, line 15: Make sure error messages are all user friendly and defined in the relevant yml files.

app/views/support_interface/countries/new.html.erb
```diff
@@ -0,0 +1,38 @@
+<% content_for :page_title, title_with_error_prefix("Create a new country", error: @form.errors.any?) %>
```
Hassanmir92, line 2: Could you add "Back" buttons on all the countries pages. I think they're missing by default. Both new and edit page.

app/forms/support_interface/new_country_form.rb
```diff
@@ -0,0 +1,62 @@
...
+  attribute :eligibility_route, :string
+  attribute :has_regions, :boolean
+  attribute :region_names, :string
+
+  validates :location, presence: true
+  validate :location_must_be_valid_country, if: -> { location.present? }
+  validate :country_must_not_already_exist, if: -> { location.present? }
+  validates :eligibility_route, inclusion: { in: %w[expanded reduced standard] }
+  validates :has_regions, inclusion: { in: [true, false] }
+  validates :region_names, presence: true, if: :has_regions
```
Hassanmir92, line 17: You need an appropriate error message for this. Currently not set in yml file

spec/forms/support_interface/new_country_form_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+    end
+
+    context "when form is invalid" do
+      let(:location) { "" }
+
+      it "raises ActiveRecord::RecordInvalid" do
+        expect { subject }.to raise_error(ActiveRecord::RecordInvalid)
+      end
+
+      it "does not create a country" do
```
Hassanmir92, line 152: I think here, I would recommend the following if you make the change in the form:

```ruby
      expect {
          subject
      }.not_to change(Country, :count)
```

However, if you do want to still go with the `save!` and raise the error, then this will work:

```ruby
      expect {
        begin
          subject
        rescue ActiveRecord::RecordInvalid
          # Do Nothing
        end
      }.not_to change(Country, :count)
```

app/forms/support_interface/new_country_form.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def save!
+    raise ActiveRecord::RecordInvalid, self if invalid?
```
Hassanmir92: I recommend a different approach here.

If you do:

```ruby
return false if invalid?
```

and also change the method name to `save` without the `!` (since you're not raising anymore).

Then in the controller, instead of what you have now, you can do:

```ruby
      authorize [:support_interface, Country]
      @form = SupportInterface::NewCountryForm.new(new_country_params)
      if @form.save
        redirect_to %i[support_interface countries]
      else
        render :new, status: :unprocessable_entity
      end

```

This then simplifies your test as well possible.

app/forms/support_interface/new_country_form.rb
```diff
@@ -0,0 +1,62 @@
...
+    country =
+      Country.new(
+        code: country_code,
+        eligibility_enabled: false,
+        eligibility_skip_questions: eligibility_route == "reduced",
+        subject_limited: eligibility_route == "expanded",
+      )
+
+    ActiveRecord::Base.transaction do
+      country.save!
```
syed87, line 31: just check this save!, the linter added it back in

spec/system/support_interface/countries_spec.rb
```diff
@@ -62,6 +62,16 @@
...
   end
 
+  it "allows visiting new country page" do
+    given_countries_exist
+    given_i_am_authorized_as_a_support_user
+    when_i_visit_the_countries_page
+
+    expect(page).to have_link("Add a new country")
+    and_i_click_new_country
+    then_i_see_the_new_country_page
```
Hassanmir92, line 72: I still don't see this finished? 

Shouldn't there be specs on you filling out the form and submitting and seeing the country added? 

I expect to probably see 2. One where you add multiple regions and one without.

Hassanmir92 (approved):
Great work mate. Thanks for addressing all comments. 

Only thing is that we're not using page objects on the system specs for support interface countries spec, but as discussed you just wanted to remain consistent with how tests are written in that file which is fair. 

I'll create a tech debt ticket to sort this out at a later point.

PR #3184, opened 2026-04-29
[AQTS-1455] Rails configuration for forcing ssl and browser Cache-Control

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1455

Uncommenting config.force_ssl = true in `config/environments/production.rb` to explicitly enforce HTTPS, HSTS, and secure cookie flags. This ensures all traffic is served over HTTPS and enables HSTS, instructing browsers to only connect via secure channels.

Adding a before_action in ApplicationController to set `Cache-Control: no-store` on all responses, preventing browsers from caching pages. this prevents sensitive personal data from being stored in browser cache, reducing risk on shared or public devices.


syed87 (approved):
lgtm

PR #3213, opened 2026-05-21
AQTS 1344 Updating wording for prioritisation reference

syed87:
link: https://dfedigital.atlassian.net/browse/AQTS-1344

**Why**

The email sent to prioritisation referees was updated to use the term “work experience”, but parts of the service and CMS still used “work history”.
This caused inconsistent wording across the journey and could be confusing for users.

**What**

Updated “work history” to “work experience” in the prioritisation reference journey:

About the applicant page
Check your answers page
Validation error message

Updated the same wording in CMS:

Pre‑assessment task names
Reference review screens
Prioritisation decision screens (accepted and rejected)


Hassanmir92 (commented):
Do we need to change "work history and experience" inside the prioritisation_reference_reminder email?

Hassanmir92 (approved):

PR #3219, opened 2026-05-26
Adding ADR for tracking email deliveries

Hassanmir92:
This PR adds a new ADR document for our decisions made for tracking email deliveries on our service.

In addition to the ADR I also ensure that we get errors if mail-notify version is updated with the fix we've already created a patch for, and if so we could then remove our monkey patch.


syed87 (approved):
lgtm

syed87 (approved):
lgtm

PR #3225, opened 2026-05-30
AQTS 1421 check your answers task

syed87:
**Jira:** [https://dfedigital.atlassian.net/browse/AQTS-1421](https://dfedigital.atlassian.net/jira/software/projects/AQTS/boards/207?selectedIssue=AQTS-1421)

### What is changing
This PR introduces a task-based "Check your answers" section to the consent request flow. 

### Why are we making this change?
Currently, the layout of the consent request flow causes confusion for applicants. Many mistakenly believe they have completed the process after uploading their documents, missing the final step to review and submit their answers.

By adopting the task-list pattern previously successfully implemented for Further Information (FI) requests, we are making the final review step an explicit task. This standardizes the user experience across the service, prevents premature drop-offs, and ensures applicants explicitly review and submit their consent documents.

### DIFF
--- a/app/view_objects/teacher_interface/consent_requests_view_object.rb
+++ b/app/view_objects/teacher_interface/consent_requests_view_object.rb
@@ -38,6 +38,34 @@ module TeacherInterface
       end
     end
 
+    def check_your_answers_task_group
+      {
+        heading:
+          I18n.t(
+            :check_your_answers,
+            scope: %i[
+              teacher_interface
+              further_information_request
+              show
+              section
+            ],
+          ),
+        items: [
+          {
+            title:
+              I18n.t(
+                "teacher_interface.further_information_request.show.check",
+              ),
+            href:
+              if can_submit?
+                %i[check teacher_interface application_form consent_requests]
+              end,
+            status: can_submit? ? "not_started" : "cannot_start",
+          },
+        ],
+      }
+    end
+
     private
 
     attr_reader :application_form
--- a/app/view_objects/teacher_interface/consent_requests_view_object.rb
+++ b/app/view_objects/teacher_interface/consent_requests_view_object.rb
@@ -24,8 +24,6 @@ module TeacherInterface
       end
     end
 
-    alias_method :can_check_answers?, :can_submit?
-
     def check_your_answers_fields
       consent_requests.each_with_object({}) do |consent_request, memo|
         memo[consent_request.id] = {
@@ -51,10 +49,10 @@ module TeacherInterface
           {
             title: I18n.t("teacher_interface.consent_requests.index.check"),
             href:
-              if can_check_answers?
+              if can_submit?
                 %i[check teacher_interface application_form consent_requests]
               end,
-            status: can_check_answers? ? "not_started" : "cannot_start",
+            status: can_submit? ? "not_started" : "cannot_start",
           },
         ],
       }
--- a/app/view_objects/teacher_interface/consent_requests_view_object.rb
+++ b/app/view_objects/teacher_interface/consent_requests_view_object.rb
@@ -24,6 +24,8 @@ module TeacherInterface
       end
     end
 
+    alias_method :can_check_answers?, :can_submit?
+
     def check_your_answers_fields
       consent_requests.each_with_object({}) do |consent_request, memo|
         memo[consent_request.id] = {
@@ -49,10 +51,10 @@ module TeacherInterface
           {
             title: I18n.t("teacher_interface.consent_requests.index.check"),
             href:
-              if can_submit?
+              if can_check_answers?
                 %i[check teacher_interface application_form consent_requests]
               end,
-            status: can_submit? ? "not_started" : "cannot_start",
+            status: can_check_answers? ? "not_started" : "cannot_start",
           },
         ],
       }
--- a/app/view_objects/teacher_interface/consent_requests_view_object.rb
+++ b/app/view_objects/teacher_interface/consent_requests_view_object.rb
@@ -45,8 +45,8 @@ module TeacherInterface
             :check_your_answers,
             scope: %i[
               teacher_interface
-              further_information_request
-              show
+              consent_requests
+              index
               section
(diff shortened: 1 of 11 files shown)


app/view_objects/teacher_interface/consent_requests_view_object.rb
```diff
@@ -24,6 +24,8 @@ def can_submit?
       end
     end
 
+    alias_method :can_check_answers?, :can_submit?
```
Hassanmir92: Is this necessary?

We could just use `can_submit?` in the `check_your_answers_task_group`?

app/views/teacher_interface/consent_requests/index.html.erb
Hassanmir92: I think this should probably also be a h2, similar to the check your answers? As right now we have a page with incorrect order of heading tags.

Hassanmir92: Given that you now have "Check your answers" as a task item, this Continue button becomes redundant.

spec/system/teacher_interface/consent_spec.rb
```diff
@@ -88,7 +88,7 @@ def and_i_see_sign_out_content
 
   def and_i_see_the_download_and_upload_tasks
     task_lists = teacher_consent_requests_page.task_lists
-    expect(task_lists.count).to eq(1)
+    expect(task_lists.count).to eq(2)
```
Hassanmir92, line 91: I would have expected more updates to the system spec to reflect the new change of clicking the task instead of clicking "Continue". 

I suspect by removing the "Continue" button as mentioned on my previous comment, you will actually have this become a failing system spec.

spec/view_objects/teacher_interface/consent_requests_view_object_spec.rb
```diff
@@ -134,4 +134,88 @@
...
+          unsigned_document_downloaded: true,
+        )
+      end
+
+      before do
+        create(
+          :upload,
+          :clean,
+          document: consent_request.signed_consent_document,
+        )
```
Hassanmir92, line 174: Instead of this, I think you can add `with_signed_upload` as a trait to your `consent_requests` which will avoid having to add this `before do`

spec/view_objects/teacher_interface/consent_requests_view_object_spec.rb
```diff
@@ -134,4 +134,88 @@
...
+
+    context "when items are complete" do
+      before do
+        consent_request.update!(unsigned_document_downloaded: true)
+        create(
+          :upload,
+          :clean,
+          document: consent_request.signed_consent_document,
+        )
+      end
```
Hassanmir92, line 278: Similar to a previous comment around using `with_signed_upload`

What you can do is add a before do in the "when items are incomplete" context as well:

```ruby

before do
        create(
        :consent_request,
        :requested,
        assessment: application_form.assessment,
      )
end
```

and then inside this context have:

```ruby
before do
        create(
        :consent_request,
        :requested,
        :with_signed_upload,
        assessment: application_form.assessment,
      )
end 
```

Might be more clear?

spec/view_objects/teacher_interface/consent_requests_view_object_spec.rb
```diff
@@ -134,4 +134,88 @@
...
+      it "disables check your answers in task list group" do
+        expect(check_your_answers_task_group[:heading]).to eq(
+          "Check your answers",
+        )
+        expect(check_your_answers_task_group[:items].first[:status]).to eq(
+          "cannot_start",
+        )
+        expect(check_your_answers_task_group[:items].first[:href]).to be_nil
+      end
+    end
```
Hassanmir92, line 246: I would recommend adding another context "when the unsigned document has been download but signed consent not uploaded"?

Hassanmir92: I would also consider having a scenario for "with multiple consent requests"?

spec/spec_helper.rb
```diff
@@ -20,6 +20,10 @@
 require "rspec/core/formatters/base_text_formatter"
 
 RSpec.configure do |config|
+  if ENV["COVERAGE"]
+    require "simplecov"
+    SimpleCov.start "rails"
+  end
```
Hassanmir92: Is this meant for this PR?

Hassanmir92 (commented):
Nice work mate :) 

Great consistent approach to how this is handled to the further information requests. I've left some comments for your consideration.

spec/view_objects/teacher_interface/consent_requests_view_object_spec.rb
```diff
@@ -134,4 +134,142 @@
...
+          :requested,
+          :with_signed_upload,
+          assessment: application_form.assessment,
+          unsigned_document_downloaded: true,
+        )
+      end
+
+      it { is_expected.to be true }
+    end
+  end
```
Hassanmir92, line 221: I think this test could make use of some additional contexts similar to below such as:

1. Scenario for when unsigned document downloaded but without a signed upload.
2. Multiple consent with one completed but other one not?
3. Both completed?

Hassanmir92 (commented):
Really nice work mate.

Left one small comment which is a nice improvement in test coverage.

app/views/teacher_interface/consent_requests/index.html.erb
```diff
@@ -4,7 +4,7 @@
 <h1 class="govuk-heading-xl">Consent documents overview</h1>
```
Hassanmir92: I believe AC7 is missing here.

Hassanmir92:
```
<span class="govuk-caption-l">Your application</span>
<h1 class="govuk-heading-l">Consent documents overview</h1>
```

It might be that the heading needs to be govuk-heading-l instead of xl

Hassanmir92 (approved):

PR #3241, opened 2026-06-11
[AQTS-1478] Show email delivery failures on CMS

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1478

## What

This add an "Email failures" page to the assessor CMS so staff can see which Notify emails failed to deliver. The page shows failed deliveries for the current and previous week (Mon-Sun) across two tabs, each labelled with its date range and failure count, plus a banner summarising last week's delivery rate ("X out of Y emails failed to send between … and …"). We currently provide this to managers via a manual report every week and this work eliminates the manual workaround.

The `EmailDeliveryFailuresIndexViewObject` is where most of the logic lives as it builds the tab data, counts, and paginated records. Performance has been considered to ensure the page stays efficient and preventing N+1 queries.

--- 

Other notes:

I considered a partial index (`notify_completed_at` scoped to the failure statuses) but deliberately chose not include it here. At current volumes in production, the query completes in well under a millisecond and the planner would prefer a seq scan at this size anyway. The index is cheap to add when justified, so the call is to defer it until the table is materially larger and the page is confirmed as a bottleneck, rather than index on a guess.

### DIFF
--- a/app/view_objects/assessor_interface/email_delivery_failures_index_view_object.rb
+++ b/app/view_objects/assessor_interface/email_delivery_failures_index_view_object.rb
@@ -70,6 +70,7 @@ class AssessorInterface::EmailDeliveryFailuresIndexViewObject
           notify_completed_at: :desc,
         ),
         count: active_tab_failures_count,
+        limit: 100,
       )
   end
 
--- a/app/view_objects/assessor_interface/email_delivery_failures_index_view_object.rb
+++ b/app/view_objects/assessor_interface/email_delivery_failures_index_view_object.rb
@@ -17,15 +17,15 @@ class AssessorInterface::EmailDeliveryFailuresIndexViewObject
   end
 
   def this_week_tab_name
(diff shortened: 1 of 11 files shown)


app/view_objects/assessor_interface/email_delivery_failures_index_view_object.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+class AssessorInterface::EmailDeliveryFailuresIndexViewObject
+  include ActionView::Helpers::FormOptionsHelper
```
syed87: This helper module might not been needed as none of it's methods are used in this file.

syed87 (approved):
lgtm

app/view_objects/assessor_interface/email_delivery_failures_index_view_object.rb
Hassanmir92: Absolutely correct :) Removed.

PR #3250, opened 2026-06-17
[AQTS-1496] Fix issue with prioritisation reference reminder to applicant

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1496

## Problem
Our service sends reminders for prioritisation requests to applicants and their referees if they're still outstanding. For the rare scenario of multiple prioritisation reference requests on a single application, our assessors can make a decision on the prioritisation status even if only 1 has been received. Once the assessor makes the prioritisation decision, the other outstanding requests can no longer be filled out and the referees no longer receive reminders. However, the applicant still seems to be receiving these reminders which they shouldn't be. 

## Change
Updating `#should_send_reminder_email?` method on `ApplicationForm` so that once a prioritisation decision has been made, no prioritisation reference reminders will be sent to applicants. This method is responsible for handling reminders directly to the applicant while the `should_send_reminder_email?` on `PrioritisationReferenceRequest` is responsible determining on whether to send the reminders to the referee themselves.

### DIFF
--- a/app/models/application_form.rb
+++ b/app/models/application_form.rb
@@ -332,6 +332,8 @@ class ApplicationForm < ApplicationRecord
         )
       end
     when "prioritisation_references"
+      return false if assessment.prioritisation_decision_at.present?
+
       prioritisation_reference_requests_not_yet_received_or_rejected.any? do |prioritisation_reference_request|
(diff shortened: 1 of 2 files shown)


syed87 (approved):
lgtm

PR #3252, opened 2026-06-18
AQTS 1484 Update sign in email content

syed87:
https://dfedigital.atlassian.net/browse/AQTS-1484

This pr change removes two versions of the sign in email for first time sign in and following sign ins to have just one email version and alters the text to reduce the number of zendesk replies to this email.

By adding inset text to the sign in link it should be clearer to the user how to sign in


Hassanmir92 (approved):
Nice and simple :)

PR #3258, opened 2026-06-22
[AQTS-1493] Remove PDF generation of unsigned consent

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1493

### Context

Following our [decision](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/adr/00013-pdf-generation.md) to replace PDF generation server side with "Print this page" functionality, this PR removes the consent letter PDF generation feature and instead allows admins to save PDF via "Print this page" functionality.

### Changes

1. Introduce the new "Print this page" button with JS controller for the functionality.
2. Ensure correct elements (excluding header, navbar and footer) are removed from printable version.
3. Delete ConsentLetter class
4. Remove `prawn` as a dependency from our Gemfile.


syed87 (approved):
lgtm

PR #3262, opened 2026-06-24
AQTS 1514 double redirect bug

syed87:
**ticket:** https://dfedigital.atlassian.net/browse/AQTS-1514

This ticket address the bug found in sentry for a double redirect and refactors code to be cleaner by removing the 2nd redirect

### DIFF
--- a/spec/requests/teacher_interface/consent_requests_spec.rb
+++ b/spec/requests/teacher_interface/consent_requests_spec.rb
@@ -58,7 +58,7 @@ RSpec.describe "Teacher Interface - Consent Requests", type: :request do
       end
     end
 
-    context "when both consent requests are requested" do
+    context "when both consent requests are not received" do
       before do
         create(:requested_consent_request, assessment:)
         create(:requested_consent_request, assessment:)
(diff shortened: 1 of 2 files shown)


spec/requests/teacher_interface/consent_requests_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        perform
+      end
+
+      it "does not attempt to receive the requestables again" do
+        expect(ReceiveRequestable).not_to receive(:call)
+        perform
+      end
+    end
+
+    context "when there are unreceived consent requests" do
```
Hassanmir92: This is not really a real scenario. I think we can remove it.

spec/requests/teacher_interface/consent_requests_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+           )
+    end
+
+    context "when all consent requests are already received" do
+      before do
+        create(:received_consent_request, assessment:)
+        create(:received_consent_request, assessment:)
+      end
+
+      it "redirects safely to the application form without DoubleRenderError" do
```
Hassanmir92, line 28: It's probably better to just say "redirects safely to the application form". We don't need the "without DoubleRenderError" as that was just the consequence of the previous code which now your test covers.

Hassanmir92 (approved):
Nice work :)

PR #3263, opened 2026-06-25
[AQTS-1528] Filter by region state in CMS

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1528

Screenshot 2026-06-25 at 14 23 53

Adds a State/territory filter to the CMS application list. When a country that has multiple regions is selected in the existing country filter, a second filter appears letting you narrow results down to a specific region within that country.

Countries with multiple regions (e.g. those where QTS eligibility or assessment differs by state/province) couldn't be filtered any further than country level. Assessors had to scan the full list to find applications for a particular region. This adds the missing level of granularity.

- Region filter is hidden by default.
- It only renders once a country with multiple regions is the active country filter.
- Selecting a different country clears any previously applied region filter.

### DIFF
--- a/app/forms/assessor_interface/filter_form.rb
+++ b/app/forms/assessor_interface/filter_form.rb
@@ -8,6 +8,7 @@ class AssessorInterface::FilterForm
                 :date_of_birth,
                 :email,
                 :location,
+                :region_ids,
                 :name,
                 :reference,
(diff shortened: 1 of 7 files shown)


syed87 (approved):
lgtm

PR #3267, opened 2026-06-26
Remove access to default device session pages that are not used

Hassanmir92:
We had access to our old sign in page still on the service. This page should not be in use as users use Entra ID as the way to authenticate. This PR removes it to avoid any spam requests.


syed87 (approved):
lgtm

PR #3269, opened 2026-06-30
[AQTS-1544] Remove differentiation of redirect on sign in for magic link flow

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1544

This PR addresses ensures there's no differentiation between whether an application exists or not for the user with the magic link authentication flow.


syed87 (approved):
lgtm

PR #3271, opened 2026-07-02
AQTS 1515 FI submit bug

syed87:
**Ticket:**
https://dfedigital.atlassian.net/browse/AQTS-1515

### What

Updated the Further Information (FI) submission flow to handle duplicate submissions gracefully.
If an assessor submits the FI request twice it is only processed once and the user is redirected to the Application Overview page.

### Why

Previously, a double submission resulted in a 500 "Something went wrong" error page.
Redirecting users to the Application Overview page after a successful submission provides a smoother user experience and prevents unnecessary errors.


Hassanmir92 (approved):
Nice work mate :)

PR #3273, opened 2026-07-02
[AQTS-1550] Ensure personas can never be enabled in production

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1550

This is an additional layer so that we never allow the `/personas` feature to become active due to human error within production.


syed87 (approved):
lgtm

PR #3277, opened 2026-07-06
[AQTS-1548] Harden staff Entra ID sign-in identity matching

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1548

Staff sign in already relies on our own single DfE tenant being used in the OmniAuth strategy, so cross-tenant sign-in was never possible. This change enhances the callback on top of that existing control to make identity matching extra bullet-proof.

## Change

* Match staff by the immutable Entra identifier (tid + oid) first, rather than by email. Email is now only used for first-time linking of an account that is invited.
* Refuse to relink once an account is already bound to a uid, so an email that later points at a different Entra identity can't attach to an existing record.

### DIFF
--- a/app/controllers/staff/omniauth_callbacks_controller.rb
+++ b/app/controllers/staff/omniauth_callbacks_controller.rb
@@ -8,17 +8,32 @@ class Staff::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     email = auth["info"]["email"]
     azure_ad_uid = auth["uid"]
 
-    @staff = Staff.find_by_email(email)
+    return fail_sign_in if azure_ad_uid.blank?
 
-    if @staff&.update(azure_ad_uid:)
-      @staff.confirm unless @staff.confirmed?
-      sign_in_and_redirect @staff, event: :authentication
+    staff = Staff.find_by(azure_ad_uid:)
+
+    if staff.nil?
+      staff = Staff.find_by_email(email)
+
+      if staff.present? && staff.azure_ad_uid.nil?
+        staff.update!(azure_ad_uid:)
+        staff.confirm unless staff.confirmed?
+
+        sign_in_and_redirect staff, event: :authentication
+      else
+        fail_sign_in
+      end
     else
-      redirect_to new_staff_session_url, alert: t(".failure")
+      staff.confirm unless staff.confirmed?
+      sign_in_and_redirect staff, event: :authentication
     end
   end
 
   protected
+
+  def fail_sign_in
+    redirect_to new_staff_session_url, alert: t(".failure")
+  end
 
   def after_sign_in_path_for(resource)
     stored_location_for(resource) || assessor_interface_root_path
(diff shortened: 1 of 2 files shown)


app/controllers/staff/omniauth_callbacks_controller.rb
```diff
@@ -8,18 +8,33 @@ def entra_id
...
-    @staff = Staff.find_by_email(email)
+    return fail_sign_in if azure_ad_uid.blank?
 
-    if @staff&.update(azure_ad_uid:)
-      @staff.confirm unless @staff.confirmed?
-      sign_in_and_redirect @staff, event: :authentication
+    staff = Staff.find_by(azure_ad_uid:)
+
+    if staff.nil?
+      staff = Staff.find_by_email(email)
```
elceebee: Do we care if the azure_ad_uid has already been assigned? If we want to assure that it's the first time the user has signed in, I think we want: 
`Staff.find_by(email:, azure_ad_uid: nil)`

elceebee, line 16: Ah! Never mind, you are doing that on line 18! Ignore me :)

elceebee (approved):
Nice!

app/controllers/staff/omniauth_callbacks_controller.rb
```diff
@@ -8,18 +8,33 @@ def entra_id
...
+
+      if staff.present? && staff.azure_ad_uid.nil?
```
syed87, line 18: `if staff.present? && !staff.archived? && staff.azure_ad_uid.nil?`

This will prevent staff who were invited but didn't complete the invitation from signing in, currently they will still get assigned a uuid when they try to sign in but will not be authorised to visit any pages

app/controllers/staff/omniauth_callbacks_controller.rb
```diff
@@ -8,18 +8,33 @@ def entra_id
...
+        staff.update!(azure_ad_uid:)
+        staff.confirm unless staff.confirmed?
+
+        sign_in_and_redirect staff, event: :authentication
+      else
+        fail_sign_in
+      end
     else
```
syed87, line 26: `return fail_sign_in if staff.archived?`

This will stop archived users from signing in even though they don't have authorisation to visit any pages

syed87:
`def active_for_authentication?
  super && !archived?
end
`

this is a 2nd option for preventing archived users from authenticating

app/controllers/staff/omniauth_callbacks_controller.rb
Hassanmir92: This is correct behaviour. They still get signed in, they're just not authorised to visit the page.

app/controllers/staff/omniauth_callbacks_controller.rb
Hassanmir92: Similar to the above. Authorisation and Authentication are separate concerns. I wouldn't recommend we mix them up. It's expected behaviour that the archived user authenticates, but they simply are not authorised to view anything.

syed87 (approved):
lgtm

PR #3281, opened 2026-07-06
When authentication fails with Entra ID, there is no flash message

Hassanmir92:
Spec fix with flash message on authentication failure


syed87 (approved):
lgtm

PR #3282, opened 2026-07-07
AQTS-1536-remove-MailDeliveryFailure

syed87:


app/views/anonymous/test_notify_error.text.erb
```diff
@@ -1 +0,0 @@
-This is used by the ApplicationMailer spec.
```
Hassanmir92, line 1: Didn't even realise this was here! 

Good thing it's being removed :)

PR #3288, opened 2026-07-07
AQTS-1536-Removing-MailDeliveryFailure

syed87:
**Ticket:**
https://dfedigital.atlassian.net/browse/AQTS-1536

### Why this change is being made
MailDeliveryFailure did not fully meet business requirements, which was to track all email delivery failures (including those sent to applicants and referees). EmailDelivery is now in place to prevent invalid email formats from being submitted in the first place.

### What has changed

Removed the MailDeliveryFailure model and corresponding database table.
Deleted any associated logic.


Hassanmir92 (approved):
Nice work :)

PR #3289, opened 2026-07-07
[AQTS-1543] Remove old code for sign in access to CMS

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1543

Staff authentication is handled entirely by Entra ID, but the default Devise flows (password sign-in, password reset, invitation confirmation, account unlock, remember-me) were left over from before. They serve no purpose and risk creating a route into the CMS that bypasses Entra ID and therefore 2FA. This removes that surface so Entra ID is the only way staff can sign in throughout all environments.

## Changes

* Removed all Devise staff routes except sign-out and staff invitations (new/create). Manual sign-in, password reset, confirmation and unlock routes no longer exist.
* Controllers/views: deleted the passwords, confirmations and unlocks controllers and views. Adjustments to the sessions controller to the only actions still used (destroy, signed_out) and removed the invitation set-password page.
*  Removed :recoverable, :rememberable, :lockable from Staff model. This is all redundant with Entra ID.
* Staff invitation email now always points to the Entra ID sign-in path (no more password-set link).
* Removed the EnforceEntraIdSignIn guard, which existed only to block the manual routes that are now gone.
* Entra ID callback failures redirect to root.

## Out of scope

* No database columns dropped. The now-unused Devise columns (reset_password_token, reset_password_sent_at, remember_created_at, failed_attempts, unlock_token, locked_at) stay for now and will be removed in a separate phase.
* Removing :database_authenticatable and :confirmable. The confirmable is still used in the Entra ID callback and database_authenticable may cause session issues and if we were to remove it, we'd want that to be a change of its own.


app/controllers/concerns/streamed_response_authenticatable.rb
```diff
@@ -10,7 +10,7 @@ module StreamedResponseAuthenticatable
   # and if the user is not authentication Devise will throw :warden
 
   def authenticate_or_redirect(scope)
-    redirect_to [:new, scope, :session] unless authenticated_user(scope)
+    redirect_to failure_redirect_for(scope) unless authenticated_user(scope)
```
Hassanmir92, line 13: This needed an update since the new_staff_session path no longer exists and we force the user to `omniauth_authorize_path(:staff, :entra_id)` path instead.

app/controllers/staff/invitations_controller.rb
```diff
@@ -30,8 +30,4 @@ def configure_permitted_parameters
   def after_invite_path_for(inviter, invitee)
     invitee.is_a?(Staff) ? assessor_interface_staff_index_path : super
   end
-
-  def after_accept_path_for(resource)
```
Hassanmir92, line 34: Removed as we no longer have an invitation "accept" path.

syed87 (approved):
lgtm

PR #3290, opened 2026-07-08
Remove-archived-staff-from-personas-page

syed87:
This is a small improvement for local host and testing environment so that the personas page doesn't include any staff which have been created and then archived.

### DIFF
--- a/app/controllers/personas_controller.rb
+++ b/app/controllers/personas_controller.rb
@@ -136,7 +136,7 @@ class PersonasController < ApplicationController
         withdraw_permission: :desc,
         manage_staff_permission: :desc,
         email: :asc,
-      )
+      ).where.not(archived: true)
   end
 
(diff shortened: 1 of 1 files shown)


app/controllers/personas_controller.rb
```diff
@@ -126,17 +126,19 @@ def load_eligible_personas
...
+          change_name_permission: :desc,
+          change_work_history_and_qualification_permission: :desc,
+          reverse_decision_permission: :desc,
+          support_console_permission: :desc,
+          verify_permission: :desc,
+          withdraw_permission: :desc,
+          manage_staff_permission: :desc,
+          email: :asc,
+        )
+        .where.not(archived: true)
```
Hassanmir92, line 141: I think it's actually important to have archived in the personas. 

This is the only way to allow us to test archived users.

PR #3295, opened 2026-07-14
mvp code

syed87:


app/controllers/assessor_interface/assessment_recommendation_award_controller.rb
```diff
@@ -111,6 +111,14 @@ def update_confirm
         if @form.confirmation
           ActiveRecord::Base.transaction do
             assessment.award!
+            application_form.update!(verifier: current_staff)
+
+            CreateTimelineEvent.call(
+              "verification_decision_made",
+              application_form:,
+              user: current_staff,
+            )
```
Hassanmir92, line 120: I know it's mvp, but possibly having the lines of code as a service object might be useful? 

Something like a `AssignApplicationFormVerifier`? 

We have `AssignApplicationFormAssessor` and `AssignApplicationFormReviewer` so it can be similar to that.

PR #3296, opened 2026-07-15
AQTS-1530-verified-by-feature

syed87:
**Ticket**: https://dfedigital.atlassian.net/browse/AQTS-1530

### Why is this change being made?
Currently, there is no visibility of which admin makes the final verification decision on an application. This makes it difficult for senior managers to understand verification workloads, track decision ownership, and audit outcomes when applications progress through review or award stages.
This change introduces a new Verified by field to capture and display the admin responsible for the most recent verification decision, providing greater transparency and auditability across the verification process.

### What has changed?
Added verification decision ownership

Added a new Verified by field beneath the existing Reviewer field.
The field is automatically populated with the name of the admin who completes the Verification decision task.
Added timeline entries to record when a verification decision is made and by whom.
Screenshot 2026-07-30 at 22 07 13

### DIFF
--- a/app/controllers/assessor_interface/assessment_recommendation_award_controller.rb
+++ b/app/controllers/assessor_interface/assessment_recommendation_award_controller.rb
@@ -109,10 +109,7 @@ module AssessorInterface
 
       if @form.valid?
         if @form.confirmation
-          ActiveRecord::Base.transaction do
-            assessment.award!
-            CreateTRSTRNRequest.call(application_form:, user: current_staff)
-          end
+          SendApplicationForQTSAward.call(assessment:, user: current_staff)
 
           redirect_to [:status, :assessor_interface, application_form]
         else
--- a/app/services/assign_application_form_verifier.rb
+++ b/app/services/assign_application_form_verifier.rb
@@ -3,9 +3,8 @@
 class AssignApplicationFormVerifier
   include ServicePattern
 
-  def initialize(application_form:, user:, verifier:)
+  def initialize(application_form:, verifier:)
     @application_form = application_form
-    @user = user
     @verifier = verifier
   end
 
@@ -17,5 +16,5 @@ class AssignApplicationFormVerifier
 
   private
 
-  attr_reader :application_form, :user, :verifier
+  attr_reader :application_form, :verifier
 end
--- a/app/services/assign_application_form_verifier.rb
+++ b/app/services/assign_application_form_verifier.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class AssignApplicationFormVerifier
+  include ServicePattern
+
+  def initialize(application_form:, user:, verifier:)
+    @application_form = application_form
+    @user = user
+    @verifier = verifier
+  end
+
+  def call
+    return if application_form.verifier == verifier
+
+    application_form.update!(verifier:)
+  end
+
+  private
+
+  attr_reader :application_form, :user, :verifier
+end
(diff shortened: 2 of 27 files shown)


spec/helpers/application_form_helper_spec.rb
```diff
@@ -199,6 +199,25 @@
       end
     end
 
+    context "include_verifier false" do
```
Hassanmir92: This test should be for the context of "include_verifier true" because the default is false (opposite of the reviewer one).

app/services/assign_application_form_verifier.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class AssignApplicationFormVerifier
+  include ServicePattern
+
+  def initialize(application_form:, user:, verifier:)
```
Hassanmir92: Why is the user and verifier required? Surely it's the user that is the verifier.

I think this should only take in `application_form:, user:` and then verifier is the user that's passed in. 

This is different to assigning assessor/reviewer since those can be assigned by a specific user, but in this case it's always assigned by the user themselves right? So we don't need both.

app/controllers/assessor_interface/assessment_recommendation_award_controller.rb
```diff
@@ -109,10 +109,7 @@ def update_confirm
 
       if @form.valid?
         if @form.confirmation
-          ActiveRecord::Base.transaction do
-            assessment.award!
-            CreateTRSTRNRequest.call(application_form:, user: current_staff)
-          end
+          SendApplicationForQTSAward.call(assessment:, user: current_staff)
```
Hassanmir92, line 112: 💪

app/services/send_application_for_qts_award.rb
```diff
@@ -0,0 +1,28 @@
...
+class SendApplicationForQTSAward
+  include ServicePattern
+
+  def initialize(assessment:, user:)
+    @assessment = assessment
+    @user = user
+  end
+
+  def call
+    in_verify_stage = assessment.verify?
```
Hassanmir92: This should be a private method I think and instead trigger the `AssignApplicationFormVerifier` before the `assessment.award!`?.

Or you could instead rely on the application forms stage. So instead of doing `assessment.verify?`, you do `application_form.verification_stage?`. 

```ruby
def application_form_in_verification_stage?
  application_form.verification_stage?
end
```

Hassanmir92 (commented):
Great work mate. 

Just one comment around alternative ways to determine whether an application is being awarded from verification stage. Either will help improve part of the code.

spec/system/assessor_interface/completing_assessment_spec.rb
```diff
@@ -231,6 +231,7 @@ def given_there_is_an_awardable_application_form
 
     application_form.assessment.verify!
     application_form.assessment.update!(induction_required: false)
+    ApplicationFormStatusUpdater.call(application_form:, user: assessor_user)
```
Hassanmir92: Instead of running this, can you not just update the stage manually?

syed87: yeah I can, I just thought this was more realistic as this is how it's done in the codebase

spec/services/review_assessment_spec.rb
```diff
@@ -24,4 +24,9 @@
   it "changes the application form stage" do
     expect { call }.to change(application_form, :stage).to("review")
   end
+
+  it "assigns the verifier" do
+    call
+    expect(application_form.verifier).to eq(user)
+  end
```
Hassanmir92, line 30: For consistency with above, would it make change to be the following?

```ruby
expect { call }.to change(application_form, :verifier).to(user)
```

Hassanmir92 (approved):
Great work mate :)

PR #3301, opened 2026-07-20
[AQTS-1569] Remove auto relink to re-registered GOV.UK One Login accounts

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1569

## Why

Applicant GOV.UK One Login callbacks matched on email and would re-link whatever One Login UID everytime. That means a re-registered One Login account with a new, different UID but with the same email, would silently be linked to an existing application. This was originally intentional as this was something GOV.UK One Login allowed and was happening regularly, but we have changed our service's tolerance for re-linking. This change makes the immutable One Login UID the primary identifier and refuses to relink an application that's already bound to a different UID.

## What

Callback logic:

1. Match by UID -> sign in applicant.
2. No UID match:
  a) Match by email with no linked UID -> first-time GOV.UK One Login use, link the UID and sign in
  b) Match by email already linked to a different UID -> refuse sign-in and no UID re-link. Logged in Sentry as error.
  c) No email match -> create a new application, link the UID, and sign in

### DIFF
--- a/app/controllers/teachers/omniauth_callbacks_controller.rb
+++ b/app/controllers/teachers/omniauth_callbacks_controller.rb
@@ -17,7 +17,7 @@ class Teachers::OmniauthCallbacksController < ApplicationController
       session[:id_token] = auth&.credentials&.id_token
 
       teacher =
-        FindOrCreateTeacherFromGovOne.call(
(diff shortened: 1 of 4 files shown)


syed87 (approved):
lgtm

PR #3303, opened 2026-07-22
[AQTS-1584] Fix edge-case when applicant goes to decision review request when in draft application

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1586

We've been seeing some errors related to an applicant who has previously been declined and created a new draft application. After doing so, they attempted to submit a decision review using the link they received in their email. However, once they go to the decision review page, they see a 500 error which we have observed within Sentry. 

This PR addresses the issue and ensures they're taken to their draft application instead.


syed87 (approved):
lgtm

PR #3312, opened 2026-07-27
[AQTS-1590] Introducing SolidCache

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1590

## Context

This is our first step of the "remove Redis" work. Redis currently backs two things: the Sidekiq queue and `Rails.cache`. This PR moves `Rails.cache` to Solid Cache used by RackAttack (we have no other custom caching on our service).

## Change

- Adding `solid_cache` gem and the `solid_cache_entries` migration (Excluding `solid_cache_entries` from dfe-analytics streaming).
- Updating `config.cache_store` to `:solid_cache_store` in production.
- We are running the cache on the same primary database (no separate cache DB).
- config/cache.yml: Maximum age of entries will be 7 days, the maximum size of the cache entries will be 256MB and expiry method is set to be triggered as a background job instead of within a puma thread.

## Rollout

- Test in review environment to ensure records are being created, counters incremented and expired for RackAttack entries.
- In the case we need to roll it back, it'll be a one-line revert of `config.cache_store` to `:redis_cache_store`.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -62,7 +62,7 @@ Rails.application.configure do
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL") }
+  config.cache_store = :solid_cache_store
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
   config.active_job.queue_adapter = :sidekiq
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ gem "sentry-ruby"
 gem "sidekiq", "<7"
 gem "sidekiq-cron"
 gem "sitemap_generator"
+gem "solid_cache"
 gem "stimulus-rails"
 gem "validate_url"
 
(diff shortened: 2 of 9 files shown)


CatalinVoineag (approved):
Looks good 👍🏻 Just 1 suggestion

config/environments/production.rb
```diff
@@ -62,7 +62,7 @@
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL") }
+  config.cache_store = :solid_cache_store
```
CatalinVoineag, line 65: This can also be added to [development.rb](https://github.com/DFE-Digital/publish-teacher-training/blob/main/config/environments/development.rb#L28)

So your dev env uses the same setup.

When you run `bin/rails dev:cache` on your local machine it should cache things in your local DB. It's a good way to test things. And when you're done just run `bin/rails dev:cache` again

Hassanmir92: Solid advice :) 

Addressed: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/3312/commits/dc78d7ecbe154237130910a377df3f634678f36a

syed87 (approved):
lgtm

PR #3316, opened 2026-07-29
[AQTS-1539] Introducing SolidQueue with limited jobs

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1539

## Change

First steps of the Solid Queue migration: 
1. Introducing Solid Queue (gem, data migration, queue config) - Configured to run on our primary database (no separate queue DB).
2. Adding new Terraform solid queue worker app with 2Gi memory in production. Once we fully make the switch, we'll remove this and update the "worker_application" module to have the solid_queue command instead. Based on our observations on how much the worker ends up consuming memory, we will scale down if 2Gi ends up being unnecessary.
4. Introducing Mission Control UI for managing jobs similar to Sidekiq Web UI.
5. Deliberately chosen a small set of jobs to use the solid_queue adapter without changing the global adapter from sidekiq. 

The `config.active_job.queue_adapter` stays `:sidekiq`. Migrated jobs inherit from a temporary `SolidQueueApplicationJob` base that sets `self.queue_adapter = :solid_queue`. This way only these jobs run on Solid Queue while everything else stays on Sidekiq. This base class is temporary and once we flip the global adapter, these jobs return to `ApplicationJob` and the class is removed (except for the retry_on logic which is described below).

### Jobs migrated (and why these)

- `DestroyApplicationFormsJob` - nightly, low risk and easy to verify it runs correctly. Good first exercise of recurring-job handling.
- `TrimSessionsJob` - nightly, low risk and easy to verify it runs correctly. Further recurring-job coverage.
- `UpdateWorkingDaysJob` - nightly, low risk, easy to verify it runs correctly and is more compute-intensive so it's a good test on how heavier work behaves on Solid Queue.
- `UpdateTRSMatchJob` - This is triggered on application submission, then re-triggers itself every 24h throughout assessment (could be a few months) until it matches a TRS record to is completed. Existing ones in production currently sit in Sidekiq's Scheduled set (~2.2k in production and ~330 in this review env). We're expecting over one 24 hour cycle the whole list goes from Sidekiq's "Scheduled" set and reappears in Solid Queue's "Scheduled Jobs". This is low risk and if any are lost we can re-trigger for all applications still in assessment.
- `ClearSolidQueueFinishedJobsJob` - Given we haven't switched the global adapter yet to solid_queue, the clearing of finished jobs will end up running on Sidekiq unless we specifically pin the adapter in a custom job. 

### Concurrency (threads & processes)

Initially, following some of the other services, we're going with the following as a starting point:

1. We have 1 queue worker process (combining default, mailer and analytics) with 5 threads initially.

All are set as env vars, so it's easy to adjust per environment whenever justified.

### Retries

Solid Queue has no built-in retry (unlike Sidekiq), so `SolidQueueApplicationJob` sets `retry_on` with `wait` as `:polynomially_longer`. https://edgeapi.rubyonrails.org/classes/ActiveJob/Exceptions/ClassMethods.html.

## Testing (review)

- [x] Ensure all migrated scheduled jobs run on the review app.
- [x] Ensure UpdateTRSMatchJob now gets triggered on SolidQueue and the existing ~330 UpdateTRSMatchJob's move from Sidekiq into SolidQueue
- [x] Ensure Finished jobs get cleared as expected.
- [x] Confirm Mission Control shows jobs/queues as expected and share learnings with team before going live.
- [x] Ensure retry functionality works as expected.

## Out of scope

- Flipping the global `queue_adapter` (later PR after at least 1 week of closely monitoring).
- Full recurring/cron migration off `sidekiq-cron` and Redis removal.

## Rollback

In the case that anything goes wrong, we can have a revert PR where all the migrated jobs return to the Sidekiq adapter (back to inheriting from ApplicationJob) and recurring.yml changes back to schedule.yml.


dp-daly (approved):
Looks great! 👌

terraform/application/config/production/variables.yml
```diff
@@ -1,2 +1,8 @@
 ---
 TRS_API_URL: https://teacher-qualifications-api.education.gov.uk
+DEFAULT_QUEUE_THREADS: 1
+DEFAULT_QUEUE_PROCESSES: 1
+MAILER_QUEUE_THREADS: 1
+MAILER_QUEUE_PROCESSES: 1
+ANALYTICS_QUEUE_THREADS: 1
+ANALYTICS_QUEUE_PROCESSES: 1
```
dcyoung-dev: If I remember correctly, it's possible to set a default variable value and then override it in each of the variables.yml files.

dcyoung-dev (approved):

terraform/application/config/production/variables.yml
Hassanmir92: Yh correct, I have done that, but I also thought it's easier for someone else doing it to quickly know where to make the changes whenever they need to.

dcyoung-dev (approved):

RMcVelia (approved):

syed87 (approved):
lgtm

PR #3325, opened 2026-08-04
AQTS-1511-holdingpage-for-removed-countries

syed87:
**Ticket**: https://dfedigital.atlassian.net/browse/AQTS-1511

### Why is this change being made?

Countries eligible for QTS recognition can be removed from the service as policy changes occur. Applicants who started, but did not submit, an application before a country is closed currently retain a draft application, which can create confusion when they are no longer eligible to apply through the service.

This change ensures that applicants from newly closed countries are clearly informed that their application has been closed due to changes in the service and are signposted to alternative guidance and routes into teaching. At the same time, applicants from countries that remain eligible are able to continue accessing their draft applications without interruption.
Screenshot 2026-08-04 at 12 32 47


Hassanmir92 (approved):
✅

PR #3326, opened 2026-08-04
[AQTS-1539] Move all remaining jobs from Sidekiq to SolidQueue.

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1539

## What

Part 2 of our Sidekiq to Solid Queue migration.

1. Moving all remaining jobs from Sidekiq to Solid Queue by switching the Active Job adapter to `:solid_queue`. With the adapter flipped, the temporary `SolidQueueApplicationJob` is removed and all jobs inherit from `ApplicationJob` again. The retry logic has been moved onto `ApplicationJob`. 
2. We also moved all recurring jobs move to Solid Queue's `recurring.yml` and removed `sidekiq-cron`. 
3. The support console header now links to Solid Queue instead of Sidekiq.

## Not in scope

- Removing the `sidekiq` gem and worker app - We will keep sidekiq running for at least 1 week so both scheduled/retry sets can drain and to keep the ability for an easier rollback incase something goes wrong.
- Removing Redis - separate follow-up, and only after Sidekiq is fully removed.
- Updating docs and architecture diagram to reflect these changes.

## Testing

We will monitor this within review environment until next week before we merge.

- [x] All jobs run on Solid Queue and complete.
- [x] Recurring jobs run from the Solid Queue scheduler without a problem.
- [x] Failed jobs still retry as expected.
- [x] No new Sidekiq jobs enqueue
- [x] Monitor memory consumption with full scheduled job runs.
- [x] Mission Control shows all jobs/queues correctly and watch daily deletion of finished jobs.


syed87 (approved):
lgtm

PR #3329, opened 2026-08-04
[AQTS-1539] Exclude job arguments in logs and remove info level logs for SolidQueue

Hassanmir92:
Setting `active_job.log_arguments = false` to keep job arguments (which can in the future carry PII once we migrate all of our jobs to Solid Queue) out of logs.
Also setting log level for Solid Queue worker to `:warn` to avoid any unnecessary noise in work logs consistent with how Sidekiq used to be configured.


syed87 (approved):
lgtm

PR #3331, opened 2026-08-04
Update how email delivery record is created

Hassanmir92:
## What

This PR changes how the `EmailDelivery` audit record is created. Instead of enqueuing `EmailDeliveryAuditJob` with the email details as arguments, the record is now created directly in an `after_deliver` callback, which then enqueues the `EmailDeliveryNotifyStatusUpdateJob`.

## Why

`EmailDeliveryAuditJob` previously took the recipient address, subject, and notify details as arguments. This may lead to arguments surfacing in logs and sentry events in event of a failure where we may not want them to. Our job arguments should always only include records.

The only thing enqueued now is `EmailDeliveryNotifyStatusUpdateJob`, which carries the EmailDelivery id and not the details.

## Post deploy

Remove `EmailDeliveryAuditJob` job. Reason I'm not doing it as part of this PR is to ensure if there's a job in progress, no errors are raised.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -38,5 +38,7 @@ class ApplicationMailer < Mail::Notify::Mailer
         email_delivery,
       )
     end
+  rescue StandardError => e
+    Sentry.capture_exception(e)
   end
 end
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -3,7 +3,7 @@
 class ApplicationMailer < Mail::Notify::Mailer
   default from: I18n.t("service.email.enquiries")
 
-  after_deliver :enqueue_email_delivery_audit
+  after_deliver :create_email_delivery_audit
 
   GOVUK_NOTIFY_TEMPLATE_ID =
     ENV.fetch(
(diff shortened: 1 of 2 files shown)


app/mailers/application_mailer.rb
```diff
@@ -13,19 +13,30 @@ class ApplicationMailer < Mail::Notify::Mailer
...
+        reference_request: params[:reference_request],
+        prioritisation_reference_request:
+          params[:prioritisation_reference_request],
+      )
+
+    if notify_id
+      EmailDeliveryNotifyStatusUpdateJob.set(wait: 1.minute).perform_later(
+        email_delivery,
+      )
+    end
```
Hassanmir92, line 40: This is taken directly from https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/app/jobs/email_delivery_audit_job.rb 

Once this goes in, we can delete that job since it's no longer in use

syed87, line 40: If this failed to write to db even though the email was successfully delivered would it try to send another email?

if so we could potentially add a rescue here.


`rescue StandardError => e
  Rails.logger.error(
    "Failed to audit email delivery (#{mailer_name}##{action_name}): #{e.message}",
  )`

syed87 (approved):
lgtm

app/mailers/application_mailer.rb
Hassanmir92: Nice! - While unlikely since we don't have any validations in place, it's good practice and I'll add it.

PR #3332, opened 2026-08-05
Delete EmailDeliveryAuditJob as no longer used

Hassanmir92:
Following our update [here](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/3331) this job is no longer in use and can be deleted.


syed87 (approved):
lgtm

PR #3335, opened 2026-08-07
[AQTS-1579] Uploading LoPS received into service by assessor/admin

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1579

## What

This PR introduces a new upload functionality by assessors/admins for Letter of Professional Standing (LoPS) that have been received by teaching authorities. 

In this PR, I've added a new "attachment" attribute to `ProfessionalStandingRequestLocationForm` instead of including the `UploadableForm` which is mostly designed for the teacher interface and created some unnecessary blockers to achieving this feature. As a result, I've brought in only the relevant parts of that module into this form directly to keep things simple. 

----

### Screenshot for blank form 
Screenshot 2026-08-07 at 13 01 29

---- 

### Screenshot for when file already uploaded
Screenshot 2026-08-07 at 12 57 53

### DIFF
--- a/app/forms/assessor_interface/professional_standing_request_location_form.rb
+++ b/app/forms/assessor_interface/professional_standing_request_location_form.rb
@@ -13,10 +13,8 @@ class AssessorInterface::ProfessionalStandingRequestLocationForm
   attribute :location_note, :string
 
   attribute :attachment
-  validates :attachment,
-            file_upload: true,
-            presence: true,
-            unless: -> { document.completed? }
+  validates :attachment, file_upload: true
+  validates :attachment, presence: true, unless: -> { document.completed? }
 
   def save
     return false if invalid?
--- a/app/forms/assessor_interface/professional_standing_request_location_form.rb
+++ b/app/forms/assessor_interface/professional_standing_request_location_form.rb
@@ -8,7 +8,7 @@ class AssessorInterface::ProfessionalStandingRequestLocationForm
   validates :requestable, :user, presence: true
 
   attribute :received, :boolean
-  validates :received, inclusion: [true, false]
+  validates :received, inclusion: [true]
 
   attribute :location_note, :string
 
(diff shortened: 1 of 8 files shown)


app/forms/assessor_interface/professional_standing_request_location_form.rb
```diff
@@ -8,18 +8,24 @@ class AssessorInterface::ProfessionalStandingRequestLocationForm
...
   def save
     return false if invalid?
 
     ActiveRecord::Base.transaction do
+      create_upload! if attachment.present?
+
       if received && !requestable.received?
         ReceiveRequestable.call(requestable:, user:)
-      elsif !received && requestable.received?
-        UnreceiveRequestable.call(requestable:, user:)
```
Hassanmir92, line 22: It was never possible to unreceive the LoPS. I believe this code was left over from how this form used to be and since then things have changed.

app/forms/assessor_interface/professional_standing_request_location_form.rb
```diff
@@ -8,18 +8,24 @@ class AssessorInterface::ProfessionalStandingRequestLocationForm
   validates :requestable, :user, presence: true
 
   attribute :received, :boolean
-  validates :received, inclusion: [true, false]
+  validates :received, inclusion: [true]
 
   attribute :location_note, :string
 
+  attribute :attachment
+  validates :attachment,
```
syed87: Found a bug here.

I was able to bypass `FileUploadValidator`. If a written statement is already on file, the "Replace the uploaded file" option will accept anything. I got a `.rtf` through. The same file is correctly rejected on an application with nothing uploaded yet.

The cause is the `unless:` here. Rails applies `if`/`unless` to every validator in the same `validates` call, so when `document.completed?` is true it isn't just skipping `presence` — it skips `file_upload` too.

In theory this should also allow files above 50mb and encrypted files as well.

Fix:

```
attribute :attachment
validates :attachment, file_upload: true 
validates :attachment, presence: true, unless: -> { document&.completed? }
```

Hassanmir92: Actioned: https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/3335/commits/8f18430aeffd467c231a183196540be0fb7b3083

syed87 (approved):
Lgtm

PR #3348, opened 2026-08-18
[AQTS-1609] Remove Sidekiq and Redis

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1609

Final step of the Solid Queue migration. Removing Sidekiq and Redis now that all background jobs run on Solid Queue and caching runs on Solid Cache. I also removed ActionCable which was added as part of the default template at the beginning of the project, which we don't use and was another Redis consumer.

With this, the service no longer depends on Redis.

## Changes

- Remove `sidekiq` from the codebase.
- Remove the Sidekiq worker application from Terraform.
- Remove Redis from infrastructure and references to REDIS_URL.
- Remove Action Cable (unused).


RMcVelia (approved):
infra/terraform changes look good

elceebee (approved):
LGTM! Glad this has gone so smoothly.

syed87 (approved):
lgtm

PR #3350, opened 2026-08-20
[AQTS-1608] Improve performance of update working days job

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1608

## What

Improves the performance of `UpdateWorkingDaysJob` which was causing a CPU spike on the worker pod when its nine calculation passes ran together nightly at the same time.

Two key changes:

1. Split into separate jobs - The single job that ran nine passes back-to-back is now a coordinator that enqueues nine independent `WorkingDays::*` jobs on a staggered schedule (15-minute steps). This spreads the CPU load across 2 hours instead of it all at the same time.
2. Scope to fewer rows - The jobs that calculate values linked to a complete state (e.g. submitted to completed) now only update rows where the value hasn't been set yet. This is because most of values are fixed once already set, so there's no need to re-write the entire scope every night. For the completed ones where there is a possibility for reverse decision, I've added a reset of the working days integer in the `rollback_assessment` service to ensure we don't end up with invalid working day values. The two "since today" updates (applications since submission and assessments since started) are not scoped, as their values change daily and must be recalculated for all records.

## Potential future improvements

1. Updates remain using `update!` (not batched `update_all`) because the dfe-analytics callback must fire on each change to stream to BigQuery. Batching would skip that callback. Once dfe-analytics moves to Airbyte (a DB replica) rather than model callbacks, the need for callback triggers disappear. At that point we can drop per row `update!` in favour of batched `update_all` / `upsert_all`, reducing individual writes into a handful of statements.
2. Review with stakeholders whether the two "since today" updates need to happen when the application has moved to completed. If not, this will allow us to remove completed applications from the scope of updates. This removes the risk of this becoming a growing issue as the number of applications grow. 

## Testing

- [x] Each `WorkingDays::*` job sets its column correctly.
- [x] The coordinator enqueues all nine jobs with the expected delays.

### DIFF
--- a/app/jobs/update_working_days_job.rb
+++ b/app/jobs/update_working_days_job.rb
@@ -2,171 +2,32 @@
 
 class UpdateWorkingDaysJob < ApplicationJob
   def perform
-    update_application_forms_since_submission
-    update_application_forms_submitted_to_completed
+    WorkingDays::UpdateApplicationsSinceSubmissionJob.perform_later
+    WorkingDays::UpdateApplicationsSubmittedToCompletedJob.set(
+      wait: 15.minutes,
+    ).perform_later
 
-    update_assessments_since_started
-    update_assessments_started_to_completed
-    update_assessments_submission_to_started
-    update_assessment_started_to_verification
-    update_assessment_submission_to_verification
-    update_assessment_submission_to_prioritisation_decision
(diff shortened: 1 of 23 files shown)


syed87 (approved):
lgtm

PR #3351, opened 2026-08-20
AQTS-1577-award-to-confirm

syed87:
**Ticket:** [[AQTS-1574](https://dfedigital.atlassian.net/browse/AQTS-1574)]

## Summary

Updates content across the service, replacing “Award” with “Confirm” in line with the approved content changes.

## Changes

- Updated the eligibility checker start and “You’re eligible” pages.
- Updated application form content, including work history and Northern Ireland guidance.
- Updated the main application task list and important banner.
- Updated applicant and referee email content.
- Updated service page content and statuses.
- Updated CMS labels, decisions, declarations and confirmation messages.
- Updated assessment status content to refer to confirming or declining QTS.


app/lib/application_form_status_updater.rb
```diff
@@ -118,7 +118,7 @@ def statuses
         if trs_trn_request.potential_duplicate?
           %w[potential_duplicate_in_dqt]
         else
-          %w[awarded_pending_checks]
+          %w[confirmed_pending_checks]
```
syed87: This updates the statuses value in the application form and then this is modified to be used in the view it'self by removing the underscored ect. One potential issue is that older applications will still have stored the old wording and hence that wording will still show in the view for them on one specific page (final award page).

app/view_objects/assessor_interface/application_forms_show_view_object.rb
```diff
@@ -29,7 +29,7 @@ def task_list_sections
   end
 
   def status
-    application_form.statuses.map(&:humanize).to_sentence
+    application_form.statuses.map { |status| status_label(status) }.to_sentence
```
Hassanmir92: 💪

app/view_objects/assessor_interface/application_forms_show_view_object.rb
```diff
@@ -99,6 +99,10 @@ def management_tasks
            to: :assessment
   delegate :canonical_email, to: :teacher
 
+  def status_label(status)
+    I18n.t("application_form.status.#{status}", default: status.humanize)
```
Hassanmir92: Is `application_form.status` defined anywhere? 

I would have expected this to be `I18n.t("components.status_tag.#{status}")`

app/views/assessor_interface/assessment_recommendation_award/edit.html.erb
```diff
@@ -22,18 +22,18 @@
...
+  <p class="govuk-body">If you’re happy to proceed, read the declaration and if you want to confirm, select ‘Accept and confirm QTS’.</p>
   <p class="govuk-body"><%= t("assessor_interface.assessment_recommendation.back_instruction") %></p>
 
   <%= f.govuk_check_boxes_fieldset :declaration, multiple: false, small: true do %>
     <%= f.govuk_check_box :declaration,
                           true,
                           false,
                           link_errors: true,
-                          label: { text: "I’ve reviewed this QTS application and I’m satisfied that the applicant meets all the requirements. I understand that by continuing, the applicant will be awarded qualified teacher status." } %>
+                          label: { text: "I’ve reviewed this QTS application and I’m satisfied that the applicant meets all the requirements. I understand that by continuing, the applicant's qualified teacher status will be confirmed." } %>
```
Hassanmir92: ’ instead of '

config/locales/en.yml
```diff
@@ -4,6 +4,10 @@ en:
     success: Success
     warning: Warning
 
+  application_form:
+    status:
+      awarded_pending_checks: "confirmed pending checks"
```
Hassanmir92: I don't think you this is correct. Isn't the status QTS pending now for this? Which is already defined in "components.status_tag"

Hassanmir92 (commented):
Nice work mate, but 2 additional places I think worth updating as well as some comments:

1. Support console when updating the region form, the all sections request has an option which says "QTS can be awarded without work references and applicants...". We could change that to "QTS can be confirmed without work references and applicants"?
2. In the docs/application-forms.md maybe update award to confirm? (Don't change ADR's since those are point in time documentations).

app/view_objects/assessor_interface/application_forms_show_view_object.rb
syed87: it's now defined in config/locales/en.yml

config/locales/en.yml
syed87: "QTS pending" is saved under the same symbol awarded_pending_checks  but in config/locales/components.en.yml hence I couldn't make use of that one. I need the symbol awarded_pending_checks but I need it to point to the new content "confirmed pending checks" that's why I've defined it here.

Hassanmir92 (approved):

PR #3352, opened 2026-08-20
Remove unused code for EEA

Hassanmir92:
During some work, we identified `TRS::RecognitionRoute` which is not in use and legacy left over code. This PR removes it and any code related to it.


syed87 (approved):
lgtm

PR #3353, opened 2026-08-24
Update SLA "Country" column name to "Country/Region"

Hassanmir92:
The column was returning Country name or Region name so as a result we've updated the column name to reflect that.


syed87 (approved):
lgtm

PR #3356, opened 2026-08-25
[AQTS-1623] Working days for completed applications

Hassanmir92:
Ticket: https://dfedigital.atlassian.net/browse/AQTS-1623

On an application, there is a “Working days since submission“ which we no longer update when application moves to complete (awarded, declined or withdrawn). As a result, we want to show “Working days completed in“ for completed applications.

This shift is to reduce the unnecessary CPU used to update working days on applications that it is no longer relevant for as if we were to leave `UpdateApplicationsSinceSubmissionJob` and `UpdateAssessmentsSinceStartedJob` as they were, then the more applications we get, the more it becomes a growing problem. Instead completed applications care about `working_days_between_submitted_and_completed` and `working_days_between_started_and_completed` for analytics.

1. Updating `UpdateApplicationsSinceSubmissionJob` to scope by applications that are assessable (not in draft or completed stages).
2.  Updating `UpdateAssessmentsSinceStartedJob` to scope by assessments where the application form it belongs to is still assessable (not in draft or completed stages).
3. Covering an edge-case where completed are rolled back into assessment again and re-calculating working days since they're no longer automatically by the nightly job.
4. UI update to now display "Working days completed in" instead of "Working days since submission" for completed applications within application list view and overview.


syed87 (approved):
lgtm

